PR #46, opened 2017-10-30
Fix for single switch rules failure

a-ovchinnikov:
Due to a naming inconsistency rules for single
switch were not built properly. This commit fixes
the issue.


carmine:
The neo4j properties were fixed in a separate PR. Please adjust this PR (ie merge latest changes)

carmine:
This was resolved through a separate pull request

PR #53, opened 2017-11-07
Alternative way to pass configuration into dockers

surabujin:
Simplified way to pass configuration into containers. It use read only
"bind mounts" passed from host system into docker containers. This
"volume" should be pre populated with custom configuration for all
containers.


PinGwynn:
@surabujin what the reason for removing northbound properties from ansible template ?

carmine:
Similarly, what is the motivation for this change?

Issue #13, opened 2017-10-10
Create OpenTSDB topology


jonvestal:
Currently every topology which needs to commit to OpenTSDB does it individually.  We would like to create a Storm Topology that listens to a Kafka topic which has pre-formatted messages and then pushes those messages into OpenTSDB only if:

1. Current datapoint <> previous datapoint
2. Previous datapoint is >= 10 mins old

carmine:
@nikitamarchenko will tackle ..

PR #72, opened 2017-11-15
#13: fixed tests

nikitacherevko:


services/src/messaging/src/main/java/org/openkilda/messaging/info/InfoMessage.java
```diff
@@ -76,9 +76,9 @@ public InfoMessage(@JsonProperty(PAYLOAD) final InfoData data,
      * @param timestamp     timestamp value
      * @param correlationId message correlation id
      */
-    public InfoMessage(@JsonProperty(PAYLOAD) final InfoData data,
-                       @JsonProperty(TIMESTAMP) final long timestamp,
-                       @JsonProperty(CORRELATION_ID) final String correlationId) {
+    public InfoMessage(final InfoData data,
```
carmine: Why remove @JsonProperty?

services/src/messaging/src/main/java/org/openkilda/messaging/Message.java
```diff
@@ -99,8 +99,7 @@ public Message(@JsonProperty(TIMESTAMP) final long timestamp,
      * @param timestamp     message timestamp
      * @param correlationId message correlation id
      */
-    public Message(@JsonProperty(TIMESTAMP) final long timestamp,
-                   @JsonProperty(CORRELATION_ID) final String correlationId) {
+    public Message(final long timestamp, final String correlationId) {
```
carmine: Why remove @JsonProperty?

nikitacherevko: We have already created constructor which is responsible for the creation objects from json. Also we mark only one constructor with @JsonCreator and it parameters with @JsonPropertyr.
https://stackoverflow.com/questions/15931082/how-to-deserialize-a-class-with-overloaded-constructors-using-jsoncreator
This constructor we use just to create object from javacode (not deserializing), so we don't need it marked with @JsonCreator

services/src/messaging/src/main/java/org/openkilda/messaging/info/InfoMessage.java
nikitacherevko: the same is here (please take a look at the comment bellow)

PR #149, opened 2017-12-29
Added templates for saving log files in json for processing by filebeat

PinGwynn:
Updated ansible templates. Added an ability to generate log files in JSON format for processing by filebeat.

You can select logging options via variables in templates/defaults/main.yaml
For example:
json_file: False
logstash: True


services/src/openkilda-gui/src/main/resources/application.properties
```diff
@@ -28,27 +28,24 @@ logging.level.org.hibernate.SQL=INFO
 logging.pattern.file=%d{yyyy-MM-dd HH:mm:ss} [%thread] %-5level %logger - %X{correlationid} : %msg%n
 
 
-#Base url and port
-base.url=http://127.0.0.1
-
 #SWITCH URL
-switch.base.url=${base.url}:8081/wm/core
+switch.base.url=127.0.0.1:8081/wm/core
```
carmine: These have changed from 127.0.0.1 to ${base.url} and now back to 127.0.0.1.  Please describe why we are changing back to 127.0.0.1.

Issue #168, opened 2018-01-10
Extend Northbound API to get switches and ISLs data


nikitacherevko:
We should place all endpoints for KILDA-GUI in one service and it might be helpful for atdd testing. Right now we have only flow related data there, so in the scope of this issue we have to add links and switches endpoints

PR #176, opened 2018-01-11
added switches and links endpoints to topology-engine-rest and NB

nikitacherevko:


nikitacherevko:
closes #168

services/src/northbound/src/main/resources/northbound.properties
```diff
@@ -12,3 +12,4 @@ server.contextPath=/api/v1
 kafka.flow.topic=kilda.flow
 kafka.health.topic=kilda.health.check
 kafka.northbound.topic=kilda.northbound
+topology.engine.rest.endpoint=http://127.0.0.1:80
```
carmine: Can we use the domain name instead of IP? I'm adding 'hostname' to both topo-eng and topo-rest

services/src/northbound/src/test/resources/northbound.properties
```diff
@@ -10,3 +10,4 @@ kafka.hosts=kafka.pendev:9092
 kafka.groupid=northbound-consumer
 server.contextPath=/api/v1
 kafka.topic=kilda-northbound
+topology.engine.rest.endpoint=http://127.0.0.1:80
```
carmine: same comment as above, use the domain name.

PR #190, opened 2018-01-17
Fix RC on CacheBolt execute tuples before internal state were restored

nikitamarchenko:
We mark tuples as failed while state is not restored

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/ctrl/ClearStateAction.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/ctrl/ClearStateAction.java
@@ -1,0 +1,19 @@
+package org.openkilda.wfm.ctrl;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.google.common.annotations.VisibleForTesting;
+import org.openkilda.wfm.MessageFormatException;
+import org.openkilda.wfm.UnsupportedActionException;
+
+@VisibleForTesting
+class ClearStateAction extends CtrlEmbeddedAction {
+    ClearStateAction(CtrlAction master, RouteMessage message) {
+        super(master, message);
+    }
+
+    @Override
+    protected void handle()
+            throws MessageFormatException, UnsupportedActionException, JsonProcessingException {
+        getMaster().getBolt().clearState();
+    }
+}
--- a/services/wfm/src/main/java/org/openkilda/wfm/ctrl/ICtrlBolt.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/ctrl/ICtrlBolt.java
@@ -1,5 +1,6 @@
 package org.openkilda.wfm.ctrl;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.openkilda.messaging.ctrl.AbstractDumpState;
 import org.openkilda.wfm.IKildaBolt;
 
@@ -7,4 +8,7 @@ public interface ICtrlBolt extends IKildaBolt {
     AbstractDumpState dumpState();
 
     String getCtrlStreamId();
+
+    @VisibleForTesting
+    default void clearState() { }
 }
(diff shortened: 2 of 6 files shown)


services/wfm/src/main/java/org/openkilda/wfm/topology/cache/CacheBolt.java
```diff
@@ -130,6 +131,11 @@
         this.discoveryInterval = discoveryInterval;
     }
 
+    /**
+     * Initialization flag
+     */
+    boolean isRecivedCacheInfo = false;
```
carmine: isReceivedCacheInfo? ... ie missing an 'e' ??

services/wfm/src/main/java/org/openkilda/wfm/ctrl/ICtrlBolt.java
```diff
@@ -1,10 +1,14 @@
...
+import com.google.common.annotations.VisibleForTesting;
 import org.openkilda.messaging.ctrl.AbstractDumpState;
 import org.openkilda.wfm.IKildaBolt;
 
 public interface ICtrlBolt extends IKildaBolt {
     AbstractDumpState dumpState();
 
     String getCtrlStreamId();
+
+    @VisibleForTesting
```
carmine: Unclear why we are using @VisibleForTesting ..

services/wfm/src/main/java/org/openkilda/wfm/topology/cache/CacheBolt.java
```diff
@@ -20,6 +20,7 @@
 import static org.openkilda.wfm.topology.flow.StreamType.RESTORE;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.google.common.annotations.VisibleForTesting;
```
carmine: Unclear why we are using @VisibleForTesting ..

services/wfm/src/main/java/org/openkilda/wfm/topology/cache/CacheBolt.java
```diff
@@ -556,6 +574,15 @@ public AbstractDumpState dumpState() {
         return new CacheBoltState(networkDump, flowDump);
     }
 
+    @VisibleForTesting
```
carmine: Unclear why we are using @VisibleForTesting .. this is a public method.

services/wfm/src/main/java/org/openkilda/wfm/ctrl/ClearStateAction.java
```diff
@@ -0,0 +1,19 @@
+package org.openkilda.wfm.ctrl;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.google.common.annotations.VisibleForTesting;
+import org.openkilda.wfm.MessageFormatException;
+import org.openkilda.wfm.UnsupportedActionException;
+
+@VisibleForTesting
```
carmine: You'll see several comments on using @VisibleForTesting .. I think there are conventions within google guava for how they use it .. it is unclear how/why it is being used here. Most of the elements being tagged are public.

Whereas it is clear what "ClearState" should do .. and it may be useful for production scenarios in any case .. probably need a bit more documentation on this. AFAIK, this is included in the main code, not just test code. This brings up issues, let's discuss.

nikitamarchenko: When I write a test for that code I stuck with an initialization problem, Test load and init bolt only one for all use cases. So, I have no choice and create clear endpoint in bolt for simulate clean(non init) state of bolt without reload all storm topology. I used VisibleForTesting for that test code.If we will use that code for production I will remove that annotation and write some documentation.

services/wfm/src/main/java/org/openkilda/wfm/topology/cache/CacheBolt.java
nikitamarchenko: typo

services/wfm/src/main/java/org/openkilda/wfm/ctrl/ClearStateAction.java
carmine: I think you have a couple of options: 
1) since this is just for testing ... does it make sense to recreate the topology for each test scenario (ie use @Before and @After, vs. @BeforeClass and @AfterClass
2) remove the "public" and just make it package scope, ie don't put public / private / protected. This way, the test scenario can still access it.

nikitamarchenko: 1. I tried to do that, spend a lot of time and result was very unstable, storm & kafka didn't response and crash sometimes I tried to find what is wrong and spend another day.
2.  I will try to do that.

PR #229, opened 2018-01-29
184: implemented flow reinstalling

### LINKED ISSUE #184: Switch Bounces, Re-install Flows
The Ingress and Egress flow rules should be re-installed if a switch comes back online. This will require some logic within either the WFM and/or FLOW topology to detect that a previous switch, with some state, has come back online.

The TransitFlows need some further analysis. The logic needs to determine if kilda re-routed a flow. Presumably the transit disappears from the both cache and neo4j if a flow has changed.  There are some flows that might not get re-flowed - eg if the only path is through the switch that went down.

nikitacherevko:
closes #184

### DIFF
--- a/services/src/atdd/src/test/java/org/openkilda/atdd/FlowReinstallTest.java
+++ b/services/src/atdd/src/test/java/org/openkilda/atdd/FlowReinstallTest.java
@@ -1,9 +1,11 @@
 package org.openkilda.atdd;
 
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
 import cucumber.api.java.en.Then;
 import cucumber.api.java.en.When;
+import org.apache.commons.lang.StringUtils;
 import org.openkilda.SwitchesUtils;
 import org.openkilda.flow.FlowUtils;
 import org.openkilda.messaging.info.event.PathNode;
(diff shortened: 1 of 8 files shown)


services/src/atdd/src/test/java/org/openkilda/atdd/FlowReinstallTest.java
```diff
@@ -0,0 +1,48 @@
...
+        TimeUnit.SECONDS.sleep(1);
+        FlowPathPayload payload = FlowUtils.getFlowPath(FlowUtils.getFlowName(flowId));
+        assertTrue("Flow path should exist", payload != null && payload.getPath() != null);
+        List<PathNode> path = payload.getPath().getPath();
+        assertTrue(path.stream()
+                .anyMatch(node -> switchId.equalsIgnoreCase(node.getSwitchId())));
+    }
+
+    @Then("^flow (.*) is not built through (.*) switch")
+    public void flowPathDoesntContainSwitch(String flowId, String switchId) throws InterruptedException {
```
carmine: This implementation is mostly the same as "flowPathDoes" .. can we consolidate?

services/src/atdd/src/test/resources/features/mvp.1/FlowReinstall.feature
```diff
@@ -0,0 +1,64 @@
...
+
+  @MVP1.1
+  Scenario: Re-installing Ingress and Egress flows.
+
+  This scenario creates simple network and ingress/egress flows. Next this scenario turns off and on back, then checks whether
+  flows were re-installed again.
+
+    Given a clean flow topology
+    And a clean controller
+    And a clean flow topology
```
carmine: Why do we have "a clean flow topology" twice?
If we ahve a clean controller, don't we have a clean flow topology? (maybe not, but seems like we should)

services/src/atdd/src/test/resources/features/mvp.1/FlowReinstall.feature
```diff
@@ -0,0 +1,64 @@
+@FREINSTALL
+Feature: Flow re-reinstalling after switch comes back up.
+
+  @MVP1.1
+  Scenario: Re-installing Ingress and Egress flows.
+
+  This scenario creates simple network and ingress/egress flows. Next this scenario turns off and on back, then checks whether
```
carmine: Can you describe the topology a bit better in this description?  ie a-b-c .. and we turn off b ( or a or whatever).

services/src/atdd/src/test/resources/features/mvp.1/FlowReinstall.feature
```diff
@@ -0,0 +1,64 @@
...
+    When switch switch1 is turned off
+    Then flow pcet in DOWN state
+
+    When switch switch1 is turned on
+    Then flow pcet in UP state
+
+  @MVP1.1
+  Scenario: Re-installing transit flow when no other path is available.
+
+  This scenario created topology and builds flow through target switch. This switch will be turned off and on,
```
carmine: Similar to above, please include a simple example of the topology, which switch is turned off, etc.

PR #242, opened 2018-02-02
Use more robust way to "reboot" speaker

surabujin:
Use floodlight API to check results of speaker "reboot" in our tests.

### DIFF
--- a/services/src/atdd/src/test/java/org/openkilda/atdd/floodlight/ClearTablesTest.java
+++ b/services/src/atdd/src/test/java/org/openkilda/atdd/floodlight/ClearTablesTest.java
@@ -1,62 +1,39 @@
 package org.openkilda.atdd.floodlight;
 
-
 import static com.google.common.base.Charsets.UTF_8;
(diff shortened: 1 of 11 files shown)


services/src/atdd/src/test/java/org/openkilda/atdd/utils/controller/AbstractStaicEntry.java
```diff
@@ -0,0 +1,29 @@
...
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+
+import java.io.Serializable;
+
+@JsonSerialize
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonSubTypes({
+        @JsonSubTypes.Type(value = StaticFlowEntry.class, name = "COMMAND")})
+public class AbstractStaicEntry implements Serializable {
```
carmine: Is this suppose to be AbstractStaticEntry?

PR #289, opened 2018-02-13
Fix NullPointerException into FL

surabujin:
NPE arrives because of use result of IOFSwitchService.getSwitch call
without check for null value. This method return null values if there is
no switch with requested dpId.

There is a lot of calls to this method without check for null response
into org.openkilda.floodlight.switchmanager.SwitchManager. And there was
no simple way to fix them all.


surabujin:
It is not fully tested yet. I am going to add more tests and make whole ATDD test cycle (right now it checked only on @SMOKE set).

PR #305, opened 2018-02-18
storm version update

nikitacherevko:


nikitacherevko:
As we decided earlier there is no sense in updating storm when HDP doesn't contain latest storm/kafka versions

PR #323, opened 2018-02-22
fixed issue with duplicate nodes in neo4j

### LINKED ISSUE #318: It is possible to have duplicates in neo4j because of MERGE query
I've run stress test that sent several hundred messages with links discovered messages between two switches and then TE created several nodes with the same switch id.

nikitacherevko:
closes #318

### DIFF
--- a/services/neo4j/Dockerfile
+++ b/services/neo4j/Dockerfile
@@ -18,6 +18,10 @@ ADD ansible /ansible
 RUN ansible-playbook -s /ansible/deploy.yml
 ADD neo4j-config/neo4j.conf /etc/neo4j/neo4j.conf
 ADD neo4j-config/auth /var/lib/neo4j/data/dbms/auth
+ADD init/neo4j-queries.cql /app/neo4j-queries.cql
+ADD init/init.sh /app/init.sh
+RUN chmod 777 /app/*
+
 
 #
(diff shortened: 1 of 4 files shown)


services/neo4j/Dockerfile
```diff
@@ -18,6 +18,10 @@ ADD ansible /ansible
 RUN ansible-playbook -s /ansible/deploy.yml
 ADD neo4j-config/neo4j.conf /etc/neo4j/neo4j.conf
 ADD neo4j-config/auth /var/lib/neo4j/data/dbms/auth
+ADD init/neo4j-queries.cql /app/neo4j-queries.cql
+ADD init/init.sh /app/init.sh
+RUN chmod 777 /app/*
```
sergii-iakovenko: I don't think that neo4j-queries.cql requires "exec" permission.

services/neo4j/init/init.sh
```diff
@@ -0,0 +1,5 @@
+neo4j console &
+pip install cycli &&
+# Wait for neo4j to be and execute queries.
+./app/wait-for-it.sh -t 120 -h localhost -p 7473 -- cycli -f /app/neo4j-queries.cql
```
surabujin: Without "dot" at start of line will be better.

PR #326, opened 2018-02-26
added functionality to get all routes between two nodes

nikitacherevko:


services/topology-engine-rest/app/app/topology.py
```diff
@@ -613,6 +613,28 @@ def api_v1_topology_link_bandwidth(src_switch, src_port):
...
+        "RETURN switches"
+    ).format(src_switch, depth, dst_switch)
+
+    try:
+        result = neo4j_connect.data(query)
+
+        paths = [map(format_switch, x['switches']) for x in result]
+        return jsonify(paths)
+    except Exception as e:
+        return "error: {}".format(str(e))
```
surabujin: This "line" lead to response with status code 200 and broken json in body. You must return/raise someone who inherit HTTPException... InternalServerError looks most suitable here.

services/topology-engine-rest/app/app/topology.py
```diff
@@ -613,6 +613,28 @@ def api_v1_topology_link_bandwidth(src_switch, src_port):
     return neo4j_connect.data(query)[0]['r.available_bandwidth']
 
 
+@application.route('/api/v1/topology/routes/src/<src_switch>/dst/<dst_switch>')
+@login_required
+def api_v1_routes_between_nodes(src_switch, dst_switch):
+    depth = request.args.get('depth') or ''
+    query = (
+        "MATCH p="
+        "(src:switch{{name:'{}'}})-[:isl*..{}]->(dst:switch{{name:'{}'}}) "
```
surabujin: better use "named" substitution it is much more readable in big templates.

PR #361, opened 2018-03-14
Ignore non physical link port in FL

nikitamarchenko:


carmine:
What was the context for this change?

PR #363, opened 2018-03-14
fixed default correlation id

nikitacherevko:


services/src/northbound/src/main/java/org/openkilda/northbound/utils/RequestCorrelation.java
```diff
@@ -0,0 +1,16 @@
+package org.openkilda.northbound.utils;
+
+/**
+ * Correlation id for every single request request.
+ */
+public final class RequestCorrelation {
+
+    public static final String CORRELATION_ID = "correlation_id";
+
+    private static final ThreadLocal<String> ID = new ThreadLocal<>();
```
sergii-iakovenko: This should be InheritableThreadLocal to cover cases when we perform async tasks.

services/src/northbound/src/main/java/org/openkilda/northbound/controller/HealthCheckController.java
```diff
@@ -71,11 +66,10 @@
     @RequestMapping(value = "/health-check",
             method = RequestMethod.GET,
             produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
-    public ResponseEntity<HealthCheck> getHealthCheck(
-            @RequestHeader(value = CORRELATION_ID, defaultValue = DEFAULT_CORRELATION_ID) String correlationId) {
+    public ResponseEntity<HealthCheck> getHealthCheck() {
         logger.debug("getHealthCheck");
 
-        HealthCheck healthCheck = healthCheckService.getHealthCheck(correlationId);
+        HealthCheck healthCheck = healthCheckService.getHealthCheck(RequestCorrelation.getId());
```
sergii-iakovenko: Why don't we read the correlationId inside the service as it's done in other resources (flowController, etc.)?

sergii-iakovenko:
@nikitacherevko Please, create a branch in telstra/open-kilda and PR your changes into it.

PR #366, opened 2018-03-15
Update northbound.properties.j2 - added NB default pass from templates variables

PinGwynn:
Setting NB default pass/username from templates variables


PinGwynn:
Outdated

PR #382, opened 2018-03-23
TE Dockerfile refactoring

PinGwynn:


services/topology-engine/Dockerfile
```diff
@@ -14,23 +14,12 @@
...
-RUN pip install requests
-RUN pip install gevent
-RUN pip install kazoo
-RUN pip install python-logstash
-RUN chmod +x /usr/local/bin/uwsgi
+RUN apt-get update && \
+      apt-get install -y python-pip supervisor vim && \
+      pip install --upgrade pip && \
+      pip install neo4j-driver \
+      kafka py2neo gevent python-logstash && \
```
carmine: This line ends in && \  ... but there is no next line

services/topology-engine/Dockerfile
```diff
@@ -14,23 +14,12 @@
...
-RUN pip install py2neo
-RUN pip install requests
-RUN pip install gevent
-RUN pip install kazoo
-RUN pip install python-logstash
-RUN chmod +x /usr/local/bin/uwsgi
+RUN apt-get update && \
+      apt-get install -y python-pip supervisor vim && \
+      pip install --upgrade pip && \
+      pip install neo4j-driver \
```
surabujin: We don't use neo4j-driver it can be dropped too.

services/topology-engine/Dockerfile
PinGwynn: Fixed

services/topology-engine/Dockerfile
PinGwynn: Fixed

carmine:
All suggested changes merged

PR #383, opened 2018-03-23
Fix floodlight dockerfile

surabujin:
Previous patch to FL Dockerfile add feature to apply local patches on
fetched from git sources.

But if patch folder(s) is empty - docker build process is failing.
Current patch solves this issue.


carmine:
I've fixed this with another mechanism.

PR #384, opened 2018-03-23
Introduce DELETE SWITCH RULES method in Northbound.

### LINKED ISSUE #376: Add "DELETE_SWITCH_RULES" to NB API
This is a primitive command to delete all switch rules on a switch. The default behavior should keep the default rules (ie the the discovery rule and the drop-in-no-match rule). 

**This should not update the Flow Topology or the DB** - we have delete_flow command for that (it will delete a specific set of rules on all switches).

Key Features:
- [ ] The command should go between NB and FL
- [ ] There should be a flag that describes what to do about the default rules:  [DROP, DROP_ADD, IGNORE, OVERWRITE-
  - DROP - drop all rules, including base rules
  - DROP_ADD - drop all rules (ie DROP), then add the base rules back (discovery, drop unmatched)
  - IGNORE - drop all non-base rules (ie don't drop the base rules, leave them alone)
  - OVERWRITE - This is the same as IGNORE and ADD - ie drop all non-base rules, and install the base rules again
- [ ] Similar to the NB API for delete all flows  (ie DELETE .../flows), **use EXTRA_AUTH** in order for this command to work. Just requires an extra step to make this call; it is designed to help prevent accidentally calling this.

sergii-iakovenko:
The changes:
- Intro DELETE method for "/api/v1/switches/{switch-id}/rules" resource in Northbound.
- Send the command message from NB to FL with Reply-To topic passed.
- New "deleteAllNonDefaultRules" and "deleteDefaultRules" methods in FL SwitchManager.
- New ATDD tests for the "delete switch rules" command.

Fixes #376.

### DIFF
--- a/services/src/atdd/src/test/resources/features/mvp.1/NorthboundRun.feature
+++ b/services/src/atdd/src/test/resources/features/mvp.1/NorthboundRun.feature
@@ -64,3 +64,27 @@ Feature: Northbound tests
   This scenario setups flows across the entire set of switches and checks that these flows could be read from database
 
     Then flows dump contains 5 flows
+
+  @MVP1
+  Scenario: Delete non-default rules from a switch
+
+  This scenario setups a flow through a switch, deletes non-default rules from the switch and checks that the traffic is not pingable
+
+    Then flow nbdnr creation request with de:ad:be:ef:00:00:00:02 1 0 and de:ad:be:ef:00:00:00:03 2 0 and 1000 is successful
+    And flow nbdnr in UP state
+    And traffic through de:ad:be:ef:00:00:00:02 1 0 and de:ad:be:ef:00:00:00:03 2 0 and 1000 is pingable
+
+    Then delete all non-default rules on de:ad:be:ef:00:00:00:03 switch
+    And traffic through de:ad:be:ef:00:00:00:02 1 0 and de:ad:be:ef:00:00:00:03 2 0 and 1000 is not pingable
+
+  @MVP1
+  Scenario: Delete all rules from a switch
+
+  This scenario setups a flow through a switch, deletes all rules from the switch and checks that the traffic is not pingable
+
+    Then flow nbdar creation request with de:ad:be:ef:00:00:00:02 1 0 and de:ad:be:ef:00:00:00:03 2 0 and 1000 is successful
+    And flow nbdar in UP state
+    And traffic through de:ad:be:ef:00:00:00:02 1 0 and de:ad:be:ef:00:00:00:03 2 0 and 1000 is pingable
+
+    Then delete all rules on de:ad:be:ef:00:00:00:03 switch
+    And traffic through de:ad:be:ef:00:00:00:02 1 0 and de:ad:be:ef:00:00:00:03 2 0 and 1000 is not pingable
--- a/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
+++ b/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
@@ -1,6 +1,9 @@
 package org.openkilda.floodlight.kafka;
 
 import static org.openkilda.messaging.Utils.MAPPER;
+import static org.openkilda.messaging.command.switches.DefaultRulesAction.DROP;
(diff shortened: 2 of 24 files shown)


services/src/messaging/src/main/java/org/openkilda/messaging/command/switches/DefaultRulesAction.java
```diff
@@ -0,0 +1,34 @@
...
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+package org.openkilda.messaging.command.switches;
+
+/**
+ * Describes what to do about the switch default rules.
+ */
+public enum DefaultRulesAction {
```
carmine: Should this enum be embedded with the primary class that uses it?

services/src/northbound/src/main/java/org/openkilda/northbound/service/impl/SwitchServiceImpl.java
```diff
@@ -77,4 +98,19 @@ void init() {
                 .collect(Collectors.toList());
     }
 
+    @Override
+    public List<Long> deleteRules(String switchId, DefaultRulesAction defaultRules, String correlationId) {
+        LOGGER.debug("Delete switch rules request received");
+
+        messageConsumer.clear();
```
carmine: whereas we do this everywhere .. we do need to stop this practice ... prevents multiple outstanding requests to the  NB api

sergii-iakovenko: Agree, but definitely shouldn't be a part of these changes. I've created a separate issue (with a potential solution) - https://github.com/telstra/open-kilda/issues/405

services/src/messaging/src/main/java/org/openkilda/messaging/command/switches/DefaultRulesAction.java
sergii-iakovenko: An inner / nested class perfectly suits the case when it's initialized (instantiated) within the parent class only. Otherwise it looks confusing why the 3rd class works with the inner and doesn't involve the parent.
In our case DefaultRulesAction is used by REST resources too.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
```diff
@@ -57,37 +66,58 @@ public RecordHandler(ConsumerContext context, ConsumerRecord<String, String> rec
...
+
         try {
             CommandData data = message.getData();
             if (data instanceof DiscoverIslCommandData) {
                 doDiscoverIslCommand(data);
             } else if (data instanceof DiscoverPathCommandData) {
                 doDiscoverPathCommand(data);
             } else if (data instanceof InstallIngressFlow) {
-                doInstallIngressFlow(message);
+                doInstallIngressFlow(message, replyToTopic, replyDestination);
```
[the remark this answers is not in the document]
sergii-iakovenko: Agree. I'll replace this with "getDestinationForTopic" at the points of usage.

services/src/atdd/src/test/resources/features/mvp.1/NorthboundRun.feature
```diff
@@ -64,3 +64,27 @@ Feature: Northbound tests
...
+  Scenario: Delete non-default rules from a switch
+
+  This scenario setups a flow through a switch, deletes non-default rules from the switch and checks that the traffic is not pingable
+
+    Then flow nbdnr creation request with de:ad:be:ef:00:00:00:02 1 0 and de:ad:be:ef:00:00:00:03 2 0 and 1000 is successful
+    And flow nbdnr in UP state
+    And traffic through de:ad:be:ef:00:00:00:02 1 0 and de:ad:be:ef:00:00:00:03 2 0 and 1000 is pingable
+
+    Then delete all non-default rules on de:ad:be:ef:00:00:00:03 switch
+    And traffic through de:ad:be:ef:00:00:00:02 1 0 and de:ad:be:ef:00:00:00:03 2 0 and 1000 is not pingable
```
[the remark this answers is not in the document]
sergii-iakovenko: Good point, I'll add this.

PR #390, opened 2018-03-25
Issue/neo4j dead lock issue

carmine:


carmine:
Let's break this PR into a couple of separate PRs, one per major area.

PR #410, opened 2018-03-28
Integrate traffgen into atdd-staging.

sergii-iakovenko:
The changes:
- Add initialization checks for trafgen configuration. 
- Add support for 'skip' flag in TopologyDefinition.
- Improve the flow handling in ATDD-Staging. Minor clean up in FlowCrudSteps.
- Fix atdd-staging unit tests. 
- Add monkey_suite into ATDD-Staging.


services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/service/traffexam/FlowNotApplicableException.java
```diff
@@ -0,0 +1,8 @@
+package org.openkilda.atdd.staging.service.traffexam;
+
+public class FlowNotApplicableException extends Exception {
```
carmine: When adding a new exception, it's good to describe why/how it is used ... ie which class required this.

services/src/atdd-staging/src/test/java/org/openkilda/atdd/staging/tests/flow_crud/FlowCrudOverActiveSwitchesTest.java
```diff
@@ -0,0 +1,114 @@
...
+            mockFlowInTE("sw1", 10, "sw3", 10, 2);
+
+            mockFlowCrudInNorthbound();
+
+            mockTraffExam();
+
+            removedFlows.clear();
+        }
+
+        private void setup5SwitchTopology() throws IOException {
```
carmine: Is this yet another way to setup a topology?  How many ways in the test libraries do we have of setting up a topology?

PR #456, opened 2018-04-02
Add Network checker to probe stage #2

nikitamarchenko:
WFM stuff

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/cache/CacheBolt.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/cache/CacheBolt.java
@@ -652,6 +652,16 @@ public class CacheBolt
     }
 
     @Override
+    public AbstractDumpState dumpStateBySwitchId(String switchId) {
+        // Not implemented
+        NetworkDump networkDump = new NetworkDump(
+                new HashSet<>(),
+                new HashSet<>());
+        FlowDump flowDump = new FlowDump(new HashSet<>());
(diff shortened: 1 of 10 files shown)


services/wfm/src/main/java/org/openkilda/wfm/topology/flow/bolts/TransactionBolt.java
```diff
@@ -232,6 +232,12 @@ public String getCtrlStreamId() {
         return STREAM_ID_CTRL;
     }
 
+    @Override
+    public AbstractDumpState dumpStateBySwitchId(String switchId) {
+        // Not implemented
```
sergii-iakovenko: There's no TODO, so I assume this is the expected behavior. If we don't throw an exception here, can we at least log a warning message to notice calling of "not implemented" method?

services/wfm/src/main/java/org/openkilda/wfm/topology/cache/CacheBolt.java
```diff
@@ -651,6 +651,16 @@ public void clearState() {
         timePassed = 0;
     }
 
+    @Override
+    public AbstractDumpState dumpStateBySwitchId(String switchId) {
+        // Not implemented
+        NetworkDump networkDump = new NetworkDump(
+                new HashSet<>(),
```
sergii-iakovenko: Pls, use immutable Collections.emptySet for such cases.

PR #487, opened 2018-04-04
Proposal to fix "event" topology malfunctioning after FL flapping

surabujin:
"Event" topology can't recover after FL flapping. It detects FL death, switched
into "OFFLINE" mode, but never recover back to "ONLINE" mode.

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/TopologyConfig.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/TopologyConfig.java
@@ -17,7 +17,10 @@ public class TopologyConfig {
     private Integer discoveryInterval;
     private Integer discoveryTimeout;
     private Integer discoveryLimit;
-    private float discoverySpeakerFailureTimeout;
+    private final float discoverySpeakerFailureTimeout;
+    private final float discoveryDropOutdatedInputIn;
+    private final float discoveryDumpRequestTimeout;
+
     private String filterDirectory;
     private Level loggerLevel;
     private String loggerWatermark;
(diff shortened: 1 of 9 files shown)


services/wfm/src/main/java/org/openkilda/wfm/topology/TopologyConfig.java
```diff
@@ -17,7 +17,10 @@
     private Integer discoveryInterval;
     private Integer discoveryTimeout;
     private Integer discoveryLimit;
-    private float discoverySpeakerFailureTimeout;
+    private final float discoverySpeakerFailureTimeout;
```
sergii-iakovenko: There's no need to use float type when working with time ("long" fits the most cases). Moreover it may lead to undesirable issues when compare float values.

services/wfm/src/main/java/org/openkilda/wfm/topology/event/OFELinkBolt.java
```diff
@@ -416,10 +433,33 @@ private boolean isPortUpOrCached(String state) {
...
+
+    private void enableDumpRequestTimer() {
+        long expireDelay = (int) (dumpRequestTimeout * 1000);
+        dumpRequestTimer = new Timer(expireDelay, expireDelay, TimeUnit.MILLISECONDS);
+    }
+
+    private double calcInputMessageAge(Tuple tuple) {
+        long timestamp = tuple.getLongByField(KafkaRecordTranslator.FIELD_ID_TIMESTAMP);
+        long currentTime = System.currentTimeMillis();
+        return (currentTime - timestamp) / 1000.0;
```
sergii-iakovenko: Pls, use "long" type to keep time values + TimeUnit to perform conversion.

PR #541, opened 2018-04-11
fixing a NPE due to base config returning null

andy-wang-haibo:


services/wfm/src/main/java/org/openkilda/wfm/topology/opentsdb/bolts/OpenTSDBFilterBolt.java
```diff
@@ -27,42 +33,58 @@
...
 
     @Override
     public void prepare(Map stormConf, TopologyContext context, OutputCollector collector) {
         this.collector = collector;
     }
-
+    
     @Override
+	public Map<String, Object> getComponentConfiguration() {
+    	Config conf = new Config();
```
surabujin: Do not use tabs for indentation.

andy-wang-haibo:
tabs repalced.

PR #614, opened 2018-04-18
Delete switch rule based on port/vlan match conditions.

### LINKED ISSUE #595: Delete switch rule based on match
in addition to cookie use in-port and vlan, optionally priority
note: probably need to use strict mode.  Strict mode might need priority.

This issue is for delete switch rule operation only. The flow delete is to be covered by #634.

sergii-iakovenko:
Fixes #595

### DIFF
--- a/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
+++ b/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
@@ -1,9 +1,10 @@
 package org.openkilda.floodlight.kafka;
 
-import static org.openkilda.messaging.Utils.MAPPER;
 import static java.util.Arrays.asList;
+import static org.openkilda.messaging.Utils.MAPPER;
 
-
+import net.floodlightcontroller.core.IOFSwitch;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.openkilda.floodlight.converter.IOFSwitchConverter;
 import org.openkilda.floodlight.converter.OFFlowStatsConverter;
 import org.openkilda.floodlight.switchmanager.ISwitchManager;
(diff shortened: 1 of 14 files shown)


services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
```diff
@@ -440,55 +443,55 @@ private void doInstallSwitchRules(final CommandMessage message, String replyToTo
...
 
-            // toRemove is > 0 only if we are trying to delete base rule(s).
-            if (toRemove.size() > 0) {
-                removedRules.addAll(switchManager.deleteRuleWithCookie(dpid, toRemove));
-                if (deleteAction == DeleteRulesAction.REMOVE_ADD) {
-                    switchManager.installDefaultRules(dpid);
-                }
-            } else {
+            // The cases when we delete the default rules.
+            if (EnumSet.of(DeleteRulesAction.DROP_ALL, DeleteRulesAction.DROP_ALL_ADD_DEFAULTS,
```
carmine: This will create a new object, the same object, every function call.  Unless the compiler is smart enough to detect and optimize.  Suggest to make this a private static final and re-use in the method.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
```diff
@@ -440,55 +443,55 @@ private void doInstallSwitchRules(final CommandMessage message, String replyToTo
...
+                    DeleteRulesAction.REMOVE_DEFAULTS, DeleteRulesAction.REMOVE_ADD_DEFAULTS).contains(deleteAction)) {
+                removedRules.addAll(switchManager.deleteDefaultRules(dpid));
+            }
+
+            // The cases when we delete all non-default rules.
+            if (EnumSet.of(DeleteRulesAction.DROP_ALL, DeleteRulesAction.DROP_ALL_ADD_DEFAULTS,
```
carmine: Same comment here re: not creating a new set each time

PR #661, opened 2018-04-24
Alter ISL cost on port down event

surabujin:
On port down event both ISL(foward and reverse) receive significant
cost penalty.

There is no automatic rollback for this action. Manual update must be
done to recover normal ISL cost.

### DIFF
--- a/services/topology-engine/queue-engine/topologylistener/isl_utils.py
+++ b/services/topology-engine/queue-engine/topologylistener/isl_utils.py
@@ -20,8 +20,69 @@ import py2neo
 
 from topologylistener import db
 from topologylistener import exc
+from topologylistener import model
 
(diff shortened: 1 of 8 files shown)


services/topology-engine/queue-engine/topologylistener/isl_utils.py
```diff
@@ -20,10 +20,71 @@
...
+          (:switch {name: $src_switch})
+          -
+          [target:isl {
+            src_switch: $src_switch,
+            src_port: $src_port,
+            dst_switch: $dst_switch,
+            dst_port: $dst_port
+          }]
+          ->
+          (:switch {name: $dst_switch})
```
sergii-iakovenko: As for me this way of formatting makes Cypher query less readable. We should discuss the better way...

PR #782, opened 2018-05-15
Collect excetions in dedicated package

surabujin:
To make core layout more orginised - collect all "global" to wfm
exception into dedicate package.


services/wfm/src/main/java/org/openkilda/wfm/error/MessageException.java
```diff
@@ -1,4 +1,4 @@
-package org.openkilda.wfm.topology;
+package org.openkilda.wfm.error;
```
sergii-iakovenko: So "org.openkilda.wfm.error" doesn't represent any specific entity, domain, system component, tier, layer. Therefore that makes sense to call the package in plural "errors".

BTW, why not call it "exceptions"?

surabujin: As for me there is no big difference between error and exception. "Error" is shorter. :)

PS I have tried to find "java way" in naming such packages... and "error" was mentioned before exceptions.

PR #793, opened 2018-05-16
Create service for atdd-staging for management intermediate sw

nikitamarchenko:
Main purpose on that service is get HTTP rest request and configure OF rules on intermediate sw.

### DIFF
--- a/services/lock-keeper/lockkeeper.py
+++ b/services/lock-keeper/lockkeeper.py
@@ -1,0 +1,81 @@
+import os
+
+import paramiko
+from flask import Flask
+from flask import jsonify
+from flask import request
+
+app = Flask(__name__)
+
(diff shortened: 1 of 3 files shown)


services/lock-keeper/lockkeeper.py
```diff
@@ -0,0 +1,81 @@
...
+            'out_port': 52
+        }
+    ]
+
+
+def int_from_str_by_pattern(string, pattern):
+    start = string.find(pattern)
+    end = string.find(' ', start)
+    if end == -1:
+        end = len(string)
```
surabujin: `end = None` whould be enough here.

nikitamarchenko: > Return -1 if sub is not found. 

from  https://docs.python.org/3/library/stdtypes.html#str.find

PR #820, opened 2018-05-22
#809: Fixed self-looped ISLs errors in CacheBolt

### LINKED ISSUE #809: "self-loop" error in CacheBolt
During processing isl events CacheBolt fails with exception
```
Cannot add self-loop edge on node SwitchInfoData{switch_id=00:00:cc:37:ab:32:f2:94, state=ACTIVATED, address=198.18.196.131:53893, hostname=198.18.196.131, description=Sonus Networks Inc, 4 Technology Park Dr, Westford, MA 01886, USA OF_12 8.1.0.14, controller=198.18.196.131}, as self-loops are not allowed. To construct a graph that allows self-loops, call allowsSelfLoops(true) on the Builder.",
      "name": "java.lang.IllegalArgumentException
```
if both ends of received ISL are attached to the same switch.

Because we don't use "network topology" created from this and other "network events" this error does not affect our functionality. But it must be fixed.

nikitacherevko:
Fixed self-looped ISLs errors in CacheTopology/ET

closes #809


services/wfm/src/main/java/org/openkilda/wfm/topology/cache/CacheBolt.java
```diff
@@ -295,7 +295,11 @@ private void handleIslEvent(IslInfoData isl, Tuple tuple, String correlationId)
                 if (networkCache.cacheContainsIsl(isl.getId())) {
                     networkCache.updateIsl(isl);
                 } else {
-                    networkCache.createIsl(isl);
+                    if (networkCache.isSelfLoopedIsl(isl)) {
+                        logger.warn("Found self-looped ISL: {}", isl);
```
surabujin: "filter out" instead of "find" will be more correct and obvious for readers.

PR #826, opened 2018-05-23
Use last commited kafka offset for OpenTSDB topology spout

surabujin:
To not loose stats data during topology restart or any other outages
switch it to use last commited offset (instead of LATEST offset).


surabujin:
Closes 824

PR #838, opened 2018-05-25
atdd-staging: updated existing crud test to update vlans

rtretyak:
updated existing crud test to update vlans

### DIFF
--- a/services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/FlowCrudSteps.java
+++ b/services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/FlowCrudSteps.java
@@ -33,10 +33,16 @@ import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
+import com.google.common.collect.ContiguousSet;
+import com.google.common.collect.DiscreteDomain;
+import com.google.common.collect.Range;
+import com.google.common.collect.RangeSet;
+import com.google.common.collect.TreeRangeSet;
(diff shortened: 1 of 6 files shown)


services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/FlowCrudSteps.java
```diff
@@ -146,31 +139,31 @@ public void defineFlowsOverActiveTraffgens() {
...
         });
-
-        flows = builder.getFlows();
     }
 
     @And("^each flow has unique flow_id$")
     public void setUniqueFlowIdToEachFlow() {
-        flows.forEach(flow -> flow.setId(format("%s-%s", flow.getId(), UUID.randomUUID().toString())));
+        flowSetBuilder.getFlows()
+                .forEach(flow -> flow.setId(format("%s-%s", flow.getId(), UUID.randomUUID().toString())));
```
sergii-iakovenko: This is not obvious that ***Builder builds always the same object. Many think that a builder must return a new instance each time.

services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/helpers/FlowSet.java
```diff
@@ -44,6 +44,19 @@
         return unmodifiableSet(flows);
     }
 
+    /**
+     * Returns an unallocated vlan. The returned vlan is immediately added to the list of allocated vlans
+     */
+    public Integer allocateVlan() {
```
sergii-iakovenko: Any reason to use wrapper class instead of primitive? If it can be null, then a check is missing.

Issue #783, opened 2018-05-15
Incorrect flow status if new flow is requested while floodlight is down


rtretyak:
If you ever to create a new flow when the floodlight is not available (e.g. floodlight restarts or is down for some time), the flow status will forever stuck in 'In progress' status, but all the correct rules will successfully be installed after floodlight comes back up.
**Steps:**
1. Bring the floodlight down or init a floodlight restart
2. While floodlight is down send a flow creation request via Northbound
3. Bring the floodlight back up or wait while FL is up after restart
4. Call NB flow validate endpoint for the created flow > observe no discrepancies (flow rules are correctly installed)
5. Call NB to get flow status
*Expected:* flow is in 'Up' state
*Actual:* flow state is stuck in 'In progress'

discovered while testing #118

rtretyak:
Implemented atdd test for this case on atdd/floodlight_restart branch

rtretyak:
Needs to be retested

PR #846, opened 2018-05-30
#783: Added test to cover floodlight outage scenario

rtretyak:
Also removed retryPolicy bean as it causes unstable behavior in tests


rtretyak:
Cancelling this pr. Will be merged whenever #783 is implemented

Issue #769, opened 2018-05-14
Eliminate read operations from TER (switches, ISLs) - it takes too long


nikitacherevko:
Create separate storm topology to perform simple read operations from DB for NB calls.

pzakatov:
relates to #835 and #836

PR #850, opened 2018-05-30
#769: fixed atdd tests

nikitacherevko:
Fixed atdd tests

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/NeoOperationsBolt.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/NeoOperationsBolt.java
@@ -52,7 +52,8 @@ public abstract class NeoOperationsBolt extends AbstractBolt {
         final String correlationId = input.getStringByField("correlationId");
         getLogger().debug("Received operation request");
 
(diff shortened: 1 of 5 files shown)


services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/NeoOperationsBolt.java
```diff
@@ -52,7 +52,8 @@ protected void handleInput(Tuple input) {
         final String correlationId = input.getStringByField("correlationId");
         getLogger().debug("Received operation request");
 
-        try (Session session = driver.session(request.isReadRequest() ? AccessMode.READ : AccessMode.WRITE)) {
+        // todo (Nikita C): temporary solution to make all requests in read mode.
```
surabujin: Better throw exception on write session requests. In current state behavoiour become not obvious, and can make a trouble for person who will implement first "writable" handler.

nikitacherevko: Agree, but I didn't implement correctly any way to define whether it request is in the read mode or in write. Will do it soon, after the release

PR #857, opened 2018-06-02
Fix kafka producer guaranteed order feature

### LINKED ISSUE #854: FL send all messages to 0 partition of kafka
KafkaTopicOrderDescriptor::isEnabled always True for default topics

surabujin:
Due to error in `is guaranteed order` condition it's result was always
true. And guaranteed order was always enables for all produced messages.

Closes #854

### DIFF
--- a/services/src/floodlight-modules/src/test/java/org/openkilda/floodlight/kafka/ProducerTest.java
+++ b/services/src/floodlight-modules/src/test/java/org/openkilda/floodlight/kafka/ProducerTest.java
@@ -1,0 +1,113 @@
+/* Copyright 2018 Telstra Open Source
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


services/src/floodlight-modules/src/test/java/org/openkilda/floodlight/kafka/ProducerTest.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.concurrent.Future;
+
+public class ProducerTest extends EasyMockSupport {
+    private static final String TOPIC = "A";
+
+    private org.apache.kafka.clients.producer.Producer<String, String> kafkaProducer;
+    private Producer subject;
+
+    @Before
+    @SuppressWarnings("unchecked")
```
sergii-iakovenko: You can put SuppressWarnings on a specific line of code, this will give more context to a reader / maintainer.

PR #863, opened 2018-06-04
Track kafka producer send status

### LINKED ISSUE #222: Track issues in publish messages into kafka
Java kafka-client library use asynchronous publish mechanism by default. When we call `produser.send(payload);` it store payload for future asynchronous send.

This is not a problem while we have stable network connection to kafka server. But when network outage happen, and kafka-client drop some prepared to send messages, our code will know nothing about it.

Switching to synchronous only publish model can resolve this issue. But it also can lead to significant performance degradation. Need future investigation about handling issues during message publishing.

Known code involved in this issue:
* org.openkilda.floodlight.kafka.KafkaMessageProducer#postMessage

surabujin:
Provide 2 method to send kafka message. One of them return SendStatus
object, that should be used by caller to identify send status(use Future
object inside).

The second one, will return void, but track status by itself. In case of
error it will be logged.

Closes #222


services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/producer/OrderAwareWorker.java
```diff
@@ -0,0 +1,84 @@
...
+        } else {
+            record = new ProducerRecord<>(getTopic(), partition, null, payload);
+        }
+
+        Future<RecordMetadata> promise = getKafkaProducer().send(record, callback);
+        if (partition == null) {
+            try {
+                partition = promise.get().partition();
+            } catch (Exception e) {
+                // exceptions will be processes by caller on it's own promise.get() call
```
nikitamarchenko: anyway, add log error

surabujin: This code is on "library" side, it don't know what to do with errors.

If/when caller call org.openkilda.floodlight.kafka.producer.SendStatus#waitTillComplete it will receive same exception and can handle it in any suitable for him way.

If caller use method org.openkilda.floodlight.kafka.producer.Producer#sendMessageAndTrack, this error will be logged into org.openkilda.floodlight.kafka.producer.Producer#reportError (You can see it here org.openkilda.floodlight.kafka.producer.ProducerTest#errorReporting).

Do you insist on log message in this point?

PR #877, opened 2018-06-08
Increase cose on port down only once

### LINKED ISSUE #875: TE increase ISL cost on each "port down" event
ISL update cost workflow wasn't property developed. That lead to increase ISL's cost on each port down on some constant(10000). 

Cost must be increased only once and stay untouched, untill it will be decremented by operator. Later cost can be increased again (on port down event).

surabujin:
Do not increase ISL cost on repeating `port down` events. First `port
down` event trigger cost increse, but consequitive `port down` event
will not trigger cost increase.

Only it ISL cost will be reset to `normal` value, `port down` event can
trigger cost increase againe.

Closes #875

### DIFF
--- a/services/topology-engine/queue-engine/topologylistener/db.py
+++ b/services/topology-engine/queue-engine/topologylistener/db.py
@@ -36,6 +36,7 @@ def create_p2n_driver():
 
 
 def log_query(marker, q, p):
+    q = q.strip()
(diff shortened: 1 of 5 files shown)


pzakatov:
> Title: Increase cose on port down only once
@surabujin , increase what?

surabujin:
@pzakatov typo... it must be "cost". I will recreate PR.

PR #901, opened 2018-06-19
added limit for link_props properties

nikitacherevko:


services/topology-engine/queue-engine/topologylistener/model.py
```diff
@@ -28,10 +28,13 @@
 logger = logging.getLogger(__name__)
 
 
-def convert_integer(raw):
+def convert_integer(raw, limit=2147483647):
```
rtretyak: why not `sys.maxsize`/`sys.maxint`?

Issue #904, opened 2018-06-19
No response for NB if send update link props request with non-int cost


rtretyak:
1. Send update link props request with cost property set as an empty string `"cost": ""`

**Expected:** To be defined how exactly to react, but we should receive a response back from the system as soon as we see the problem
**Actual:** NB times out after X seconds waiting for response from the rest of the system.
Note: the case is reproducible for Any string value which cannot be converted to integer (not only empty string)

PR #911, opened 2018-06-20
#904: fixed saving into link_props not numeric values

nikitacherevko:


rtretyak:
This most likely depends on the bitness of operating systems. If TE is running on 64-bit and neo is running 32-bit, then it is possible that neo won't be able to swallow 2^64

PR #920, opened 2018-06-22
ISL status is set to Moved on physically re-plugging of ISL

### LINKED ISSUE #815: ISL status won't change to Moved when physically replugging ISL
*Steps*
1. given 3 switches, two of them connected: (S1)-[port1] => [port1]-(S2)    (S3)
2. when disconnect S1 from S2 (important: bring port1 down on **src switch** SW1 in case of any intermediate switches in between), and then connect it to S3 and bring the port back up so that it becomes
(S1)-[port1] => [port1]-(S3)    (S2)
3. then
expected: S1 -> S2 link becomes Moved, S1 -> S3 link becomes discovered
actual: S1 -> S2 link becomes FAILED

sergii-iakovenko:
The changes:
- Fix the ISL move detection logic: look for a moved ISL in the removedFromDiscovery list in the case when a non-active link is already created by a portUp event.
- Add support of the clearState control request by OFELinkBolt. And utilize it in ATDD tests to clean up the removedFromDiscovery list. 
- Add ATDD-Staging scenario to cover the case.

Fixes #815

### DIFF
--- a/services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/IslSteps.java
+++ b/services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/IslSteps.java
@@ -15,6 +15,10 @@
 
 package org.openkilda.atdd.staging.steps;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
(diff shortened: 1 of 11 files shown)


sergii-iakovenko:
The base branch has been changed to develop.

PR #936, opened 2018-06-25
Fixed non-numeric link_props value

nikitacherevko:


services/topology-engine/queue-engine/topologylistener/model.py
```diff
@@ -31,7 +31,10 @@
 
 def convert_integer(raw, limit=sys.maxint):
     if not isinstance(raw, (int, long)):
-        value = int(raw, 0)
+        try:
+            value = int(raw, 0)
+        except ValueError:
+            raise exc.UnacceptableDataError(raw, 'not numeric type of link_props value: {}'.format(raw))
```
surabujin: This converter used not only in link props. You can't mention link props in this error message.

PR #938, opened 2018-06-26
Change topology-engine consumer.group

vriabyk:


templates/templates/topology-engine/topology_engine.ini.j2
```diff
@@ -1,5 +1,5 @@
 [kafka]
-consumer.group=python-tpe-tl-consumer
+consumer.group=kilda-tpe
```
surabujin: We definitely should use variable here.

PR #946, opened 2018-06-27
Added possibility to remove meter from the switch

nikitacherevko:


nikitacherevko:
Need to test it on switches.

services/src/northbound/src/main/java/org/openkilda/northbound/controller/SwitchController.java
```diff
@@ -235,4 +237,15 @@ public RulesValidationResult validateRules(@PathVariable(name = "switch_id") Str
     public RulesSyncResult syncRules(@PathVariable(name = "switch_id") String switchId) {
         return switchService.syncRules(switchId);
     }
+
+    /**
+     * Removes meter from the switch.
+     */
+    @ApiOperation(value = "Delete meter from the switch", response = DeleteMeterResult.class)
+    @GetMapping(path = "/switches/{switch_id}/meter/{meter_id}")
```
rtretyak: Should be DELETE method

rtretyak:
Not working on staging, I am not receiving response back from northbound, it just hangs and then times-out. Wasn't able to retrieve any valuable info from kibana logs.

rtretyak:
We should also consider whether we need endpoints for getting meters by id, getting all meters and bulk deleting all meters (maybe as a separate task or don't need at all)

nikitacherevko:
Closes #952

services/src/northbound/src/main/java/org/openkilda/northbound/service/impl/SwitchServiceImpl.java
```diff
@@ -243,4 +246,16 @@ public RulesSyncResult syncRules(String switchId) {
         return switchMapper.toRulesSyncResult(validationResult, syncResponse.getInstalledRules());
     }
 
+    @Override
+    public DeleteMeterResult deleteMeter(String switchId, long meterId) {
+        final String requestId = RequestCorrelationId.getId();
```
abochkarev: What is the benefit of using final with variables here?

services/src/northbound/src/main/java/org/openkilda/northbound/service/SwitchService.java
```diff
@@ -86,4 +87,11 @@
      * @return the synchronization result.
      */
     RulesSyncResult syncRules(String switchId);
+
+    /**
+     * Removed meter from the switch.
```
abochkarev: Removed -> Removes. Could you please also divide the function describing section and params with \n?

services/src/northbound/src/main/java/org/openkilda/northbound/controller/SwitchController.java
```diff
@@ -235,4 +237,15 @@ public RulesValidationResult validateRules(@PathVariable(name = "switch_id") Str
     public RulesSyncResult syncRules(@PathVariable(name = "switch_id") String switchId) {
         return switchService.syncRules(switchId);
     }
+
+    /**
+     * Removes meter from the switch.
+     */
```
abochkarev: Will it be possible to add the params section here?

Issue #405, opened 2018-03-27
Async request processing in Northbound


sergii-iakovenko:
Re-implement the Northbound and Kafka communication approach. 

The potential solution looks like:
1. Create a new layer between NB services and kafka producers/consumers. It will manage all async requests and should return CompletableFuture.
2. Create simple mechanism to get the response.
3. The response dispatcher may reject new request if there're too many requests being processed. (to be discussed)
4. Use the CompletableFuture as the REST method result.

- The response dispatcher should subscribe on DeferredResult events 
(onTimeout and others) and react correspondingly. E.g. removes a correlationId->CompletableFuture entry if it waits a response for too long (timeout happens).

nikitacherevko:
Updated description due to latest offline discussion.

PR #978, opened 2018-07-05
#405: Implemented async request processing for NB

nikitacherevko:


services/src/northbound/src/main/java/org/openkilda/northbound/config/MessageConsumerConfig.java
```diff
@@ -99,12 +105,12 @@
...
     @Bean
-    public ConcurrentKafkaListenerContainerFactory<String, String> kafkaListenerContainerFactory() {
-        ConcurrentKafkaListenerContainerFactory<String, String> factory =
+    public ConcurrentKafkaListenerContainerFactory<String, Message> kafkaListenerContainerFactory() {
+        ConcurrentKafkaListenerContainerFactory<String, Message> factory =
                 new ConcurrentKafkaListenerContainerFactory<>();
         factory.setConsumerFactory(consumerFactory());
         factory.getContainerProperties().setPollTimeout(POLL_TIMEOUT);
-        //factory.setConcurrency(10);
+        factory.setConcurrency(10);
```
surabujin: It should be configuration option

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessageListener.java
```diff
@@ -0,0 +1,63 @@
...
+import org.slf4j.MDC;
+import org.slf4j.MDC.MDCCloseable;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.kafka.annotation.KafkaHandler;
+import org.springframework.kafka.annotation.KafkaListener;
+
+/**
+ * Kafka message receiver.
+ */
+@KafkaListener(id = "northbound-listener", topics = {
```
surabujin: If "id" define consumer group - then the word "listener" is excess word here.

And it must respect "environment prefix".

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessageListener.java
```diff
@@ -0,0 +1,63 @@
...
+     * Handles all messages from kafka and sends to corresponding component for further processing.
+     * @param message received message.
+     */
+    @KafkaHandler
+    public void onMessage(Message message) {
+        try (MDCCloseable closable = MDC.putCloseable(CORRELATION_ID, message.getCorrelationId())) {
+            logger.debug("Message received: {} - {}", Thread.currentThread().getId(), message);
+            messagingChannel.onResponse(message);
+
+            // fixme: it is for the support of outdated services. should be removed once KafkaConsumer will be deleted.
```
surabujin: Invalid "FIXME" definition :)

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,191 @@
...
+     * If this response is for the pended request then such request will be marked as completed.
+     * Messages might be chunked and not chunked. If chunked we need to wait until we receive the last one
+     * and only then collect all responses and complete the request.
+     * @param message received message.
+     */
+    void onResponse(Message message) {
+        if (!isValid(message)) {
+            logger.warn("Received invalid message: {}", message);
+            return;
+        } else if (message instanceof ErrorMessage) {
```
surabujin: "else" is excess here

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,191 @@
...
+            ErrorMessage error = (ErrorMessage) message;
+            logger.error("Response message is error: {}", error);
+            throw new MessageException(error);
```
surabujin: We need more "verbose" name for with exception.

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        InfoMessage infoMessage = (InfoMessage) message;
```
surabujin: It can/should be placed into "else" branch of following "if" (line 121).

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,191 @@
...
+        if (infoMessage instanceof ChunkedInfoMessage) {
+            ChunkedInfoMessage chunked = (ChunkedInfoMessage) infoMessage;
+            processChunkedMessage(chunked);
+        } else {
+            CompletableFuture<InfoMessage> request = pendingRequests.remove(infoMessage.getCorrelationId());
+            if (request != null) {
+                request.complete(infoMessage);
+            } else {
+                logger.warn("Received non-pending message: {}", infoMessage);
```
surabujin: "warn" level can be too verbose here. Especially when we will have multiple NB instances.

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,191 @@
...
+        }
+    }
+
+    /**
+     * Tries to find and collect all chunked messages into one chain.
+     */
+    private void processChunkedMessage(final ChunkedInfoMessage received) {
+        ChunkedInfoMessage message = received;
+
+        LinkedList<ChunkedInfoMessage> currentChain = messagesChains.stream()
```
surabujin: You can define a Map and use "next-request-id" of pending chains as key in it. (It will require to update this map when you receive next chunk but even in this case it will be faster - it is one map.remove + map.put operation on each received chunk.

Also this approach will significantly simplify following loop.

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,191 @@
...
+                .filter(chain -> chain.getLast().getNextRequestId().equals(received.getCorrelationId()))
+                .findAny()
```
surabujin: Will ".findFirst()"  work better here?

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,191 @@
...
+     * Requests that are in progress of processing.
+     */
+    private final Map<String, CompletableFuture<InfoMessage>> pendingRequests = new ConcurrentHashMap<>();
+    private final Map<String, CompletableFuture<List<ChunkedInfoMessage>>> pendingChunkedRequests =
+            new ConcurrentHashMap<>();
+
+    /**
+     * Collects messages that are not related to any known chain.
+     */
+    private final Map<String, ChunkedInfoMessage> chunkedMessages = new ConcurrentHashMap<>();
```
surabujin: unlinkedChunks... or unconsumedChunks is more suitable name, as for me.

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,191 @@
...
+
+                    currentChain = chain;
+                } else {
+                    chunkedMessages.put(correlationId, message);
+                }
+            }
+
+            if (currentChain != null) {
+                final String nextCorrelationId = message.getNextRequestId();
+                if (StringUtils.isEmpty(nextCorrelationId)) {
```
surabujin: Comparison with null is more verbose. Also we treat messages with "empty" correlationId as invalid messages... So why should we allow to have an empty string here as a valid value?

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,191 @@
...
+        LinkedList<ChunkedInfoMessage> currentChain = messagesChains.stream()
+                .filter(chain -> chain.getLast().getNextRequestId().equals(received.getCorrelationId()))
+                .findAny()
+                .orElse(null);
+
+        while (message != null) {
+            final String correlationId = message.getCorrelationId();
+            ChunkedInfoMessage nextMessage = null;
+
+            if (currentChain != null) {
```
surabujin: Why this condition are inside loop?

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessageListener.java
nikitacherevko: it is not related to consumer group

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    /**
+     * Processes messages that come back to NB topics, usually messages come as a response to some request.
+     * If this response is for pended request then such request will be marked as completed.
+     * Messages might be chunked and not chunked. If chunked we need to wait until we receive the last one
+     * and only then collect all responses and complete the request.
+     * @param message received message.
+     */
+    void onResponse(Message message) {
```
sergii-iakovenko: It's not clear how to handle the timeout case: a future object with no response must be automatically removed from pendingRequests and pendingChunkedRequests.

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,202 @@
...
+
+    @Autowired
+    private MessageProducer messageProducer;
+
+    /**
+     * Creates storages, that are able to remove outdated messages and requests.
+     */
+    @PostConstruct
+    public void setUp() {
+        pendingRequests = new PassiveExpiringMap<>(expiredTime, TimeUnit.MINUTES, new ConcurrentHashMap<>());
```
surabujin: What will happen with future object stored on caller side? We are responsible to "close" it with meaningfuull exception thrown in case of timeout.

But this "silent" drop approach doesn't care about called waiting for a response on future object...

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,202 @@
...
+            future.complete(infoMessage.getData());
+        }
+    }
+
+    /**
+     * Performs searching and collecting all chunked messages into one chain if possible.
+     */
+    private synchronized void processChunkedMessage(ChunkedInfoMessage received) {
+        List<ChunkedInfoMessage> chain;
+        if (messagesChains.containsKey(received.getCorrelationId())) {
```
surabujin: We can put empty list here in sendAndGetChunked method, and remove this check.

PS we don't need to call .containsKey, just
```
chain = messagesChains.remove(key);
if (chain != null) {
    chain.add(received);
    if (received.getNextCorrelationId() != null) {
        messagesChains.put();
    } else {
        handleChainComplete();
    }
}

File: services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java [hunk 16]
```diff
@@ -0,0 +1,202 @@
...
+        } else if (pendingChunkedRequests.containsKey(received.getCorrelationId())) {
+            chain = Lists.newArrayList(received);
+        } else {
+            unlinkedChunks.put(received.getCorrelationId(), received);
+            return;
+        }
+
+        ChunkedInfoMessage tail = received;
+        // Check whether next messages for current chain is already received. If yes - add them into that chain.
+        while (tail.getNextRequestId() != null && unlinkedChunks.containsKey(tail.getNextRequestId())) {
```
surabujin: Recursion can be here, you can call yourself to handle unlinked chains. :)

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,202 @@
...
+        } else {
+            messagesChains.put(tail.getNextRequestId(), chain);
+        }
+    }
+
+    /**
+     * Completes request.
+     */
+    private void completeChain(List<ChunkedInfoMessage> chain) {
+        List<InfoData> data = chain.stream()
```
surabujin: We definitely should try to remove null responses. 

PS Is null response can be in any place in the chain? Or only at last element? If only last - than why we scan all chain?
PSS We can filter out `null` values into `processChunkedMessage`

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,202 @@
...
+                .map(ChunkedInfoMessage::getData)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList());
+
+        ChunkedInfoMessage firstMessage = chain.get(0);
```
surabujin: If there is no any messages?

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
nikitacherevko: In case of huge amount of chunked messages there will be an error

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
nikitacherevko: There will be always at least one message, because it is a response from some component. And such message might have empty (null) payload (InfoData in our case)

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,202 @@
...
+    void onResponse(Message message) {
+        if (!isValid(message)) {
+            logger.warn("Received invalid message: {}", message);
+            return;
+        }
+
+        if (message instanceof ErrorMessage) {
+            ErrorMessage error = (ErrorMessage) message;
+            logger.error("Response message is error: {}", error);
+            throw new MessageException(error);
```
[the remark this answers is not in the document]
nikitacherevko: Yes, good point

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
nikitacherevko: Fixed. In case of timeout DeferredResult.onTimeout will be called by spring (see org.openkilda.northbound.utils.async.CompletableFutureReturnValueHandler.ControlledDeferredResult)

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
nikitacherevko: Fixed

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,261 @@
...
+            // adds already received messages to the current chain.
+            do {
+                ChunkedInfoMessage message = unlinkedChunks.remove(nextRequestId);
+                if (message.getData() != null) {
+                    chain.add(message);
+                }
+                nextRequestId = message.getNextRequestId();
+            } while (isMessageAlreadyReceived(nextRequestId));
+
+            // check if we it is the last element for this chain.
```
surabujin: em...

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,261 @@
...
+        unlinkedChunks.put(received.getCorrelationId(), received);
+
+        List<ChunkedInfoMessage> chain = messagesChains.remove(received.getCorrelationId());
+        if (chain != null) {
+            String nextRequestId = received.getCorrelationId();
+
+            // adds already received messages to the current chain.
+            do {
+                ChunkedInfoMessage message = unlinkedChunks.remove(nextRequestId);
+                if (message.getData() != null) {
```
surabujin: If you move this filter inside `completeChain` you can drop `chain.isEmpty()` check... because chaincorrelation id will be always in first chain element.

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,260 @@
...
+
+            completeWithError(error);
+        } else if (message instanceof InfoMessage) {
+            if (message instanceof ChunkedInfoMessage) {
+                processChunkedMessage((ChunkedInfoMessage) message);
+            } else if (pendingRequests.containsKey(message.getCorrelationId())) {
+                InfoMessage infoMessage = (InfoMessage) message;
+                pendingRequests.remove(message.getCorrelationId())
+                        .complete(infoMessage.getData());
+            }
```
sergii-iakovenko: Does it make sense to log the message if the correlation id has no corresponding CompletableFuture?

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessageConsumer.java
```diff
@@ -122,36 +94,8 @@ public Message poll(final String correlationId) {
...
-    */
-
-    /**
-     * {@inheritDoc}
-     */
     @Override
-    public void clear() {
-        //we shouldn't clear up collection, outdated messages are removing by default.
-        //messages.clear();
+    public void onResponse(Message message) {
```
surabujin: Is it thread safe?

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,238 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public CompletableFuture<InfoData> sendAndGet(String topic, Message message) {
+        CompletableFuture<InfoData> future = new CompletableFuture<>();
+
+        ListenableFuture<SendResult<String, Message>> futureResult = messageProducer.send(topic, message);
+        futureResult.addCallback(
+                sentResult -> pendingRequests.put(message.getCorrelationId(), future),
```
surabujin: From race condition point of view - better to do `pendingRequests.put(message.getCorrelationId(), future)` unconditionally i.e. even before `messageProducer.send(...)` and remove this record in case of error. In our case we already have clean via future object.

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,238 @@
...
+    /**
+     * Checks whether a message has correlationId and has known type or not.
+     */
+    private boolean isValid(Message message) {
+        if (StringUtils.isEmpty(message.getCorrelationId())) {
+            logger.warn("Received message without correlation id: {}", message);
+            return false;
+        }
+
+        if (message instanceof InfoMessage || message instanceof ErrorMessage) {
```
surabujin: Inverse condition to match error case. At the method end put `return true;`.

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,238 @@
...
+            return;
+        }
+
+        if (message instanceof ErrorMessage) {
+            ErrorMessage error = (ErrorMessage) message;
+            logger.error("Response message is error: {}", error);
+
+            completeWithError(error);
+        } else if (message instanceof InfoMessage) {
+            if (pendingChunkedRequests.containsKey(message.getCorrelationId())) {
```
surabujin: It will fail it someone send 2 message with same correlation id via `sendAndGet` and via `sendAndGetChunked` at same time.

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,238 @@
...
+                processChunkedMessage((ChunkedInfoMessage) message);
+            } else if (pendingRequests.containsKey(message.getCorrelationId())) {
```
surabujin: Is it thread safe? Is it executed from multiple threads?

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,238 @@
...
+                .complete(response);
+    }
+
+    /**
+     * Completes a request with an error response.
+     */
+    private void completeWithError(ErrorMessage error) {
+        String correlationId = error.getCorrelationId();
+
+        if (pendingRequests.containsKey(correlationId)) {
```
surabujin: same question regarding thread safety.

services/wfm/src/main/java/org/openkilda/wfm/topology/flow/bolts/CrudBolt.java
```diff
@@ -733,19 +731,16 @@ private void handleDumpRequest(CommandMessage message, Tuple tuple) {
...
-            Message response = new ChunkedInfoMessage(null, System.currentTimeMillis(), requestId, null);
+            Message response = new ChunkedInfoMessage(null, System.currentTimeMillis(), requestId, requestId,
+                    flows.size());
             outputCollector.emit(StreamType.RESPONSE.toString(), tuple, new Values(response));
         } else {
-            Iterator<BidirectionalFlow> iterator = flows.iterator();
-            do {
-                BidirectionalFlow flow = iterator.next();
-                String nextRequestId = iterator.hasNext() ? UUID.randomUUID().toString() : null;
+            for (int i = 0; i < flows.size(); i++) {
```
surabujin: `foreach` looks cheaper and more suitable here...

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessageConsumer.java
nikitacherevko: KafkaMessageConsumer.java will be deleted

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
nikitacherevko: Fixed

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,267 @@
...
+            messagesChains.remove(received.getCorrelationId());
+        }
+    }
+
+    /**
+     * Completes pending request with received responses.
+     */
+    private void completeRequest(List<ChunkedInfoMessage> chain) {
+        if (CollectionUtils.isEmpty(chain)) {
+            throw new IllegalStateException("Chain of messages should not be empty");
```
surabujin: one of the previous checks is `chain.size() == received.getTotalMessages() || received.getTotalMessages() == 0)`... it allow to think that 0 is valid size for chain.

Even if 0 is not valid size - we must notify caller(complete future) with error, not just drop this exception in code that know nothing what to do with it. We can't extract data from chunks, because there is 0 of them... but we have a `received` message in method `processChunkedMessage`, so we can use it's correlationId to lookup future object.

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,267 @@
...
+
+        ChunkedInfoMessage first = chain.get(0);
+        String requestId = first.getCorrelationId();
+
+        List<InfoData> response;
+        if (first.getData() != null) {
+            response = chain.stream()
+                    .map(ChunkedInfoMessage::getData)
+                    .collect(Collectors.toList());
+        } else {
```
surabujin: weird logic

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,267 @@
...
+            response = Collections.emptyList();
+        }
+
+        pendingChunkedRequests
```
surabujin: should we remove future object from `pendingChunkedRequests`?

services/src/messaging/src/main/java/org/openkilda/messaging/info/ChunkedInfoMessage.java
```diff
@@ -20,21 +20,35 @@
...
+        super(data, timestamp, correlationId);
+        this.messageId = messageId;
+        this.totalMessages = totalMessages;
+    }
+
+    public ChunkedInfoMessage(InfoData data, long timestamp, String correlationId,
+                              int messageIndex, int totalMessages) {
         super(data, timestamp, correlationId);
-        this.nextRequestId = nextRequestId;
+        this.messageId = correlationId + " : " + messageIndex;
```
surabujin: reverse order i.e. messageIndex + ":" + correlationId would be more "usual"... i.e. we are joining correlationIds in this way, so keep same order here looks natural.

services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaMessagingChannel.java
```diff
@@ -0,0 +1,256 @@
...
+     */
+    @Override
+    public CompletableFuture<List<InfoData>> sendAndGetChunked(String topic, Message message) {
+        CompletableFuture<List<InfoData>> future = new CompletableFuture<>();
+        pendingChunkedRequests.put(message.getCorrelationId(), future);
+
+        ListenableFuture<SendResult<String, Message>> futureResult = messageProducer.send(topic, message);
+        futureResult.addCallback(
+                sentResult -> {
+                    messagesChains.put(message.getCorrelationId(), new ArrayList<>());
```
surabujin: To be completely safe, this operation should be done before sending request i.e. in current state there is a possible race condition - if response will be processed before processing of this callbaback.

PR #980, opened 2018-07-06
Remove unused tests code

surabujin:
Remove unused test mocks and utils. Reorganize tests code layout to
match main code layout(packages).


surabujin:
Affect only test code, so no component in labels.

PR #995, opened 2018-07-12
Guaranteed delivery of ISL disco packets

### LINKED ISSUE #986: Implement 'ISL Disco packets guaranty delivery' feature
Currently Event topology might mark ISL as inactive in case when FL event wasn't able to send discovery packets because FL is too overloaded. 
We should prevent it by changing logic in event topology to start "timer" of ISL only when FL has processed message from event topology, sent the command to the switch and notified ET about that.

nikitacherevko:
closes #986


services/wfm/src/test/java/org/openkilda/wfm/isl/DiscoveryManagerTest.java
```diff
@@ -46,103 +59,209 @@
...
-                dstNode2.getSwitchDpId(), dstNode2.getPortId()); // clear attempts
-        dm.makeDiscoveryPlan();                                      // attempts: link1&2 @ 1, link3 @ 2
-        discoveryPlan = dm.makeDiscoveryPlan();                      // the attempts test is based on the 1 & 2 count
-        assertEquals(3, discoveryPlan.needDiscovery.size());
+    @Test
+    public void shouldAlwaysSendDiscoveryForActiveLinks() {
+        setupThreeLinks();
+
+        for (int attempt = 1; attempt <= maxAttemptsLimit + islHealthFailureLimit + 1; attempt++) {
+            if (attempt == 1) {
```
rtretyak: You can `verifyAllLinks(false);` before the loop starts and remove this whole if-else clause from here, only do `verifyAllLinks(true);` on line 131

services/wfm/src/main/java/org/openkilda/wfm/isl/DiscoveryManager.java
```diff
@@ -126,6 +131,10 @@ public Plan makeDiscoveryPlan() {
 
         }
 
+        if (!speakerSendsDiscoPackets) {
+            logger.warn("Speaker does not send discovery packets");
```
surabujin: Add amount (or list if it short) of affected links.

services/wfm/src/main/java/org/openkilda/wfm/isl/DiscoveryManager.java
```diff
@@ -109,6 +110,10 @@ public Plan makeDiscoveryPlan() {
                 // NB: this node can be in both discoveryFailure and needDiscovery
             }
 
+            if (speakerSendsDiscoPackets && link.isAttemptsLimitExceeded(islConsecutiveFailureLimit)) {
```
surabujin: You don't need left part of `&&`

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
```diff
@@ -174,10 +176,16 @@ private Destination getDestinationForTopic(String replyToTopic) {
         }
     }
 
-    private void doDiscoverIslCommand(DiscoverIslCommandData command) {
+    private void doDiscoverIslCommand(CommandMessage message) {
+        DiscoverIslCommandData command = (DiscoverIslCommandData) message.getData();
```
surabujin: Why for move type conversion inside method? The caller must be responsible to supply correct type (in our case caller do "instanceof" and definetely know variable type).

PS CorrelationId can/should be passed as separate argument.

services/src/messaging/src/main/java/org/openkilda/messaging/command/discovery/DiscoverIslCommandData.java
```diff
@@ -18,20 +18,13 @@
...
  * Defines the payload payload of a Message representing an command for ISL discovery.
  */
-@JsonSerialize
-@JsonInclude(JsonInclude.Include.NON_NULL)
-@JsonPropertyOrder({
-        "command",
-        "switch_id",
-        "port_no"})
+@Value
 public class DiscoverIslCommandData extends CommandData {
```
surabujin: You are using NetworkEndpoint as property inside DiscoPacketSendingConfirmation, why not use same approach here?

services/wfm/src/main/java/org/openkilda/wfm/topology/event/OFELinkBolt.java
```diff
@@ -403,6 +407,8 @@ private void dispatchMain(Tuple tuple, InfoMessage infoMessage) {
             passToTopologyEngine(tuple);
         } else if (data instanceof IslInfoData) {
             handleIslEvent(tuple, (IslInfoData) data, infoMessage.getCorrelationId());
+        } else if (data instanceof DiscoPacketSendingConfirmation) {
```
surabujin: DiscoPacketSendingConfirmation messages is much more frequent than other messages in this "dispatcher". It should be first in the "check" list.

services/wfm/src/main/java/org/openkilda/wfm/topology/event/OFELinkBolt.java
```diff
@@ -578,6 +576,11 @@ private void handleMovedIsl(Tuple tuple, String srcSwitch, int srcPort, String d
         passToTopologyEngine(tuple, message);
     }
 
+    private void handleSentDiscoPacket(DiscoPacketSendingConfirmation confirmation) {
+        logger.debug("Discovery packet is sent from {}", confirmation);
```
surabujin: I would like to alter message to something like "Got confirmation - discovery packet from {} have been sent"

Issue #892, opened 2018-06-14
Swagger specification for Northbound is incorrectly dealing with url query params


rtretyak:
Swagger spec for Northbound has various endpoints that incorrectly set query parameters.
For example, the `/api/v1/switches/<switchId>/rules` endpoint, given I set 'cookie' to '123' value, will send request to the `.../rules{?delete-action,cookie,in-port,in-vlan,priority,out-port}?cookie=123` url, which is invalid.
Expected: ability to send requests with query params from Swagger directly, Swagger to form correct urls.

PR #999, opened 2018-07-13
Disable URL templating in swagger Docket bean

phantomii:
Also non required parameters was marked as required = false

Proofs:
 - https://github.com/springfox/springfox/issues/1858
 - https://github.com/springfox/springfox/issues/1848

Task Id: #892

### DIFF
--- a/services/src/northbound/src/main/java/org/openkilda/northbound/config/SwaggerConfig.java
+++ b/services/src/northbound/src/main/java/org/openkilda/northbound/config/SwaggerConfig.java
@@ -38,7 +38,6 @@ public class SwaggerConfig {
     public Docket api() {
         return new Docket(DocumentationType.SWAGGER_2)
                 .apiInfo(apiInfo())
-                .enableUrlTemplating(true)
                 .select()
                 .apis(RequestHandlerSelectors.any())
                 .paths(Predicates.not(PathSelectors.regex("/error.*")))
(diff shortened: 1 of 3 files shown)


services/src/northbound/src/main/java/org/openkilda/northbound/config/SwaggerConfig.java
```diff
@@ -38,7 +38,6 @@
     public Docket api() {
         return new Docket(DocumentationType.SWAGGER_2)
                 .apiInfo(apiInfo())
-                .enableUrlTemplating(true)
                 .select()
                 .apis(RequestHandlerSelectors.any())
                 .paths(Predicates.not(PathSelectors.regex("/error.*")))
```
sergii-iakovenko: Why not use "docket.genericModelSubstitutes(Optional.class);" instead of adding "required = false" to each optional RequestParam?

phantomii: Because it is not works. I tried it and it didn't give expected result.

services/src/northbound/src/main/java/org/openkilda/northbound/controller/SwitchController.java
```diff
@@ -104,7 +104,7 @@ public SwitchFlowEntries getSwitchRules(
             @PathVariable("switch-id") String switchId,
             @ApiParam(value = "Results will be filtered based on matching the cookie.",
                     required = false)
-            @RequestParam("cookie") Optional<Long> cookie) {
+            @RequestParam("cookie, required = false") Optional<Long> cookie) {
```
surabujin: Mess in quoting:

it lead to:
```
http://localhost:8088/api/v1/switches/de%3Aad%3Abe%3Aef%3A00%3A00%3A00%3A07/rules?cookie%2C%20required%20%3D%20false=111222
```

PR #1002, opened 2018-07-13
Floodlight : monitorable and guaranteed processing of incoming messages.

### LINKED ISSUE #942: Investigate incoming kafka messages caching capabilities in FL
There is a change that FL cache/queue incoming kafka requests.

Scenario:
 * make heavy load on FL(massive flows reroute)
 * wait till "big" kafka-lag on TE topic/consumer group appear
 * stop TE, to stop message flow into FL
 * observe FL - it will continue to process "some" input despite 0 kafka-lag on FL/consumer group.

For goal "FL is stateless" we must eliminate any attempt to cache/queue any significant amount of incoming messages.

sergii-iakovenko:
The changes:
- The solution design is provided: [docs/design/solutions/floodlight/floodlight-guaranteed-processing.md](https://github.com/telstra/open-kilda/blob/f15801b91044653f663ac5f1bb494a030c2b9c41/docs/design/solutions/floodlight/floodlight-guaranteed-processing.md).
- FL KafkaMessageCollector has been modified to instantiate the execution thread pool with no work queue.
- FL Kafka consumer has been re-implemented to commit Kafka offsets on per record basis (with configurable interval of actual submitting of the offset).
- Introduce KafkaConsumerConfig and KafkaProducerConfig to separate configuration of the consumer and producer.

Resolves #942


services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/KafkaMessageProducer.java
```diff
@@ -99,30 +94,23 @@ public void postMessage(final String topic, final Message message) {
...
-                        "Invalid value for option %s: %s < 1", option, value));
-            }
-            heartBeat = new HeartBeat(producer, (long) (interval * 1000), topoDiscoTopic);
-        } catch (NumberFormatException e) {
-            throw new FloodlightModuleException(String.format(
-                    "Invalid value for option %s=\"%s\", expect number", option, value));
+    private void initHeartBeat(KafkaProducerConfig producerConfig, String topoDiscoTopic)
+            throws FloodlightModuleException {
+        float interval = producerConfig.getHeartBeatInterval();
+        if (interval < 1) {
```
surabujin: Can/should it be validated on "config level"?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/Consumer.java
```diff
@@ -47,32 +73,38 @@ public void run() {
...
                     }
+
+                    // force to commit after each completed batch.
+                    offsetRegistry.commitOffsets();
+
                     switchManager.safeModeTick(); // HACK alert .. should go in its own timer loop
                 }
+            } catch (InterruptException ex) {
+                // Leave if the thread has been interrupted.
+                throw ex;
```
surabujin: We leave in uncommitted state some set of processed messages. So each FL shutdown will(can) create a gap in commited/processed messages and in worst case lead to repeated processing of this messages.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/KafkaMessageProducer.java
sergii-iakovenko: Good to have it, but conf4j library doesn't provide any validation functionality. And such changes is out of the task scope, so I've introduced support of validation-api for configuration beans in another PR #1013

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/Consumer.java
sergii-iakovenko: Good point. I'll try to address this.

sergii-iakovenko: Fixed.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/Consumer.java
```diff
@@ -81,4 +115,52 @@ protected void handle(ConsumerRecord<String, String> record) {
...
+            lastCommitTime = System.currentTimeMillis();
+        }
+
+        /**
+         * Add the record's offset to the registry and perform a commit
+         * if more than autoCommitInterval ms passed since the last commit.
+         */
+        void addAndCommit(ConsumerRecord<String, String> record) {
+            TopicPartition partition = new TopicPartition(record.topic(), record.partition());
+            partitionToUncommittedOffset.put(partition, record.offset());
```
surabujin: Is the offset always grow? Should be check/filter out rewind attempts?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/Consumer.java
```diff
@@ -81,4 +115,52 @@ protected void handle(ConsumerRecord<String, String> record) {
...
+            if ((System.currentTimeMillis() - lastCommitTime) >= autoCommitInterval) {
+                commitOffsets();
+            }
+        }
+
+        /**
+         * Commits the offsets added since the last commit.
+         */
+        void commitOffsets() {
+            if (!partitionToUncommittedOffset.isEmpty()) {
```
surabujin: excess check

sergii-iakovenko: I don't think so, `consumer.commitSync` even for empty map performs some actions: acquiring/releasing a lock, checking the internal state. Even more -  why should we bother creating an empty map object when we can skip all that?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/Consumer.java
sergii-iakovenko: I'd added the check.

nikitacherevko:
The merge has been reverted due to issues with floodlight consumer

PR #1009, opened 2018-07-16
#1000 Reinvent health check monitoring

### LINKED ISSUE #1000: Fix/reinvent health check monitoring
We have health check endpoint in NB HealthCheckController. It's goal to report each component actual health check.

The problem in current implementation - in case of problem in any compoment(no response), the duration of health check call become extremely high... i.e. caller side give up to wait and drop connection with 'timeout error'.

GOAL: NB must report health check state in reasonable time 5-10 seconds.

Make obsolete #197, #50

IlyaMoiseev:
Closes #1000 
Health-check monitoring architecture proposition and implementation.
Design discussion is in the issue.


services/src/northbound/src/main/java/org/openkilda/northbound/messaging/kafka/KafkaHealthCheckMessageConsumer.java
```diff
@@ -38,13 +39,18 @@
...
 /**
- * Kafka Health-Check message producer.
+ * Kafka Health-Check message consumer.
  */
 public class KafkaHealthCheckMessageConsumer implements HealthCheckMessageConsumer {
     /**
      * Expected topologies count.
      */
-    private static final int HEALTH_CHECK_COMPONENTS_COUNT = 5;
+    private static final int HEALTH_CHECK_COMPONENTS_COUNT = ServiceType.values().length;
```
surabujin: This is no a solution.

* we don't need a separate tooling(KafkaHealthCheckMessageConsumer) for health check messages (see https://github.com/telstra/open-kilda/pull/978 how it should become in near future)
* current health check architecture produce too many false positive responses (any component timeout produce "unhealthy" state of whole system)
* current implementation is too sensitive to kafka timeouts

Prepare architecture proposition (sequence diagram), after it's approval it can be implemented.

PR #1013, opened 2018-07-17
Add support of value validation for configuration beans.

sergii-iakovenko:
The changes:
- Introduce validation-api support for configuration beans produced by FL and WFM configuration provider.

### DIFF
--- a/services/src/configuration/src/main/java/org/openkilda/config/provider/ValidatingConfigurationProvider.java
+++ b/services/src/configuration/src/main/java/org/openkilda/config/provider/ValidatingConfigurationProvider.java
@@ -17,12 +17,12 @@ package org.openkilda.config.provider;
 
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.toSet;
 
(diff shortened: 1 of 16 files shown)


sergii-iakovenko:
This must be merged after #1002!

services/src/configuration/src/main/java/org/openkilda/config/provider/ValidatingConfigurationProvider.java
```diff
@@ -0,0 +1,75 @@
...
+    public <T> T getConfiguration(Class<T> configurationType) {
+        requireNonNull(configurationType, "configurationType cannot be null");
+
+        T instance = factory.createConfiguration(configurationType, source);
+
+        Set<ConstraintViolation<T>> errors = validator.validate(instance);
+        if (!errors.isEmpty()) {
+            String errorDetails = errors.stream()
+                    .map(v -> v.getPropertyPath() + " " + v.getMessage())
+                    .collect(Collectors.joining(";"));
```
surabujin: Maybe "\n" will be better. Or even store is inside exeception as a list and allow called to format/display it in appropriate way.

PR #1021, opened 2018-07-19
Atdd staging/bugfix 1019

### LINKED ISSUE #1019: Update existing delete meters test to only work with OF_13 switches
Scenario: Remove meter on switch
Step: select a random switch and alias it as 'srcSwitch'
^ The above step should be replaced with step that only takes OF_13 switches, because OF_12 switches do not support meters.

npliashechnikov:
Implement additional step for selecting OF_13 switches, include in into testcase.
Closes #1019

### DIFF
--- a/services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/SwitchSteps.java
+++ b/services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/SwitchSteps.java
@@ -64,30 +64,26 @@ public class SwitchSteps {
         assertTrue(allSwitchesResponse.size() >= expectedSwitchesAmount);
     }
 
-    @Given("^select a random switch and alias it as '(.*)'$")
+    @Given("^select a switch and alias it as '(.*)'$")
     public void selectARandomSwitch(String switchAlias) {
         List<Switch> switches = getUnaliasedSwitches();
-        Random r = new Random();
-        Switch theSwitch = switches.get(r.nextInt(switches.size()));
-        log.info("Selected random switch with id: {}", theSwitch.getDpId());
+        Switch theSwitch = switches.get(0);
+        log.info("Selected switch with id: {}", theSwitch.getDpId());
         topologyUnderTest.addAlias(switchAlias, theSwitch);
     }
 
-    @Given("^select a random switch with Openflow version '(.*)' and alias it as '(.*)'$")
+    @Given("^select a switch with Openflow version '(.*)' and alias it as '(.*)'$")
     public void selectARandomSwitchWithSpecificOfVersion(String ofVersion, String switchAlias) {
         List<Switch> switches = getUnaliasedSwitches();
-        List<Switch> ofSwitches = new ArrayList<>();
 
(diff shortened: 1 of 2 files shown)


services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/SwitchSteps.java
```diff
@@ -72,6 +73,23 @@ public void selectARandomSwitch(String switchAlias) {
...
+        List<Switch> ofSwitches = new ArrayList<>();
+
+        for (Switch s: switches) {
+            if (ofVersion.equalsIgnoreCase(s.getOfVersion())) {
+                ofSwitches.add(s);
+            }
+        }
+
+        Random r = new Random();
+        Switch theSwitch = ofSwitches.get(r.nextInt(ofSwitches.size()));
```
surabujin: I have always thought that random input data is inacceptable for testing. Because it eliminate repeatability.

rtretyak: That's true. This happens due to the fact that we do not hardcode our topology, because we are not creating the required topology from test (the physical one) and depend on config. This means that topology can be changed at any time and our tests will not be affected.
This, unfortunately, denies us from using concrete switch ids in tests, but, on the other hand, allows our tests to be run on any topology without changing/recompiling the code. In perfect scenario such methods should return the 'same' switch every time, so that test itself won't see the difference. Does it make sense?

sergii-iakovenko: Agree with @surabujin. We can just order `switches` by switchId and pick the first one.

npliashechnikov: Actually, randomness of switch selection allowed us for the eventual reliability - we were testing all switch makes over time. :)
However, you're totally correct about issues with the reproduceability, tests should definitely be deterministic. I have opened additional issue so that we can reach both goals simultaneously.

services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/SwitchSteps.java
```diff
@@ -63,15 +64,28 @@ public void verifySwitchesAmount(int expectedSwitchesAmount) {
...
+        List<Switch> switches = getUnaliasedSwitches();
+
+        for (Switch s: switches) {
+            if (ofVersion.equalsIgnoreCase(s.getOfVersion())) {
+                log.info("Selected switch with id: {}", s.getDpId());
+                topologyUnderTest.addAlias(switchAlias, s);
+                return;
+            }
+        }
+        log.error("No switches found with Openflow version {}", ofVersion);
```
rtretyak: You should throw instead of log, because we should always stop execution at this point.
You can also call `Assume.assumeTrue(false, "msg")` in order to mark scenario as 'skipped' instead of failed (if you want so)

PR #1024, opened 2018-07-20
Add validation of flow ports against conflicts with ISL ports.

### LINKED ISSUE #183: Check the in / out ports and ensure they are not an ISL
When the user submits a flow create request, we have to validate that the inport and the outport are not ISLs ... ie the Ingress port and the egress port can't be ISL ports. Otherwise you could create a loop.

Some hints on implementation:
1) the potential right place to implement the filter, during flow creation request, is in the neo4j driver/topology-engine -> augment the path search in such a way that it returns "no path"; ie filter out the ingress/egress ports if they are an ISL. 
2) add logic to the PCE to check whether the ports are isl ports .. this could be used in conjunction with step #1 so that the right error can be returned to the user.  Or, it can be used instead of #1 (ie perform the check on all flow create requests, rather than on failures from #1.

The benefit of doing just #2 is that the query logic won't have to be changed. The downside is that every flow create request will take a bit longer.

sergii-iakovenko:
The changes:
- Add check for conflicts with ISL ports in FlowValidator.
- Introduce TopologyRepository interface (NeoDriver implements it) which declares operations on Topology entities in a storage.
- Update ATDD tests to have flows not using ISL ports.

Resolves #183

### DIFF
--- a/services/src/northbound/src/main/java/org/openkilda/northbound/service/impl/FlowServiceImpl.java
+++ b/services/src/northbound/src/main/java/org/openkilda/northbound/service/impl/FlowServiceImpl.java
@@ -711,8 +711,8 @@ public class FlowServiceImpl implements FlowService {
          * 3) Do the comparison
          */
 
-        List<Flow> flows = pathComputer.getFlow(flowId);
-        if (flows == null) {
+        List<Flow> flows = pathComputer.getFlows(flowId);
+        if (flows == null || flows.isEmpty()) {
             return null;
         }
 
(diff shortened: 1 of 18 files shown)


services/src/northbound/src/main/java/org/openkilda/northbound/service/impl/FlowServiceImpl.java
```diff
@@ -711,8 +711,8 @@ public String toString() {
          * 3) Do the comparison
          */
 
-        List<Flow> flows = pathComputer.getFlow(flowId);
-        if (flows == null) {
+        List<Flow> flows = pathComputer.getFlows(flowId);
```
surabujin: pathComputer.getFlows do not return null

PR #1033, opened 2018-07-23
#1038: Add cleanup step to 'Remove meter on switch' scenario

ylobankov:
"Remove meter on switch" scenario doesn't have a cleanup step to remove
the flow after test execution. This patch adds the missing step.

Closes: #1038


surabujin:
Until you put label "ready to merge" no one will merge and review(in most cases) your PR.

PR #1047, opened 2018-07-25
Fixed failsafe test to only pick switches with traffgens

rtretyak:


services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/service/flowmanager/FlowManagerImpl.java
```diff
@@ -71,7 +72,9 @@
     public Map<FlowPayload, List<TopologyDefinition.Isl>> createFlowsWithASwitch(int flowsAmount,
                                                                                  int alternatePaths, int bandwidth) {
 
-        final List<TopologyDefinition.Switch> switches = topologyDefinition.getActiveSwitches();
+        List<TopologyDefinition.TraffGen> traffGens = topologyDefinition.getActiveTraffGens();
```
surabujin: Will it use switch port with attached traffgen?

```
FlowPayload flow = flowSet.buildWithAnyPortsInUniqueVlan(flowId, srcSwitch, dstSwitch, bandwidth);
```
mentioned `any` port.

rtretyak: hz, let me check

rtretyak: Nice catch. It by default used to build a flow between port 10 on both sides, which by accident matched the actual traffgen ports. Updated to intentionally select traffgen ports.

PR #1048, opened 2018-07-26
Add per flow periodic ping activation switch

surabujin:
To be able to control how many flows are targeted by periodic pings and to be able to exclude not important flows from periodic pings add `periodic-pings` boolean flag into flow definition.

Only if this flag is set, flow will be periodically pinged.


services/src/atdd/src/test/java/org/openkilda/atdd/FlowCrudBasicRunTest.java
```diff
@@ -80,13 +81,17 @@ public void successfulFlowCreation(final String flowId, final String sourceSwitc
...
                 new FlowEndpointPayload(destinationSwitch, destinationPort, destinationVlan),
-                bandwidth, false, flowId, null, FlowState.UP.getState());
+                bandwidth, false, false, flowId, null, FlowState.UP.getState());
 
         FlowPayload response = FlowUtils.putFlow(flowPayload);
         assertNotNull(response);
-        response.setLastUpdated(null);
 
-        assertEquals(flowPayload, response);
+        System.out.println(String.format(
```
abochkarev: It looks like we need to think about logging...

services/src/messaging/src/test/java/org/openkilda/messaging/info/flow/FlowInfoDataTest.java
```diff
@@ -0,0 +1,86 @@
...
+                forwardFlowThread.getFlowId(),
+                new ImmutablePair<>(forwardFlowThread, reverseFlowThread),
+                FlowOperation.CREATE, "unit-test-correlation-id");
+
+        InfoMessage wrapper = new InfoMessage(origin, System.currentTimeMillis(), origin.getCorrelationId());
+        serialize(wrapper);
+        FlowInfoData decoded = (FlowInfoData) ((InfoMessage) deserialize()).getData();
+
+        Assert.assertEquals(
+                String.format("%s object have been mangled in serialisation/deserialization loop",
```
abochkarev: ...object **has** been ...

services/src/atdd/src/test/java/org/openkilda/atdd/FlowCrudBasicRunTest.java
surabujin: We definitely should.

But there is a plan to move out from atdd a lot of test that actually not acceptance tests. So there is a sense to "clean" such issues after planned reorganization.

services/src/messaging/src/test/java/org/openkilda/messaging/info/flow/FlowInfoDataTest.java
surabujin: changing

Issue #1031, opened 2018-07-23
2 ATDD scenarios failed. AssertionError: expected:<2> but was:<11>


phantomii:
git commit: 533aa9eab

how to reproduce:
```
make build-base
make compile
docker-compose build
make up-test-mode
make atdd
```
result:
```
Failed scenarios:
FlowCrudBasicRun.feature:39 # Scenario Outline: Flow Creation on Small Linear Network Topology
TopologyDiscoveryBasic.feature:73 # Scenario: Ignore not signed LLDP packets

42 Scenarios (2 failed, 40 passed)
479 Steps (2 failed, 10 skipped, 467 passed)
33m56.565s

java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertFalse(Assert.java:64)
	at org.junit.Assert.assertFalse(Assert.java:74)
	at org.openkilda.atdd.FlowCrudBasicRunTest.checkFlowCreation(FlowCrudBasicRunTest.java:115)
	at ✽.Then flow c3none with de:ad:be:ef:00:00:00:03 1 0 and de:ad:be:ef:00:00:00:05 2 0 and 10000 could be created(FlowCrudBasicRun.feature:21)

java.lang.AssertionError: expected:<2> but was:<11>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at org.openkilda.topo.TestUtils.validateTopos(TestUtils.java:48)
	at org.openkilda.atdd.TopologyDiscoveryBasicTest.the_controller_learns_the_topology(TopologyDiscoveryBasicTest.java:71)
	at ✽.And the controller learns the topology(TopologyDiscoveryBasic.feature:79)

[ERROR] Tests run: 521, Failures: 4, Errors: 0, Skipped: 10, Time elapsed: 2,038.263 s <<< FAILURE! - in org.openkilda.atdd._RunCucumberTest
[...]

PR #1052, opened 2018-07-26
Decrease parallelism of opentsdb topology on dev env

surabujin:
To decrease RAM requirement for dev enviroment, decrease parallelism of
opentsdb topology. This is the only topology that start in paralle on
devel environment.


phantomii:
Kilda began to use less resources but unfortunately this patch is not help me with issue #1031.

PR #1062, opened 2018-07-27
#1061: Extend atdd-staging scenario in failsafe suite

### LINKED ISSUE #1061: [atdd-staging] Extend 'ISL connectivity is lost, system is able to react in expected way and reroute after some timeout' scenario
**Current scenario:**
**Given** Create 1 flow with A Switch used and at least 1 alternate path between source and destination switch and 500 bandwidth

**When** ISL between switches loses connectivity
**And** remains in this state for 30 seconds
**Then** ISL status is DISCOVERED
**And** ISL status changes to FAILED
**And** flow is in UP state
**And** flow is valid per Northbound validation
**And** all active switches have correct rules installed per Northbound validation
**And** flow has traffic going with bandwidth not less than 450 and not greater than 550

**When** Changed ISL restores connectivity
**Then** ISL status changes to DISCOVERED
**And** flow is in UP state
**And** flow is valid per Northbound validation
**And** all active switches have correct rules installed per Northbound validation
**And** flow has traffic going with bandwidth not less than 450 and not greater than 550
**And** each flow can be deleted

It is needed to extend this scenario by adding extra steps for checking flow paths. 

**The extended scenario will look in such a way:**
**Given** Create 1 flow with A Switch used and at least 1 alternate path between source and destination switch and 500 bandwidth

**When** ISL between switches loses connectivity
**And** remains in this state for 30 seconds
**Then** ISL status is DISCOVERED
**And** ISL status changes to FAILED
**And** flow is in UP state
**And** flow path is changed  **<-------------------- This step is going to be added**
**And** flow is valid per Northbound validation
**And** all active switches have correct rules installed per Northbound validation
**And** flow has traffic going with bandwidth not less than 450 and not greater than 550
**When** Changed ISL restores connectivity
**Then** ISL status changes to DISCOVERED
**And** flow is in UP state
**And** flow path is unchanged  **<-------------------- This step is going to be added**
**And** flow is valid per Northbound validation
**And** all active switches have correct rules installed per Northbound validation
**And** flow has traffic going with bandwidth not less than 450 and not greater than 550
**And** each flow can be deleted

ylobankov:
This patch is intended to extend "ISL connectivity is lost, system is able to react 
in expected way and reroute after some timeout" scenario by adding extra steps 
for checking flow paths. 

Current scenario:
Scenario: ISL connectivity is lost, system is able to react in expected way and reroute after some timeout

Given Create 1 flow with A Switch used and at least 1 alternate path between source and destination switch and 500 bandwidth

When ISL between switches loses connectivity
And remains in this state for 30 seconds
Then ISL status is DISCOVERED
And ISL status changes to FAILED
And flow is in UP state
And flow is valid per Northbound validation
And all active switches have correct rules installed per Northbound validation
And flow has traffic going with bandwidth not less than 450 and not greater than 550

When Changed ISL restores connectivity
Then ISL status changes to DISCOVERED
And flow is in UP state
And flow is valid per Northbound validation
And all active switches have correct rules installed per Northbound validation
And flow has traffic going with bandwidth not less than 450 and not greater than 550
And each flow can be deleted

It is needed to extend this scenario by adding extra steps for checking flow paths.

The extended scenario will look in such a way:
Scenario: ISL connectivity is lost, system is able to react in expected way and reroute after some timeout

Given Create 1 flow with A Switch used and at least 1 alternate path between source and destination switch and 500 bandwidth

When ISL between switches loses connectivity
And remains in this state for 30 seconds
Then ISL status is DISCOVERED
And ISL status changes to FAILED
And flow is in UP state
And flow path is changed <-------------------- This step is going to be added
And flow is valid per Northbound validation
And all active switches have correct rules installed per Northbound validation
And flow has traffic going with bandwidth not less than 450 and not greater than 550
When Changed ISL restores connectivity
Then ISL status changes to DISCOVERED
And flow is in UP state
And flow path is unchanged <-------------------- This step is going to be added
And flow is valid per Northbound validation
And all active switches have correct rules installed per Northbound validation
And flow has traffic going with bandwidth not less than 450 and not greater than 550
And each flow can be deleted

Closes #1061

### DIFF
--- a/services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/service/flowmanager/FlowManagerImpl.java
+++ b/services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/service/flowmanager/FlowManagerImpl.java
@@ -128,7 +128,7 @@ public class FlowManagerImpl implements FlowManager {
         if (!foundEnoughFlows) {
             result.keySet().forEach(f -> northboundService.deleteFlow(f.getId()));
         }
-        Assume.assumeTrue("Didn't find enough of requested flows. This test cannot be run on given topology."
+        Assume.assumeTrue("Didn't find enough of requested flows. This test cannot be run on given topology. "
                 + "Do you have enough a-switch links in the topology?", foundEnoughFlows);
         return result;
     }
--- a/services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/FlowCrudSteps.java
+++ b/services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/FlowCrudSteps.java
@@ -23,9 +23,12 @@ import static org.hamcrest.Matchers.contains;
 import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.hamcrest.Matchers.empty;
 import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.everyItem;
 import static org.hamcrest.Matchers.hasKey;
 import static org.hamcrest.Matchers.hasProperty;
 import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.isIn;
+import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.notNullValue;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
@@ -84,6 +87,7 @@ import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -119,6 +123,7 @@ public class FlowCrudSteps implements En {
     private TopologyUnderTest topologyUnderTest;
 
     private Set<FlowPayload> flows;
+    private Set<FlowPathPayload> flowPaths = new HashSet<>();
     private FlowPayload flowResponse;
 
     @Given("^flows defined over active switches in the reference topology$")
@@ -138,6 +143,24 @@ public class FlowCrudSteps implements En {
         flowIsls.putAll(flowManager.createFlowsWithASwitch(flowsAmount, alternatePaths, bw));
         //temporary resaving flows before refactoring all methods to work with topologyUnderTest
         flows = flowIsls.keySet();
+        flows.forEach(flow -> flowPaths.add(northboundService.getFlowPath(flow.getId())));
+    }
+
+    @And("^flow paths? (?:is|are) changed")
+    public void flowPathIsChanged() {
+        Set<FlowPathPayload> actualFlowPaths = new HashSet<>();
+        flows.forEach(flow -> actualFlowPaths.add(northboundService.getFlowPath(flow.getId())));
+        assertThat(actualFlowPaths, everyItem(not(isIn(flowPaths))));
+
+        // Save actual flow paths
+        flowPaths = actualFlowPaths;
+    }
+
+    @And("^flow paths? (?:is|are) unchanged")
+    public void flowPathIsUnchanged() {
+        Set<FlowPathPayload> actualFlowPaths = new HashSet<>();
+        flows.forEach(flow -> actualFlowPaths.add(northboundService.getFlowPath(flow.getId())));
+        assertThat(actualFlowPaths, everyItem(isIn(flowPaths)));
     }
 
     @And("Create defined flows?")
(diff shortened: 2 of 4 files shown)


services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/FlowCrudSteps.java
```diff
@@ -138,6 +142,24 @@ public void flowsWithAlternatePaths(int flowsAmount, int alternatePaths, int bw)
         flowIsls.putAll(flowManager.createFlowsWithASwitch(flowsAmount, alternatePaths, bw));
         //temporary resaving flows before refactoring all methods to work with topologyUnderTest
         flows = flowIsls.keySet();
+        flows.forEach(flow -> flowPaths.add(northboundService.getFlowPath(flow.getId())));
```
rtretyak: You cannot get flows' path because flows are not created at this point yet. This should break the FlowCrud scenario

services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/FlowCrudSteps.java
```diff
@@ -138,6 +142,24 @@ public void flowsWithAlternatePaths(int flowsAmount, int alternatePaths, int bw)
...
+    }
+
+    @And("^flow paths? (?:is|are) changed")
+    public void flowPathIsChanged() {
+        Set<FlowPathPayload> actualFlowPaths = new HashSet<>();
+        flows.forEach(flow -> actualFlowPaths.add(northboundService.getFlowPath(flow.getId())));
+        assertThat(actualFlowPaths, everyItem(not(isIn(flowPaths))));
```
rtretyak: This is not actually true. This may work in some cases, but some parts of the path may still remain the same after reroute, it is not guaranteed that _everyItem_ will be different

services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/FlowCrudSteps.java
```diff
@@ -138,6 +142,24 @@ public void flowsWithAlternatePaths(int flowsAmount, int alternatePaths, int bw)
...
+
+        // Save actual flow paths
+        flowPaths = actualFlowPaths;
+    }
+
+    @And("^flow paths? (?:is|are) unchanged")
+    public void flowPathIsUnchanged() {
+        Set<FlowPathPayload> actualFlowPaths = new HashSet<>();
+        flows.forEach(flow -> actualFlowPaths.add(northboundService.getFlowPath(flow.getId())));
+        assertThat(actualFlowPaths, everyItem(isIn(flowPaths)));
```
rtretyak: Probably ok. But still, not verifying the sequence. The path should be literally the same, not just 'contain same items'.

services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/FlowCrudSteps.java
```diff
@@ -119,6 +122,7 @@
     private TopologyUnderTest topologyUnderTest;
 
     private Set<FlowPayload> flows;
+    private Set<FlowPathPayload> flowPaths = new HashSet<>();
```
rtretyak: Set is an **unordered** collection. How do you bind those paths to exact flows at this point? Moreover, if there are two flows with the same paths, it will swallow the path for second flow because Set cannot have duplicates. So you will end up with 2 flows and 1 flow path. What if there is a third flow. 3 flows 2 paths, what next?)

services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/FlowCrudSteps.java
ylobankov: It looks like a flow is already created at this point. We are creating the flow at L142.

services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/FlowCrudSteps.java
ylobankov: Ok. Then we need to check only paths, not all items.

services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/FlowCrudSteps.java
rtretyak: oh right, I thought it was added in the `flows defined over active switches in the reference topology` step

services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/service/flowmanager/FlowManagerImpl.java
```diff
@@ -17,6 +17,7 @@
 
 import static java.lang.String.format;
 
+import org.junit.Assume;
```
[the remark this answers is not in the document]
ylobankov: Ok, will fix it.

Issue #976, opened 2018-07-03
Replace '_' to '-' in the names of Kilda services.


abochkarev:
It's required to replace '_' to '-' in the names of Kilda services.

**How to reproduce:**
1) cd open-kilda;
2) docker-compose up;
3) stop docker compose with Ctrl+C;
4) check with which service names '_' is used;

**Actual:**
Stopping topology-engine-rest    ... done
Stopping topology-engine         ... done
Stopping floodlight              ... done
Stopping mininet                 ... done
Stopping **storm_ui**                ... done
Stopping **storm_supervisor**        ... done
Stopping **storm_nimbus**            ... done
Stopping opentsdb                ... done
Stopping **open-kilda_northbound_1** ... done
Stopping kafka                   ... done
Stopping hbase                   ... done
Stopping logstash                ... done
Stopping kibana                  ... done
Stopping neo4j                   ... done
Stopping elasticsearch           ... done
Stopping zookeeper               ... done

**Expected:**
Stopping topology-engine-rest    ... done
Stopping topology-engine         ... done
Stopping floodlight              ... done
Stopping mininet                 ... done
Stopping **storm-ui**                ... done
Stopping **storm-supervisor**        ... done
Stopping **storm-nimbus**            ... done
Stopping opentsdb                ... done
Stopping **open-kilda-northbound** ... done
Stopping kafka                   ... done
Stopping hbase                   ... done
Stopping logstash                ... done
Stopping kibana                  ... done
Stopping neo4j                   ... done
Stopping elasticsearch           ... done
Stopping zookeeper               ... done

pzakatov:
done

PR #1070, opened 2018-07-30
Replace '_' to '-' in the names of Kilda services. #976

phantomii:


abochkarev:
I will check today, @phantomii

abochkarev:
I've checked this patch on my machine. It seems everything works as expected. @phantomii Thanks. @nikitamarchenko Please review also.

Issue #1075, opened 2018-07-31
Build process changes to include Openkilda-gui


gauravchug:
Changes in build process to incorporate  kilda-gui instance at same time when other containers build.
Earlier kilda gui was build separately.

gauravchug:
Fixed #1077

pzakatov:
closed

PR #1077, opened 2018-07-31
Build process changes to include Openkilda-gui 

gauravchug:
#1075


docker-compose.yml
```diff
@@ -372,6 +372,26 @@ services:
          - opentsdb.pendev
 
 
+  openkilda-gui:
```
surabujin: `openkilda-` prefix is excess here. Whole this file is for `open-kilda`. And docker-compose use name of directory where docker-compose.yml as `project name`...

docker-compose.yml
```diff
@@ -372,6 +372,26 @@ services:
...
+    container_name: openkilda-gui
+    build:
+      context: services/src/openkilda-gui
+    image: "kilda/openkilda-gui:${full_build_number:-latest}"
+    ports:
+      - "1010:1010"
```
surabujin: Usage of port number smaller than 1024 is a bad idea.

docker-compose.yml
```diff
@@ -372,6 +372,26 @@ services:
          - opentsdb.pendev
 
 
+  openkilda-gui:
+    container_name: openkilda-gui
+    build:
+      context: services/src/openkilda-gui
+    image: "kilda/openkilda-gui:${full_build_number:-latest}"
```
surabujin: We do not produce docker image `kilda/openkilda-gui`.

You need or "build" or "image".

docker-compose.yml
gauravchug: Its actually our project name `openkilda-gui`

docker-compose.yml
gauravchug: I followed the same steps like we did for others (floodlight, wfm, neo4j)

docker-compose.yml
ashishaggarwalCS: This port is used from the starting of this project and if we change this now we have to notify users who all are using kilda gui.
Ideally when using GUI port should not be visible.

surabujin: I do not ask to change "production" TCP port, I am asking to not use ports numbers smaller than 1024 in dev/local env.

docker-compose.yml
surabujin: and?

docker-compose.yml
surabujin: Is it a separate project? or part of some other project?

gauravchug: It is a separate project under open-kilda like NB and FL.

docker-compose.yml
gauravchug: Ok. For GUI it is better to run on 80 port so that no one need to add additional port information while opening the GUI. 
Port 80 was consumed by TPE service, once TPE shift to another port will change 1010 to 80.

docker-compose.yml
gauravchug: What you refer as Build or Image. Please suggest

docker-compose.yml
surabujin: One more time - I din't suggest to change "production application port", I have asked to change "development port" and do not use reserved/common used port for it.

docker-compose.yml
surabujin: External/foreign services are build into images into Makefile, our services are build by docker-compose.

services/src/openkilda-gui/docker-compose.yml
```diff
@@ -3,4 +3,4 @@ services:
   openkilda_gui:
```
surabujin: Do you still need separate docker-compose.yml?

services/src/openkilda-gui/README.md
```diff
@@ -11,13 +11,13 @@ This project holds the OPEN KILDA GUI service for Open Kilda Controller.
 #### Configuration
 
 * The **_openkilda-gui_** feature must be installed in OPEN KILDA.
-* The GUI listens on port 1010
```
surabujin: your service still use port 1010, if this document states how to configure openkilda-gui, it must refer port 1010. If this document refer how to use local devel environment you should describe steps how to run it using docker-compose and/or make commands.

services/src/openkilda-gui/docker-compose.yml
gauravchug: No need, i believe.
removing

services/src/openkilda-gui/README.md
gauravchug: Yep, I already defined server port is 1010 but when we build using docker-compose or make command then port 8010 listens.

PR #1083, opened 2018-07-31
Added 'Can't resolve 'kafka.pendev' host' message to makefile

phantomii:
Kafka.pendev host should be resolved on local machine. if the host is not resolved on your local machine, the correct error message will be printed before any tests ran.


phantomii:
Guys this patch is not actual for review or other actions.

Issue #1079, opened 2018-07-31
Topology Engine has no safe fallback mechanism if receive bad response from Neo


rtretyak:
**Steps:**
1. Create a flow with `'` sign in flow description. This will reveal the #1051 issue and create an invalid DB request
1.1. Alternative way: use steps described in #1076
**Expected:** eventhough tpe ended up in error, it should be able to continue operating instead of infinitely trying to fail-retry.
**Actual:** TPE is 'dead' and requires manual healing.

pzakatov:
@surabujin , please verify this ticket

npliashechnikov:
Verified - the issue can no longer be reproduced on current develop branch

Issue #907, opened 2018-06-19
TE hangs for couple minutes if send link property with ` character


rtretyak:
1. send link property with ``` ` ``` character in key ```"props": { "`cost": 123 }```

**Expected:** Any readable error of what went wrong, error received immediately
**Actual:** NB times out waiting for response from the system. Any subsequent calls with any valid data to update link properties are also dead.
System restores after couple of minutes

rtretyak:
no longer valid

PR #1084, opened 2018-08-01
Remove upper py2neo version limit

surabujin:
Newer py2neo version doen't have a problem with neo4j error response
handling. It allow us to remove some set of crutches added in TE to
overcome this problem.


surabujin:
Relates issues:
* #1079
* #907

PR #1090, opened 2018-08-01
Create a way to delete a switch from Kilda cache

### LINKED ISSUE #1074: Create a way to delete a switch from Kilda cache
This must be implemented in Cache topology.
Cache topology must read the CommandMessage from Kafka and delete the switch from CacheBolt.

A part of #328 Create a way to delete switch

Settis:
Closes #1074 issue.

### DIFF
--- a/services/src/messaging/src/main/java/org/openkilda/messaging/command/switches/SwitchDeleteRequest.java
+++ b/services/src/messaging/src/main/java/org/openkilda/messaging/command/switches/SwitchDeleteRequest.java
@@ -1,0 +1,43 @@
+/* Copyright 2018 Telstra Open Source
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+package org.openkilda.messaging.command.switches;
+
+import org.openkilda.messaging.command.CommandData;
+import org.openkilda.messaging.model.SwitchId;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.NonNull;
+import lombok.Value;
+
+@Value
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class SwitchDeleteRequest extends CommandData {
+
+    @JsonProperty("switch_id")
+    private SwitchId switchId;
+
+    /**
+     * Constructs a delete switch request.
+     *
+     * @param switchId switch id to delete rules from.
+     */
+    @JsonCreator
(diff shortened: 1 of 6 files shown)


services/src/messaging/src/main/java/org/openkilda/messaging/info/event/SwitchState.java
```diff
@@ -96,18 +99,11 @@ public String toString() {
...
-                .filter(item -> item.type.equalsIgnoreCase(state))
-                .findFirst()
-                .orElseThrow(() -> new IllegalArgumentException(String.format("Incorrect switch state: %s", state)));
-    }
-
     public boolean isActive() {
         return this == ADDED || this == ACTIVATED;
     }
 
     public boolean isInactive() {
```
surabujin: Dangerous combination - 2 method that is supposed to return opposite values, but calculate it in different way. Can we drop one of this(`isActive`, `isInactive`) methods?

services/wfm/src/main/java/org/openkilda/wfm/topology/cache/CacheBolt.java
```diff
@@ -257,6 +257,10 @@ private void handleSwitchEvent(SwitchInfoData sw, Tuple tuple, String correlatio
                 //emitRerouteCommands(affectedFlows, tuple, correlationId, FlowOperation.UPDATE, reason);
                 break;
 
+            case DELETED:
```
surabujin: I extremely don't like the idea to use `SwitchInfoData` for this operation. `SwitchInfoData` represent (at least now) hardware events produced by FL. Adding this new not hardware state mixes our "control" logic with hardware signals.

You need to create separate set of messages to control switch.

services/src/messaging/src/main/java/org/openkilda/messaging/info/event/SwitchState.java
Settis: Good point. I was also thinking about it. In fact `isInactive` method is not used in the project and can be deleted. As you can see, we have 3 states (`CHANGED`, `VALIDATING` & `CACHED`) which not belongs to active or inactive. This is not clear. In general, I can remove `isInactive` method.

services/wfm/src/main/java/org/openkilda/wfm/topology/cache/CacheBolt.java
Settis: Make sense. I'll change it.

services/src/messaging/src/main/java/org/openkilda/messaging/info/event/SwitchState.java
surabujin: do it.

Settis:
Well, you are right. Let me done #1102 first. After we will have skeleton for unit testing, I can proceed with this pull.

services/src/messaging/src/main/java/org/openkilda/messaging/command/switches/SwitchDeleteRequest.java
```diff
@@ -0,0 +1,47 @@
...
+    @JsonProperty("switch_id")
+    private String switchId;
+
+    /**
+     * Constructs a delete switch request.
+     *
+     * @param switchId switch id to delete rules from.
+     */
+    @JsonCreator
+    public SwitchDeleteRequest(@JsonProperty("switch_id") String switchId) {
```
surabujin: You can delegate "null check" to the lombok - add `@NonNull` here.

Settis:
Sonar can check only non forked repositories. I've recreated this pull request with local branch #1215. Let's proceed with this new pull request.

services/src/messaging/src/main/java/org/openkilda/messaging/command/switches/SwitchDeleteRequest.java
Settis: Done.

Issue #1087, opened 2018-08-01
Package net-tools is not available, but is referred to by another package.


phantomii:
commit: 7df9d8dcea903dd75b34c457aa17ab54ee9b3748 (latest develop branch)

```
Building mininet
Step 1/30 : FROM kilda/base-ubuntu
 ---> 66d4e71693f5
Step 2/30 : ENV DEBIAN_FRONTEND noninteractive
 ---> Using cache
 ---> 60623834bd5b
Step 3/30 : ENV MININET_REPO https://github.com/mininet/mininet.git
 ---> Using cache
 ---> 9f3ab6c4dfa6
Step 4/30 : ENV MININET_INSTALLER mininet/util/install.sh
 ---> Using cache
 ---> 39861cb9cc70
Step 5/30 : ENV INSTALLER_SWITCHES -fbinptvwyx
 ---> Using cache
 ---> dd157c2c599e
Step 6/30 : RUN apt-get update
 ---> Using cache
 ---> 72da6144c405
Step 7/30 : RUN apt-get -qy install net-tools
 ---> Running in 0017c342968b
Reading package lists...
Building dependency tree...
Reading state information...
Package net-tools is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
```

PR #1098, opened 2018-08-02
Reformat docker files  #1087

phantomii:
apt-get XXX commands must be in one RUN context. See best practices in https://docs.docker.com/v17.09/engine/userguide/eng-image/dockerfile_best-practices/#run RUN section.

### DIFF
--- a/base/base-builder/Dockerfile
+++ b/base/base-builder/Dockerfile
@@ -19,10 +19,15 @@ FROM ubuntu:yakkety
 
 ENV DEBIAN_FRONTEND noninteractive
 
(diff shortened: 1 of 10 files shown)


surabujin:
As far as I know, we do not use
* `base/base-builder`
* `base/base-java`
* `base/base-maven`

Need to recheck and drop them completely.

phantomii:
Thanks @surabujin. I think it is first patch of series. I faced with ineffective images building. I think I can speed up building images. It help us to speed up "pull requests" testing and so on.

Issue #659, opened 2018-04-23
500 on create flow in case "Path was not found"


pzakatov:
We need be more specific on this error:

- Code: 404 Not Found. 
- Error message: "Can't create flow. Not enough bandwidth found or path not found"

npliashechnikov:
Verified - now Kilda generates HTTP 404 on both "Create Flow" and "Update Flow" API calls, although the error messages are slightly different:
Create flow: Could not create flow: Not enough bandwidth found or path not found
Update flow: Could not update flow: Path was not found

PR #1112, opened 2018-08-06
Fix path was not found issue

Levitan:
Issue link - https://github.com/telstra/open-kilda/issues/659


services/src/northbound/src/main/java/org/openkilda/northbound/utils/NorthboundExceptionHandler.java
```diff
@@ -65,6 +65,11 @@
                 break;
             case OPERATION_TIMED_OUT:
             case INTERNAL_ERROR:
+                status = HttpStatus.INTERNAL_SERVER_ERROR;
+                break;
+            case CREATION_FAILURE:
```
sergii-iakovenko: #659 requests to change the status code for UnroutablePathException only, but this code changes affect all CREATION_FAILURE errors.

PR #1131, opened 2018-08-08
Drop unused ancient project components

surabujin:


services/src/tools/mininet/README.md
```diff
@@ -1,69 +0,0 @@
-# Mininet Tools
```
sergii-iakovenko: Does it make sense to move this documentation to services/mininet instead of deleting? As we have no documentation there and this one may be still actual.

surabujin: This "tool" is an REST client for mininet REST wrapper (I don't check is it match with current state of mininent REST wrapper). And documentation describe how to use it...

We don't use it(REST client) anywhere. That why it is in "remove list". And because this documentation describe how to use this client only, there is no reason to keep it.

PR #1143, opened 2018-08-13
Enabled mandatory run update-props as part of make targets

PinGwynn:
All currents (already templated) properties files moved as examples.
Added run update-props as a dependency of "run-dev" and "up-test-mode" make targets.

For now, only one place for management properties files is templates. All local changes will be overwritten as part of update-props.


.travis.yml
```diff
@@ -10,6 +10,12 @@ jobs:
         - make update-props-dryrun
     - stage: unit
       language: java
+      before_install:
+        - sudo pip install --upgrade pip
+        - sudo pip install ansible
+      #  - sudo apt-get update
```
abochkarev: Just to clarify: do we really need to have commented out commands in the travis.yml?

PinGwynn: Thank you. Comments have been removed

base/hacks/load_kilda_dev_ubuntu.sh
```diff
@@ -85,7 +85,7 @@ echo ##  PYTHON ETC
 echo ##########################
 
 sudo apt-get update
-sudo apt-get install -y apt-utils python python-pip
```
surabujin: Where is it(this whole file) used?

PinGwynn: This file uses for prepare local dev environment. It installs all necessary stuff, like java, python, compose, etc...

surabujin: i.e. it should be started manually to prepare local env(linux ubuntu) to build open-kilda?

Can't find its mentions in docs/wiki (actually I can't find its mentions anywhere).

PR #1152, opened 2018-08-16
Deny creation of self-looped ISL

### LINKED ISSUE #1059: Deny creation of self-looped ISL
Customer equipment attached to our ports can return our discovery packet back to our switch... It can be a loop on:
- different port on same switch
- same port (i.e. port used to send discovery package)
- different switch

We can easily detect first 2 cases. The last case should be covered into https://github.com/telstra/open-kilda/issues/579.

So we must prevent creation of self-looped ISLs.

IlyaMoiseev:
Closes #1059
Cover cases with loop on the same port and between two ports


services/wfm/src/main/java/org/openkilda/wfm/topology/event/OfeLinkBolt.java
```diff
@@ -504,6 +504,12 @@ private void handleIslEvent(Tuple tuple, IslInfoData discoveredIsl, String corre
          *  one place.
          */
         if (IslChangeType.DISCOVERED.equals(state)) {
+            if (discoveredIsl.isSelfLooped()) {
```
IlyaMoiseev: Is needed to remove such looped link from discovery for some time?

services/wfm/src/test/java/org/openkilda/wfm/topology/event/OfeLinkBoltTest.java
```diff
@@ -132,4 +137,28 @@ public void shouldNotResetDiscoveryStatusOnSync() throws JsonProcessingException
                         hasProperty("destination", hasProperty("datapath", is(new SwitchId("ff:02")))),
                         hasProperty("active", is(true)))));
     }
+
+    @Test
+    public void shouldNotProcessLoopedIsl() throws JsonProcessingException {
+        final SwitchId switchId = new SwitchId("00:01");
```
abochkarev: If I'm not mistaken now we may omit final in tests.

services/wfm/src/main/java/org/openkilda/wfm/topology/event/OfeLinkBolt.java
nikitamarchenko: no

rtretyak:
I updated mininet rest api here to allow topologies with self-loops. This makes testing it locally an easier task

rtretyak:
Self-loop ISLs will create pretty much noise in logs. Is that expected?
image
@nikitamarchenko @IlyaMoiseev

rtretyak:
I added a basic test to see if the self-loop is not going to be discovered. This should be enough for now

nikitacherevko:
Could you please resolve conflicts? And then add "Ready to Merge" label again

PR #1160, opened 2018-08-20
Update README.md

abochkarev:
- added the instruction for installing ansible;
- made changes to commands for building Kilda controller;


README.md
```diff
@@ -17,13 +17,18 @@ virtualenv --python=python2 .venv
 pip install tox
 ```
 
+You also need to install ansible. For that purpose execute the following command:
+
+```
+sudo apt install ansible
```
abochkarev: Maybe we need to omit this instruction. As was @ylobankov noticed it won't work for Mac.

PR #1172, opened 2018-08-21
[WFM] fix checkstyle suppress.

Settis:


Settis:
Sonar can check only non forked repositories. I've recreated this pull request with local branch #1177. Let's proceed with this new pull request.

PR #1177, opened 2018-08-21
[WFM] fix checkstyle suppress.

Settis:


abochkarev:
@nikitamarchenko @surabujin Colleagues, what do you think about merging this patch? I've looked at it and seems it resolves the most of checkstyle suppressings in WFM. As an exceptional case, I would merge it, just to not return to fixing suppressing in WFM (except javadocs, of course).

PR #1179, opened 2018-08-21
Changed external NB port for all-in-one env from 8088 to 8080

PinGwynn:


templates/defaults/main.yaml
```diff
@@ -23,7 +23,7 @@ topology_engine_endpoint: "http://127.0.0.1"
 topology_engine_rest_endpoint: "http://127.0.0.1"
 topology_engine_rest_port: "80"
 northbound_endpoint: "http://127.0.0.1"
```
gauravchug: Change `127.0.0.1` to `northbound.pendev`

templates/defaults/main.yaml
```diff
@@ -22,8 +22,8 @@ mininet_rest_port: "38080"
 topology_engine_endpoint: "http://127.0.0.1"
 topology_engine_rest_endpoint: "http://127.0.0.1"
 topology_engine_rest_port: "80"
-northbound_endpoint: "http://127.0.0.1"
-northbound_rest_port: "8088"
+northbound_endpoint: "http://northbound.pendev"
```
abochkarev: @PinGwynn Am I correct that 'northbound.pendev' should be resolved in /etc/hosts for the local env? If so, would it be possible to write about in readme file? Also, how about the other endpoints, I mean: openkildagui_host, floodlight_endpoint, mininet_endpoint, topology_engine_endpoint, topology_engine_rest_endpoint. Are you planning to replace them with the DNS?

PinGwynn: No, this endpoint and variable use in kilda gui. But as we have added it under the docker-compose control, we have to change endpoints to "internal" docker-compose names.

abochkarev: @PinGwynn Ok, I got it.

PR #1181, opened 2018-08-22
Fix duplication of ISL records in DiscoveryManager.

sergii-iakovenko:
The changes:
- Update DiscoveryManager to avoid duplication of ISLs on port registration.

### DIFF
--- a/services/src/messaging/src/main/java/org/openkilda/messaging/ctrl/state/OFELinkBoltState.java
+++ b/services/src/messaging/src/main/java/org/openkilda/messaging/ctrl/state/OFELinkBoltState.java
@@ -1,15 +1,15 @@
 package org.openkilda.messaging.ctrl.state;
 
+import org.openkilda.messaging.ctrl.AbstractDumpState;
+import org.openkilda.messaging.ctrl.state.visitor.DumpStateVisitor;
+import org.openkilda.messaging.model.DiscoveryLink;
+
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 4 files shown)


services/src/messaging/src/main/java/org/openkilda/messaging/ctrl/state/OFELinkBoltState.java
```diff
@@ -35,7 +35,7 @@ public void accept(DumpStateVisitor visitor) {
         visitor.visit(this);
     }
 
-    public List<?> getDiscovery() {
+    public Set<?> getDiscovery() {
```
abochkarev: Would it be possible to return Set<DiscoveryLink> here instead of Set<?>?

services/wfm/src/test/java/org/openkilda/wfm/isl/DiscoveryManagerTest.java
```diff
@@ -337,31 +342,28 @@ public void handleSwitchUp() {
         dm.handleDiscovered(srcNode3.getDatapath(), srcNode3.getPortNumber(),
                 dstNode3.getDatapath(), dstNode3.getPortNumber());
 
-        List<DiscoveryLink> links;
+        Set<DiscoveryLink> links;
```
abochkarev: We could define links in one line:
`Set<DiscoveryLink> links = dm.findAllBySwitch(srcNode1.getDatapath());`

services/src/messaging/src/main/java/org/openkilda/messaging/ctrl/state/OFELinkBoltState.java
sergii-iakovenko: Fixed.

services/wfm/src/test/java/org/openkilda/wfm/isl/DiscoveryManagerTest.java
sergii-iakovenko: Fixed.

Issue #1085, opened 2018-08-01
New tests: Flow bandwidth


rtretyak:
Introduce new test in functional-tests module (or set of tests) that will verify:
- [x] Available bandwidth on ISLs changes respectively to bandwidth of created flows through those ISLs
- [x]  In case of not enough available bandwidth on the shortest path, longer path is chosen (longer/shorter depends on summary cost of all ISLs of the path)
- [x]  Not able to create flow with bandwidth exceeding the minimum available bandwidth amount for flow's ISLs (some ISLs should have higher bw than the others, system should only allow flows that are not higher than the minimum bw)
- [x]  Able to overexceed bandwidth limit when creating flow if use 'ignore_bandwidth: true'
- [x]  Verify bandwidth restrictions for Update flow operation
- [x]  verify case #919
- [x]  user should not be able to reroute to a more preferable path if that path has not enough bandwidth (#1092)
- [x]  user should be able to reroute to a more preferable path if it has enough bandwidth (#1154)

PR #1184, opened 2018-08-22
Add test to verify available bandwidth on ISLs

ylobankov:
The idea of the test is to verify available bandwidth on ISLs that changes
respectively when creating a flow.

Part of #1085.

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/PathHelper.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/PathHelper.groovy
@@ -58,10 +58,10 @@ class PathHelper {
             def src = path[i - 1]
             def dst = path[i]
             def involvedIsl = topology.isls.find {
-                (it.srcSwitch.dpId == src.switchId && it.srcPort == src.portNo && it.dstPort == dst.portNo &&
-                        it.dstSwitch.dpId == dst.switchId) ||
-                        (it.dstSwitch.dpId == src.switchId && it.dstPort == src.portNo && it.srcPort == dst.portNo &&
-                                it.srcSwitch.dpId == dst.switchId)
+                (it.srcSwitch?.dpId == src.switchId && it?.srcPort == src.portNo &&
+                        it.dstPort == dst.portNo && it.dstSwitch.dpId == dst.switchId) ||
+                        (it.dstSwitch?.dpId == src.switchId && it?.dstPort == src.portNo &&
+                                it.srcPort == dst.portNo && it.srcSwitch.dpId == dst.switchId)
             } ?: Isl.factory(topology.switches.find { it.dpId == src.switchId },
                     src.portNo, topology.switches.find { it.dpId == dst.switchId }, dst.portNo, 0, null)
             involvedIsls << involvedIsl
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/BandwidthSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/BandwidthSpec.groovy
@@ -8,8 +8,10 @@ import org.openkilda.messaging.payload.flow.FlowState
 import org.openkilda.testing.model.topology.TopologyDefinition
 import org.openkilda.testing.model.topology.TopologyDefinition.Switch
 import org.openkilda.testing.service.northbound.NorthboundService
+import org.openkilda.testing.service.topology.TopologyEngineService
 import org.openkilda.testing.tools.IslUtils
 import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.web.client.HttpClientErrorException
 
 class BandwidthSpec extends BaseSpecification {
 
@@ -19,6 +21,8 @@ class BandwidthSpec extends BaseSpecification {
     FlowHelper flowHelper
     @Autowired
     NorthboundService northboundService
+    @Autowired
+    TopologyEngineService topologyEngineService
     @Autowired
     PathHelper pathHelper
     @Autowired
@@ -58,6 +62,55 @@ class BandwidthSpec extends BaseSpecification {
                 def bwAfterFlow = islUtils.getIslInfo(linksAfterFlow, it).availableBandwidth
                 bwAfterFlow == bwBeforeFlow - maximumBandwidth
             }
+        }
+
+        cleanup: "Delete created flow"
+        flow?.id && northboundService.deleteFlow(flow.id)
+    }
+
+    def "Unable to exceed bandwidth limit on ISL when creating a flow"() {
+        given: "Two active switches"
+        def switches = topology.getActiveSwitches()
+        def (Switch srcSwitch, Switch dstSwitch) = [switches, switches].combinations()
+                .find { Switch src, Switch dst -> src != dst }
+        assert srcSwitch && dstSwitch
+
+        when: "Create a flow with a bandwidth that exceeds available bandwidth on ISL"
+        def flow = flowHelper.randomFlow(srcSwitch, dstSwitch)
+        def possibleFlowPaths = topologyEngineService.getPaths(srcSwitch.dpId, dstSwitch.dpId)*.path
+        def allLinks = northboundService.getAllLinks()
+        def involvedBandwidths = []
+
+        possibleFlowPaths.each { path ->
+            pathHelper.getInvolvedIsls(path).each { link ->
+                involvedBandwidths.add(islUtils.getIslInfo(allLinks, link).availableBandwidth)
+            }
+        }
+        flow.maximumBandwidth = involvedBandwidths.max() + 1
+        flow.ignoreBandwidth = false
+        northboundService.addFlow(flow)
+
+        then: "Flow is not created because flow path should not be found"
+        def exc = thrown(HttpClientErrorException)
+        exc.rawStatusCode == 404
+    }
+
+    def "Able to exceed bandwidth limit on ISL when creating a flow with ignore_bandwidth = true"() {
+        given: "Two active switches"
+        def switches = topology.getActiveSwitches()
+        def (Switch srcSwitch, Switch dstSwitch) = [switches, switches].combinations()
+                .find { Switch src, Switch dst -> src != dst }
+        assert srcSwitch && dstSwitch
+
+        when: "Create a flow with a bandwidth that exceeds available bandwidth on ISL (ignore_bandwidth = true)"
+        def flow = flowHelper.randomFlow(srcSwitch, dstSwitch)
+        flow.maximumBandwidth = Integer.MAX_VALUE
+        flow.ignoreBandwidth = true
+        northboundService.addFlow(flow)
+
+        then: "Flow is successfully created"
+        Wrappers.wait(5) {
+            northboundService.getFlowStatus(flow.id).status == FlowState.UP
         }
 
         cleanup: "Delete created flow"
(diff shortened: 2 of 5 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/BandwidthSpec.groovy
```diff
@@ -0,0 +1,73 @@
...
+        def maximumBandwidth = 1000
+        List<List<PathNode>> allPaths = []
+
+        def switches = topology.getActiveSwitches()
+        def (Switch srcSwitch, Switch dstSwitch) = [switches, switches].combinations()
+                .findAll { src, dst -> src != dst }.find { Switch src, Switch dst ->
+            allPaths = topologyEngineService.getPaths(src.dpId, dst.dpId)*.path
+            allPaths.size() > 1
+        }
+        srcSwitch && dstSwitch
```
rtretyak: not a verification

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/BandwidthSpec.groovy
```diff
@@ -0,0 +1,73 @@
...
+    def "Available bandwidth on ISLs changes respectively when creating a flow"() {
+        when: "Create flow with a valid bandwidth"
+        def maximumBandwidth = 1000
+        List<List<PathNode>> allPaths = []
+
+        def switches = topology.getActiveSwitches()
+        def (Switch srcSwitch, Switch dstSwitch) = [switches, switches].combinations()
+                .findAll { src, dst -> src != dst }.find { Switch src, Switch dst ->
+            allPaths = topologyEngineService.getPaths(src.dpId, dst.dpId)*.path
+            allPaths.size() > 1
```
rtretyak: Why do you need a src-dst switch pair that have more than 1 alternative path? This should not matter in this test

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/BandwidthSpec.groovy
```diff
@@ -0,0 +1,73 @@
...
+        def flow = flowHelper.randomFlow(srcSwitch, dstSwitch)
+        def links = northboundService.getAllLinks()
+        def forwardLink = links.find() {
+            it.path.first().switchId == flow.source.datapath &&
+                    it.path.last().switchId == flow.destination.datapath
+        }
+        def reverseLink = links.find {
+            it.path.first().switchId == flow.destination.datapath &&
+                    it.path.last().switchId == flow.source.datapath
+        }
```
rtretyak: The above two finds does not consider ports. two switches can have more than one link (this even was an actual case on Staging for some time)

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/BandwidthSpec.groovy
```diff
@@ -0,0 +1,73 @@
...
+        def availableBandwidthForward = forwardLink.availableBandwidth
+        def availableBandwidthReverse = reverseLink.availableBandwidth
+
+        flow.maximumBandwidth = maximumBandwidth
+        northboundService.addFlow(flow)
+
+        then: "Flow is really created and has 'Up' status"
+        northboundService.getFlowStatus(flow.id).status == FlowState.UP
```
rtretyak: You'll have to perform some kind of 'wait' operation here, because flow becomes Up not immediately

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/BandwidthSpec.groovy
rtretyak: How do you guarantee that the two selected switches will be neighboring switches? You algorithm will only work if 
 srcSwitch and dstSwitch will have no transit switches in between (will have a direct link in between), which is not guaranteed by you switch selection steps.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/BandwidthSpec.groovy
```diff
@@ -0,0 +1,73 @@
...
+            it.path.first().switchId == flow.destination.datapath &&
+                    it.path.last().switchId == flow.source.datapath
+        }
+
+        actualForwardLink.availableBandwidth == availableBandwidthForward - maximumBandwidth
+        actualReverseLink.availableBandwidth == availableBandwidthReverse - maximumBandwidth
+        
+        cleanup: "Delete created flow"
+        flow?.id && northboundService.deleteFlow(flow.id)
+    }
```
rtretyak: Please look into "Should be able to reroute to a better path if it has enough bandwidth" where I check the availableBandwidth after reroute. Your case should perform pretty much the same logic but for Create/Update operations (not reroute). What I recommend to do:
- Pick up 2 switches (it actually does not matter which switches to pick, so probably just pick 2 first switches or 2 last switches or whatever, but not random to ensure repeatability)
- get all available paths for these switches by calling `topologyEngineService.getPaths`
- create a method in PathHelper that will take a path as input and will return the actual **cost** of that path
- given you have such tool you can now find a path that will have the lowest cost (Kilda will select this path when building a flow)
- Given you have potential path that will be used, you can now call `def involvedIsls = pathHelper.getInvolvedIsls(myPath)` to get a list of ISLs involved in your path.
- ^ This will be a list of our **internal ISL representations**(from topology.yaml). In order to get actual ISL state (from NB) for each of these ISLs you'll have to call `islUtils.getIslInfo(myInternalIslRepresentation)`. So that it becomes:
```
def allLinks = northbound.getAllLinks()
def actualLinksUsedInOurPath = involvedIsls.collect { islUtils.getIslInfo(allLinks, it) }
```
- you can see the current bandwidth of all these isls: `println actualLinksUsedInOurPath*.availableBandwidth`
^this is you current state before building a flow
--------------------------------------
- Build a flow for your selected switches
- Call `PathHelper.convert(northboundService.getFlowPath(flow.id))` to get the Path which was chosen for your flow by Kilda
- you can compare now that the chosen path matches the path you have predicted above (matches the path with lowest cost)
- now rerun the getAllLinks code and recheck the updated availableBandwidth matches your expectations

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/PathHelper.groovy
```diff
@@ -60,7 +60,7 @@ class PathHelper {
             def involvedIsl = topology.isls.find {
                 (it.srcSwitch.dpId == src.switchId && it.srcPort == src.portNo && it.dstPort == dst.portNo &&
```
rtretyak: Should probably use safe navigation here too

ylobankov: Ok, let's add it here too.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/BandwidthSpec.groovy
```diff
@@ -63,4 +64,35 @@ class BandwidthSpec extends BaseSpecification {
...
+        given: "Two active switches"
+        def switches = topology.getActiveSwitches()
+        def (Switch srcSwitch, Switch dstSwitch) = [switches, switches].combinations()
+                .find { Switch src, Switch dst -> src != dst }
+        assert srcSwitch && dstSwitch
+
+        when: "Create a flow with a bandwidth that exceeds available bandwidth on ISL (ignore_bandwidth = false)"
+        def flow = flowHelper.randomFlow(srcSwitch, dstSwitch)
+        def maxAvailableBw = (northboundService.getAllLinks()*.availableBandwidth).max()
+        flow.maximumBandwidth = maxAvailableBw + 100
```
rtretyak: why +100 and not +1?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/BandwidthSpec.groovy
```diff
@@ -63,4 +64,35 @@ class BandwidthSpec extends BaseSpecification {
         cleanup: "Delete created flow"
         flow?.id && northboundService.deleteFlow(flow.id)
     }
+
+    def "Able to exceed bandwidth limit on ISL when creating a flow"() {
```
rtretyak: ?
Maybe 'Unable'?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/BandwidthSpec.groovy
```diff
@@ -63,4 +64,35 @@ class BandwidthSpec extends BaseSpecification {
...
+        given: "Two active switches"
+        def switches = topology.getActiveSwitches()
+        def (Switch srcSwitch, Switch dstSwitch) = [switches, switches].combinations()
+                .find { Switch src, Switch dst -> src != dst }
+        assert srcSwitch && dstSwitch
+
+        when: "Create a flow with a bandwidth that exceeds available bandwidth on ISL (ignore_bandwidth = false)"
+        def flow = flowHelper.randomFlow(srcSwitch, dstSwitch)
+        def maxAvailableBw = (northboundService.getAllLinks()*.availableBandwidth).max()
```
rtretyak: This is not 100% accurate. You pick max bandwidth of the whole system, while in fact you should select maxBandwidth across all available paths

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/BandwidthSpec.groovy
```diff
@@ -63,4 +64,35 @@ class BandwidthSpec extends BaseSpecification {
...
+        flow.maximumBandwidth = maxAvailableBw + 100
+        flow.ignoreBandwidth = false
+        northboundService.addFlow(flow)
+
+        then: "Flow is not created because flow path should not be found"
+        def exc = thrown(HttpClientErrorException)
+        exc.rawStatusCode == 404
+
+        when: "Create a flow with a bandwidth that exceeds available bandwidth on ISL (ignore_bandwidth = true)"
```
rtretyak: This looks like a separate test

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/BandwidthSpec.groovy
```diff
@@ -63,4 +67,53 @@ class BandwidthSpec extends BaseSpecification {
...
+        flow.maximumBandwidth = involvedBandwidths.max() + 1
+        flow.ignoreBandwidth = false
+        northboundService.addFlow(flow)
+
+        then: "Flow is not created because flow path should not be found"
+        def exc = thrown(HttpClientErrorException)
+        exc.rawStatusCode == 404
+    }
+
+    def "Able to exceed bandwidth limit on ISL when creating a flow"() {
```
rtretyak: This name is not corresponding with the previos test name. You cannot have two tests
`Unable to exceed bandwidth limit on ISL when creating a flow`
`Able to exceed bandwidth limit on ISL when creating a flow`
Not clear what is the expected system behavior. You should specify that this scenario uses ignoreBandwidth as an exceptional case

PR #1189, opened 2018-08-22
Added test for reroute behavior when link fails and no alt paths available

### LINKED ISSUE #1174: New test: flows should go down when link fails and there's no ability to reroute
To be added to functional-tests module.

rtretyak:
Closes #1174

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/AutoReroutesSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/AutoReroutesSpec.groovy
@@ -1,0 +1,65 @@
+package org.openkilda.functionaltests.spec
+
+import org.openkilda.functionaltests.BaseSpecification
+import org.openkilda.functionaltests.helpers.FlowHelper
+import org.openkilda.functionaltests.helpers.PathHelper
+import org.openkilda.functionaltests.helpers.Wrappers
+import org.openkilda.messaging.model.SwitchId
+import org.openkilda.messaging.payload.flow.FlowState
+import org.openkilda.northbound.dto.switches.PortDto
+import org.openkilda.testing.model.topology.TopologyDefinition
+import org.openkilda.testing.model.topology.TopologyDefinition.Switch
+import org.openkilda.testing.service.northbound.NorthboundService
+import org.openkilda.testing.service.topology.TopologyEngineService
+import org.springframework.beans.factory.annotation.Autowired
+
+class AutoReroutesSpec extends BaseSpecification {
+    @Autowired
+    TopologyDefinition topology
+    @Autowired
+    TopologyEngineService topologyEngineService
+    @Autowired
+    FlowHelper flowHelper
+    @Autowired
+    PathHelper pathHelper
+    @Autowired
+    NorthboundService northboundService
+
+    def flowStatusTimeout = 7
+
+    def "Flow should go Down when its link fails and there is no ability to reroute"() {
+        given: "A flow"
+        def (Switch srcSwitch, Switch dstSwitch) = topology.getActiveSwitches()[0..1]
+        def allPaths = topologyEngineService.getPaths(srcSwitch.dpId, dstSwitch.dpId)*.path
+        def flow = flowHelper.randomFlow(srcSwitch, dstSwitch)
+        northboundService.addFlow(flow)
+        def currentPath = PathHelper.convert(northboundService.getFlowPath(flow.id))
+
+        and: "Ports that lead to alternative paths are brought down to deny alternative paths"
+        def altPaths = allPaths.findAll { it != currentPath }
+        List<PortDto> broughtDownPorts = altPaths.collect { path ->
+            def src = path.first()
+            def dst = path.last()
+            [northboundService.portDown(src.switchId, src.portNo),
+             northboundService.portDown(dst.switchId, dst.portNo)]
+        }.flatten()
+
+        when: "One of the flow's ISLs goes down"
+        def isl = pathHelper.getInvolvedIsls(currentPath).first()
+        northboundService.portDown(isl.dstSwitch.dpId, isl.dstPort)
+
+        then: "Flow becomes 'Down'"
+        Wrappers.wait(flowStatusTimeout) { northboundService.getFlowStatus(flow.id).status == FlowState.DOWN }
+
+        when: "ISL goes back up"
+        northboundService.portUp(isl.dstSwitch.dpId, isl.dstPort)
+
+        then: "Flow becomes 'Up'"
+        Wrappers.wait(flowStatusTimeout) { northboundService.getFlowStatus(flow.id).status == FlowState.UP }
+
+        and: "Restore topology to original state, remove flow"
+        broughtDownPorts.each { northboundService.portUp(new SwitchId(it.switchId), it.portNumber) }
+        northboundService.deleteFlow(flow.id)
+        //TODO(rtretiak): restore costs that were changed due to portdowns
+    }
+}
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/IntentionalRerouteSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/IntentionalRerouteSpec.groovy
@@ -33,7 +33,7 @@ class IntentionalRerouteSpec extends BaseSpecification {
         def switches = topology.getActiveSwitches()
         List<List<PathNode>> allPaths = []
         def (Switch srcSwitch, Switch dstSwitch) = [switches, switches].combinations()
-                .findAll { src, dst -> src != dst }.find { Switch src, Switch dst ->
+                .findAll { src, dst -> src != dst }.unique { it.sort() }.find { Switch src, Switch dst ->
             allPaths = topologyEngineService.getPaths(src.dpId, dst.dpId)*.path
             allPaths.size() > 1
         }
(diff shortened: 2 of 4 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/Wrappers.groovy
```diff
@@ -0,0 +1,93 @@
...
+
+import groovy.util.logging.Slf4j
+import org.junit.runners.model.MultipleFailureException
+
+@Slf4j
+class Wrappers {
+
+    /**
+     * Retries some operation for a given number of times. Retry is considered failed if any throwable is thrown.
+     * Throws MultipleFailureException if closure fails to execute in given amount of tries.
```
abochkarev: You could move `Throws MultipleFailureException if closure fails to execute in given amount of tries.` to the javadocs block @throws.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/Wrappers.groovy
```diff
@@ -0,0 +1,93 @@
...
+        while (retries++ < times) {
+            try {
+                return body.call()
+            } catch (Throwable t) {
+                ex.add(t)
+                handler.call(t)
+                sleep(retryInterval * 1000 as long)
+            }
+        }
+        throw new MultipleFailureException(ex)
```
abochkarev: It could be great to add some custom message here: like throw new MultipleFailureException("<Error short description>", ex)

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/Wrappers.groovy
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * Wait for closure to return a result. Returns false if success condition was not met.
+     *
+     * @param timeout time to wait in seconds
+     * @param retryInterval pause between retries (in seconds)
+     * @param closure : A closure that returns following:
+     *          - True if check is successful
+     *          - False or null if check is unsuccessful
+     * @return True if wait was successful, false otherwise
```
abochkarev: Are you planning to add something in the @throws section?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/Wrappers.groovy
```diff
@@ -0,0 +1,93 @@
...
+     * @param name some name of closure/code block that is being executed
+     * @param closure code block to benchmark
+     * @return result of the given closure
+     */
+    static def benchmark(name, closure) {
+        def start = System.currentTimeMillis()
+        def result = closure.call()
+        def now = System.currentTimeMillis()
+        log.debug("$name took " + (now - start) + "ms")
+        result
```
abochkarev: Sometimes uses return, sometimes not. Was it done specially?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/Wrappers.groovy
```diff
@@ -0,0 +1,93 @@
...
+        int retries = 0
+
+        List<Throwable> ex = []
+        while (retries++ < times) {
+            try {
+                return body.call()
+            } catch (Throwable t) {
+                ex.add(t)
+                handler.call(t)
+                sleep(retryInterval * 1000 as long)
```
abochkarev: Would it be helpful to use` TimeUnit.SECONDS.sleep(2) ` here?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/AutoReroutesSpec.groovy
```diff
@@ -0,0 +1,65 @@
...
+    @Autowired
+    TopologyEngineService topologyEngineService
+    @Autowired
+    FlowHelper flowHelper
+    @Autowired
+    PathHelper pathHelper
+    @Autowired
+    NorthboundService northboundService
+
+    def flowStatusTimeout = 7
```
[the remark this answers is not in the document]
rtretyak: I don't think that this should be done. This timeout is not configurable on Kilda side either. This operation should happen 'as soon as possible', so I just picked a reasonable value that I find acceptable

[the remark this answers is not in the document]
abochkarev: Yup.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/Wrappers.groovy
rtretyak: In Groovy 'return' keyword is optional, but I agree that for such large blocks it is better to explicitly mention it. Will add

abochkarev: Yes, I know. Just to ask)

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/Wrappers.groovy
rtretyak: this method never throws

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/Wrappers.groovy
rtretyak: The default org.junit.runners.model.MultipleFailureException implementation don't have such constructor. Don't think it worth to create one

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/Wrappers.groovy
abochkarev: Hah, yes. You're catching all throwables. Agree. We don't need this section here.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/Wrappers.groovy
abochkarev: Up to you)

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/Wrappers.groovy
rtretyak: This won't accept fractional number (say the default 0.5 value will become invalid)

abochkarev: There are also MILLISECONDS, MICROSECONDS, NANOSECONDS. Maybe something from it could be helpful?

abochkarev: In fact, it doesn't matter how to "sleep", it just improves readability.

rtretyak: If I change it to MILLISECONDS, then
1. The unit will differ from 'timeout' unit (which is Seconds) and it will be better to change it too
2. The caller will have to specify both values in milliseconds, which is more verbose, less intuitive and at some point less readable (I prefer to think of timeouts/intervals in seconds)

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/Wrappers.groovy
```diff
@@ -0,0 +1,93 @@
...
+                    sleep(sleepTime)
+                }
+            } catch (Throwable t) {
+                log.debug("Wait check threw exception: ${t.message}.\nSleeping for ${sleepTime} ms...")
+                sleep(sleepTime)
+            }
+        }
+        log.debug("Closure failed to succeed within ${timeout} seconds, aborting.")
+        return false
+    }
```
npliashechnikov: It'd be nice to include another wait() method which would fail early if we've met a failure condition, like this:

`/**
--
* Wait for closure to return a result. Returns false if success condition was not met due to error or timeout.
*
* @param timeout time to wait in seconds
* @param retryInterval pause between retries (in seconds)
* @param closure: A tri-state closure that returns following:
*          - Object or true if check is successful
*          - False if action is failed and it doesn't make sense to wait anymore
*          - Null if success/failure criteria is not yet met
* @return Closure call result if wait was successful, false otherwise
*/
def static waitTristate(double timeout, double retryInterval = 0.5, Closure closure) {
long endTime = System.currentTimeMillis() + (long)(timeout * 1000)
long sleepTime = (long)(retryInterval * 1000)
while(System.currentTimeMillis() < endTime) {
try {
def result = closure.call()
if(result) {
return result
} else if (result == false) { //Checking if closure returned false or null.
log.debug("An error occured during wait, aborting.")
return false
} else {
log.debug("No wait results yet. Sleeping for ${sleepTime} ms...")
sleep(sleepTime)
}
} catch(Throwable t) {
log.debug("Wait check threw exception: ${t.message}.\nSleeping for ${sleepTime} ms...")
sleep(sleepTime)
}
}
log.debug("Closure failed to succeed within ${timeout} seconds, aborting.")
return false
}`

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/Wrappers.groovy
abochkarev: Np, Roman. I just give suggestions which could be helpful. If n * 1000 is more readable - I'm ok)

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/Wrappers.groovy
rtretyak: already discussed, not going to do here

PR #1201, opened 2018-08-27
[WFM/Cache topology] Add Reroute throttling bolt.

### LINKED ISSUE #1192: Do not reissue same reroute for a flow for some time
Multiple reroutes on a flow within a short time period cause race condition in installation of switch rules. As the result the flow ends up in non-operation condition and a number of missing / excessive rules are left on switches.

In order to prevent the race condition, we need to combine reroute commands issued within the configured time window, so they're never performed concurrently:
- FlowRerouteRequest which going to CrudBolt from NB or CacheBolt must pass a new step - reduction of repeat requests.
- Regardless of the origin and cause of FlowRerouteRequest, we combine all commands for the same flowId into a one and hold on unless the configured time is passed from the previous reroute command.

#### The solution based on tick tuples and StatefulBolt ####
- A new bolt is introduced in Flow topology. The bolt intercepts requests coming to CrudBolt.
- The bolt checks whether another FlowRerouteRequest for the same ID passed within the reduction time window (just an in-memory map of flowId to timestamp). And if it's so then the bolt schedules the command to be sent at the end of the time window. Scheduled commands are kept in memory (the bolt state).
-  The bolt checks the scheduled commands periodically (E.g. on a tick tuple) and sends them if it's time.

#### The solution based on Storm's  [IWindowedBolt](http://storm.apache.org/releases/2.0.0-SNAPSHOT/Windowing.html) ####
- A new bolt (extends BaseStatefulWindowedBolt) is introduced in Flow topology. The bolt intercepts requests coming to CrudBolt.
- The bolt is configured to slide the window every 1 second (or less), and the length of window equals to the reduction time window.
- The bolt process only tuples with FlowRerouteRequest partitioned by flowId. 
- The bolt checks whether the window contains FlowRerouteRequest with correlationId equals to saved in memory (a map of flowId to correlationId). And doesn't emit any tuple until the FlowRerouteRequest with saved correlationId is out of the window.
- The bolt emits FlowRerouteRequest when the previous one goes off the window. And saves the correlationId of the last FlowRerouteRequest seen in the window.

Settis:
Closes #1192

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/AbstractTopologyConfig.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/AbstractTopologyConfig.java
@@ -35,6 +35,9 @@ public interface AbstractTopologyConfig {
     @Key("parallelism")
     int getParallelism();
 
+    @Key("parallelism.new")
+    int getNewParallelism();
+
     @Key("workers")
     int getWorkers();
 
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/cache/CacheBolt.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/cache/CacheBolt.java
@@ -21,9 +21,6 @@ import static org.openkilda.messaging.Utils.MAPPER;
 import org.openkilda.messaging.BaseMessage;
 import org.openkilda.messaging.Destination;
 import org.openkilda.messaging.Message;
-import org.openkilda.messaging.Utils;
-import org.openkilda.messaging.command.CommandMessage;
-import org.openkilda.messaging.command.flow.FlowRerouteRequest;
 import org.openkilda.messaging.ctrl.AbstractDumpState;
 import org.openkilda.messaging.ctrl.state.CacheBoltState;
 import org.openkilda.messaging.ctrl.state.FlowDump;
@@ -53,13 +50,13 @@ import org.openkilda.wfm.ctrl.ICtrlBolt;
 import org.openkilda.wfm.share.utils.PathComputerFlowFetcher;
 import org.openkilda.wfm.topology.AbstractTopology;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.google.common.annotations.VisibleForTesting;
 import org.apache.storm.state.InMemoryKeyValueState;
 import org.apache.storm.task.OutputCollector;
 import org.apache.storm.task.TopologyContext;
 import org.apache.storm.topology.OutputFieldsDeclarer;
 import org.apache.storm.topology.base.BaseStatefulBolt;
+import org.apache.storm.tuple.Fields;
 import org.apache.storm.tuple.Tuple;
 import org.apache.storm.tuple.Values;
 import org.slf4j.Logger;
@@ -76,6 +73,8 @@ public class CacheBolt
         extends BaseStatefulBolt<InMemoryKeyValueState<String, Cache>>
         implements ICtrlBolt {
     public static final String STREAM_ID_CTRL = "ctrl";
+    public static final String FLOW_ID_FIELD = "flowId";
+    public static final String CORRELATION_ID_FIELD = "correlationId";
 
     /**
      * Network cache key.
@@ -222,7 +221,7 @@ public class CacheBolt
     @Override
     public void declareOutputFields(OutputFieldsDeclarer output) {
         output.declareStream(StreamType.TPE.toString(), AbstractTopology.fieldMessage);
-        output.declareStream(StreamType.WFM_DUMP.toString(), AbstractTopology.fieldMessage);
+        output.declareStream(StreamType.WFM_REROUTE.toString(), new Fields(FLOW_ID_FIELD, CORRELATION_ID_FIELD));
         output.declareStream(StreamType.OFE.toString(), AbstractTopology.fieldMessage);
         // FIXME(dbogun): use proper tuple format
         output.declareStream(STREAM_ID_CTRL, AbstractTopology.fieldMessage);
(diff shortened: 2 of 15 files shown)


services/wfm/src/main/java/org/openkilda/wfm/topology/cache/FlowThrottlingBolt.java
```diff
@@ -0,0 +1,86 @@
...
+            } catch (JsonProcessingException exception) {
+                logger.error("Could not format flow reroute request by flow={}", flowId, exception);
+            }
+        }
+        _collector.ack(tuple);
+    }
+
+    @Override
+    protected void doWork(Tuple tuple) {
+        reroutesThrottling.putRequest(tuple.getString(0), tuple.getString(1));
```
nikitacherevko: It would be more obviously if you load value from the tuple by name instead of depending on the position in the tuple.

services/wfm/src/main/java/org/openkilda/wfm/topology/cache/service/ReroutesThrottling.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * Keeps current reroute request. Only last correlationId will be saved.
+     *
+     * @param flowId the flow ID
+     * @param correlationId the correlation ID
+     */
+    public void putRequest(String flowId, String correlationId) {
+        reroutes.put(flowId, correlationId);
```
nikitacherevko: Would you mind to add some logs here in order to detect the case when we override some requests. In other words it would be useful to see in logs what happened with reroute request in case when we just ignore it here (some requests might be overridden by a newer one)

services/wfm/src/main/java/org/openkilda/wfm/topology/cache/FlowThrottlingBolt.java
Settis: Fixed.

services/wfm/src/main/java/org/openkilda/wfm/topology/cache/service/ReroutesThrottling.java
Settis: Fixed.

services/wfm/src/main/java/org/openkilda/wfm/topology/cache/service/ReroutesThrottling.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public class ReroutesThrottling {
+
+    private static final Logger logger = LoggerFactory.getLogger(ReroutesThrottling.class);
+
+    private ExtendableTimeWindow extendableTimeWindow;
+
+    private Map<String, String> reroutes = new HashMap<>();
```
sergii-iakovenko: The map is a candidate to be kept in the bolt state. Otherwise we should mention that we lose the requests on worker failure or topology restart.

services/wfm/src/main/java/org/openkilda/wfm/topology/cache/service/ReroutesThrottling.java
```diff
@@ -0,0 +1,77 @@
...
+package org.openkilda.wfm.topology.cache.service;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class ReroutesThrottling {
```
sergii-iakovenko: Where is a unit test for it?

services/wfm/src/main/java/org/openkilda/wfm/topology/cache/CacheBolt.java
```diff
@@ -366,31 +365,14 @@ private void emitFlowMessage(InfoData data, Tuple tuple, String correlationId) t
...
-                json = Utils.MAPPER.writeValueAsString(new CommandMessage(
-                        request, System.currentTimeMillis(), correlationId, Destination.WFM));
-            } catch (JsonProcessingException exception) {
-                logger.error("Could not format flow reroute request by flow={}", flow, exception);
-                return;
-            }
-
-            Values values = new Values(json);
+            Values values = new Values(flowId, correlationId);
             outputCollector.emit(StreamType.WFM_DUMP.toString(), input, values);
```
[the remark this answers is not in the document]
Settis: Fixed.

templates/templates/wfm/topology.properties.j2
```diff
@@ -7,6 +7,9 @@ kafka.hosts = {{ kafka_hosts }}
 kafka.partitions.default = 1
 kafka.replication.default = 3
 
+reroute.throttling.timeout = 10
```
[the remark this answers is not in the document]
Settis: Done.

services/wfm/src/main/java/org/openkilda/wfm/topology/cache/CacheTopologyConfig.java
```diff
@@ -18,10 +18,17 @@
 import org.openkilda.wfm.topology.AbstractTopologyConfig;
 
 import com.sabre.oss.conf4j.annotation.Configuration;
+import com.sabre.oss.conf4j.annotation.Key;
 
 @Configuration
 public interface CacheTopologyConfig extends AbstractTopologyConfig {
 
+    @Key("reroute.throttling.timeout")
+    long getRerouteThrottlingTimeout();
```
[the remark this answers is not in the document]
Settis: Done.

services/wfm/src/main/java/org/openkilda/wfm/topology/cache/service/ReroutesThrottling.java
Settis: Done.

services/wfm/src/main/java/org/openkilda/wfm/topology/cache/service/ExtendableTimeWindow.java
```diff
@@ -0,0 +1,79 @@
...
+ *   limitations under the License.
+ */
+
+package org.openkilda.wfm.topology.cache.service;
+
+import java.time.Clock;
+import java.time.LocalDateTime;
+import java.time.temporal.ChronoUnit;
+
+public class ExtendableTimeWindow {
```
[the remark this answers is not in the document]
Settis: Done.

services/wfm/src/main/java/org/openkilda/wfm/topology/cache/service/ReroutesThrottling.java
Settis: Done.

services/wfm/src/main/java/org/openkilda/wfm/topology/AbstractTopologyConfig.java
```diff
@@ -35,6 +35,9 @@
     @Key("parallelism")
     int getParallelism();
 
+    @Key("parallelism.new")
+    int getNewParallelism();
```
sergii-iakovenko: AbstractTopologyConfig is the configuration base shared by multiple topologies, but "parallelism.new" means nothing to others except CacheTopology. Why can't it be in CacheTopologyConfig? 
BTW, What does "new" in parallelism.new means?

services/wfm/src/main/java/org/openkilda/wfm/topology/cache/CacheTopology.java
```diff
@@ -100,11 +102,17 @@ public StormTopology createTopology() throws NameCollisionException {
...
 
         /*
-         * Sends cache dump and reroute requests to `flow` topology.
+         * Sends cache dump and reroute requests to `flow` topology via throttling bolt.
          */
+        FlowThrottlingBolt flowThrottlingBolt = new FlowThrottlingBolt(
+                topologyConfig.getRerouteThrottlingMinDelay(), topologyConfig.getRerouteThrottlingMaxDelay());
+        int newParallelism = topologyConfig.getNewParallelism();
+        builder.setBolt(BOLT_ID_REROUTE_THROTTLING, flowThrottlingBolt, newParallelism)
+                .fieldsGrouping(BOLT_ID_CACHE, StreamType.WFM_REROUTE.toString(), new Fields(CacheBolt.FLOW_ID_FIELD));
```
sergii-iakovenko: Separate bolt instances don't share the state of ExtendableTimeWindow. And the grouping by flowId means that sliding of the window in one instance will not affect the window in another instance. But as per requirements we should wait for all ISLs to be settled down.

services/wfm/src/main/java/org/openkilda/wfm/topology/AbstractTopologyConfig.java
Settis: As I understand, we have all bolts with parallelism = 1. But all new bolts in Storm must have parallelism at least 2. I think that it's applicable for all topologies, that's why I put parameter in that general class. So, new bolts will have new parallelism. Maybe we must rename it somehow.

Issue #902, opened 2018-06-19
Update link props returns error if no props specified


rtretyak:
1. Send a link props Update request with no props.

*Expected:* Link props entity is successfully created with no props OR Bad Request error is returned
*Actual:* 500 internal error with null-pointer stacktrace in body is returned.

**Note:** Reproducible if missing any field (src/dst port/switch)

PR #1211, opened 2018-08-28
#902 Fix 500 error when creating link props.

dpoltavets:
This fix solves issue 902: if you do not specify the switch id/port when creating link props, we'll see a response with error 500.

### DIFF
--- a/services/src/messaging/src/main/java/org/openkilda/messaging/model/SwitchId.java
+++ b/services/src/messaging/src/main/java/org/openkilda/messaging/model/SwitchId.java
@@ -19,6 +19,7 @@ import org.openkilda.messaging.error.SwitchIdFormatException;
 
 import com.fasterxml.jackson.annotation.JsonValue;
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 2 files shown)


services/src/messaging/src/main/java/org/openkilda/messaging/model/SwitchId.java
```diff
@@ -31,29 +32,31 @@
...
+     * Switch id numerical representation.
      */
-    private final long switchId;
+    private final long switchIdNum;
 
     /**
      * Construct an instance based on the colon separated representation of a switch id.
      */
     public SwitchId(String switchId) {
+        Preconditions.checkArgument(Objects.nonNull(switchId), "Switch id must not be null");
```
sergii-iakovenko: Objects.nonNull over-complicates the code. "switchId != null" is completely cover the case. Objects.nonNull has been designed to be used as a functional predicate.

Objects.requireNonNull is what you needed: "This method is designed primarily for doing parameter validation in methods and constructors"

PR #1228, opened 2018-08-29
Update ATDD-Staging Failsafe suite to wait until a flow is rerouted;

### LINKED ISSUE #1227: Update ATDD-Staging tests to take rerouting sliding window into account

npliashechnikov:
Fix functional-tests on machines with non-US date/time format
Closes #1227

### DIFF
--- a/services/src/atdd-staging/src/main/resources/features/failsafe_behaviour.feature
+++ b/services/src/atdd-staging/src/main/resources/features/failsafe_behaviour.feature
@@ -10,9 +10,10 @@ Feature: Failsafe Suite
     Given Create 1 flow with A Switch used and at least 1 alternate path between source and destination switch and 500 bandwidth
 
     When ISL between switches loses connectivity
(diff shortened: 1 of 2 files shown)


services/src/atdd-staging/src/main/resources/features/failsafe_behaviour.feature
```diff
@@ -10,9 +10,10 @@ Feature: Failsafe Suite
     Given Create 1 flow with A Switch used and at least 1 alternate path between source and destination switch and 500 bandwidth
 
     When ISL between switches loses connectivity
-    And remains in this state for 30 seconds
+    And remains in this state for 240 seconds
```
rtretyak: I think it is better to update our staging config and set a lesser timeout. Why wait for 240 seconds if it is configurable

npliashechnikov:
Will ask Infra engineer to change discovery timeout to 30 seconds on staging, then set the correct time in the test steps.

Issue #796, opened 2018-05-17
Extend NB API with switch ports information from FL


nikitamarchenko:
Proxy call <FL>/wm/core/switch/all/port-desc/json across NB for UI

UI need two endpoints 
- from switch => /wm/core/switch/<DPID>/port-desc/json
- from sw+port => /wm/core/switch/<DPID>/port-desc/json + filter by port_number

Json format from FL

PR #1233, opened 2018-08-30
#796 Add NB API with switch ports information from FL.

dpoltavets:
Closes: #796


services/src/northbound/src/main/java/org/openkilda/northbound/controller/SwitchController.java
```diff
@@ -275,4 +277,36 @@ public PortDto configurePort(
...
+
+    /**
+     * Get port description of the switch.
+     *
+     * @param switchId the switch id.
+     * @param port the port of the switch.
+     * @return port description.
+     */
+    @ApiOperation(value = "Get port description from the switch", response = PortDescription.class)
+    @GetMapping(value = "/switches/{switch-id}/port_desc/{port}",
```
sergii-iakovenko: This URI doesn't follow RESTful API style: there's no such entity like "port_desc" to represent it as a resource. Reconsider to use just ".../ports/{port}"

docs/design/usecase/switch-get-port-description.md
```diff
@@ -0,0 +1,10 @@
+# Get port description of the switch
```
sergii-iakovenko: The sentence is not clear and should be to be rephrased.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/converter/OfPortDescConverter.java
```diff
@@ -0,0 +1,63 @@
...
+
+import org.projectfloodlight.openflow.protocol.OFPortConfig;
+import org.projectfloodlight.openflow.protocol.OFPortDesc;
+import org.projectfloodlight.openflow.protocol.OFPortFeatures;
+import org.projectfloodlight.openflow.protocol.OFPortState;
+
+public final class OfPortDescConverter {
+
+    /**
+     * Convert {@link OFPortDesc} to format that kilda supports.
```
sergii-iakovenko: "format that kilda supports." means a lot. Can we just mention a class or a payload for a specific response?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
```diff
@@ -756,6 +765,75 @@ private void doConfigurePort(final CommandMessage message, final String replyToT
...
+        try {
+            SwitchId switchId = request.getSwitchId();
+            logger.debug("Get port description for switch {}", switchId);
+            SwitchPortsDescription switchPortsDescription = getSwitchPortsDescription(switchId);
+
+            int port = request.getPortNumber();
+            PortDescription response =
+                    switchPortsDescription
+                            .getPortsDescription()
+                            .stream()
```
sergii-iakovenko: No reason to split "PortDescription response = switchPortsDescription.getPortsDescription()" on 3 lines. This just negatively affects readability.

Issue #1229, opened 2018-08-30
Deleting max_bandwidth link property removes it from neo4j schema


npliashechnikov:
Creating new issue after the partial fix for #1076 that fixed the invalid max_bandwidth data type.

Max_bandwidth link property is set by default in the neo4j. If user modifies max_bandwidth using "Update link properties" request and then removes link properties by "Delete link properties" request, max_bandwidth field gets removed from the ISL properties.
Instead, it should be reverted to the default value.

PR #1238, opened 2018-08-31
#1229 Recovery of max bandwidth when deleting link props.

dpoltavets:
When deleting the link props, the max bandwidth is restored from the database.

Closes: #1229


services/topology-engine/queue-engine/topologylistener/isl_utils.py
```diff
@@ -333,6 +333,9 @@ def del_props(tx, isl, props):
           (:switch {name: $dst_switch})""") + '\nREMOVE '.join(remove)
     db.log_query('ISL drop props', q, p)
     stats = tx.run(q, p).stats()
+    if 'max_bandwidth' in props:
+        db_isl = fetch(tx, isl)
+        set_props(tx, isl, {'max_bandwidth': db_isl['default_max_bandwidth']})
```
abochkarev: I would change` db_isl['default_max_bandwidth']` to `db_isl.get('default_max_bandwidth', 0)` to avoid KeyError exception.

PR #1239, opened 2018-08-31
[Floodlight] Fix checkstyle issues.

Settis:


abochkarev:
@Settis Looks like need to resolve sonar issues. Could you please do it?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/converter/OFlowStatsConverter.java
```diff
@@ -45,7 +46,7 @@
 /**
  * Utility class that converts OFlowStats from the switch to kilda known format for further processing.
  */
-public final class OFFlowStatsConverter {
+public final class OFlowStatsConverter {
```
sergii-iakovenko: Double "F" wasn't a mistake as came from OpenFlow + FlowStats (see org.projectfloodlight.openflow.protocol.OFFlowStatsEntry). I think the better way is to lower the first letter - OfFlowStatsConverter.

abochkarev: I will fix.

abochkarev: Fixed.

PR #1240, opened 2018-08-31
Added test for new self-loop ISL to a different port

rtretyak:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/IslMovingSpec.groovy
```diff
@@ -71,4 +75,34 @@ class IslMovingSpec extends BaseSpecification {
         then: "Original ISL becomes Discovered again"
         islUtils.waitForIslStatus([isl, islUtils.reverseIsl(isl)], DISCOVERED)
     }
+
+    def "New ISL is not getting discovered when adding new self-loop ISL (different port)"() {
+        given: "2 a-switch links on a single switch"
```
ylobankov: I would prefer to have 'Two' instead of 2.

Issue #1042, opened 2018-07-24
False positive missing OF rules detection


surabujin:
If switch will disconnect during switch `/switches/{switch_id}/rules/validate` call, all rules related to this switch will be reported as "missing".

```
July 24th 2018, 06:47:03.735	kilda-floodlight	ERROR	Could not get flow stats for ff:fe:00:00:00:00:00:01.
July 24th 2018, 06:47:03.734	kilda-floodlight	ERROR	Disconnecting switch [ff:fe:00:00:00:00:00:01(0x0) from 127.1.0.2/127.1.0.2:48634] due to IO Error: Connection reset by peer
July 24th 2018, 06:47:03.734	kilda-floodlight	INFO	[[ff:fe:00:00:00:00:00:01(0x0) from 127.1.0.2/127.1.0.2:48634]] Disconnected connection
July 24th 2018, 06:47:03.379	kilda-floodlight	DEBUG	Loading installed rules for switch ff:fe:00:00:00:00:00:01
```

The problem is in `org.openkilda.floodlight.switchmanager.SwitchManager#dumpFlowTable` it silently return emty list if `future.get()` raise an exception.

We need to distinguish "missing rules" and errors during report preparation. We should report that switch is unmanaged and the set of affected flows, and not report all OF rules as missing.

abochkarev:
This issue will require:
investigation - 2d;
collecting requirements - 3d;
implementation - 3d;
testing - 2d;

PR #1285, opened 2018-09-06
#1042 Output an error in the case when we can not validate the rules.

dpoltavets:
Added error output in the case when we can not get the switch rules when validating the rules.

Closes: #1042

### DIFF
--- a/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
+++ b/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
@@ -56,6 +56,7 @@ import org.openkilda.messaging.command.switches.SwitchRulesInstallRequest;
 import org.openkilda.messaging.error.ErrorData;
 import org.openkilda.messaging.error.ErrorMessage;
 import org.openkilda.messaging.error.ErrorType;
+import org.openkilda.messaging.error.rule.DumpRulesErrorData;
 import org.openkilda.messaging.floodlight.request.PingRequest;
 import org.openkilda.messaging.info.InfoMessage;
 import org.openkilda.messaging.info.discovery.DiscoPacketSendingConfirmation;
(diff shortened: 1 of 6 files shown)


services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
```diff
@@ -633,8 +634,8 @@ private void doDumpRulesRequest(final CommandMessage message,  String replyToTop
             context.getKafkaProducer().postMessage(replyToTopic, infoMessage);
         } catch (SwitchOperationException e) {
             logger.info("Dump rules is unsuccessful. Switch {} not found", request.getSwitchId());
-            ErrorData errorData = new ErrorData(ErrorType.DATA_INVALID, e.getMessage(),
-                    request.getSwitchId().toString());
+            ErrorData errorData = new DumpRulesErrorData(ErrorType.NOT_FOUND, e.getMessage(),
+                    "The switch was not found when requesting a rules dump.");
```
abochkarev: It would be great to add a switch id to the error message.

services/src/messaging/src/main/java/org/openkilda/messaging/error/rule/DumpRulesErrorData.java
```diff
@@ -0,0 +1,51 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+
+/**
+ * Defines the payload of a Message representing an error of dumping rules.
+ */
+@JsonSerialize
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder({
```
abochkarev: Would it be possible to remove JsonPropertyOrder from here?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
dpoltavets: This is the error description. The error message contains the switch id.

services/src/messaging/src/main/java/org/openkilda/messaging/error/rule/DumpRulesErrorData.java
dpoltavets: Yes, we can do it. I've fixed it.

Issue #1271, opened 2018-09-05
Remove maven dependency duplicates and improve dependency tree


niksv:
Some maven dependencies have duplicates. Also some of them have duplicate info (e.g. version in parent and in child dependency). It needs optimization

niksv:
estimates: 3 days

niksv:
PR: https://github.com/telstra/open-kilda/pull/1296

sergii-iakovenko:
PR #1296 has been reverted due to the issue in OpenTsdb Topology. See #1389

niksv:
At this moment Kilda uses Gradle instead maven

PR #1296, opened 2018-09-07
#1271 Maven dependencies optimisation

niksv:
- remove useless <execution> tags
 - move <version> tag to parent pom.xml

### DIFF
--- a/services/src/pom.xml
+++ b/services/src/pom.xml
@@ -69,7 +69,7 @@
         <validation-api.version>2.0.0.Final</validation-api.version>
         <hibernate-validator.version>6.0.10.Final</hibernate-validator.version>
         <glassfish-el.version>3.0.1-b09</glassfish-el.version>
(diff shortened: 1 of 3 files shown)


services/src/pom.xml
```diff
@@ -69,6 +69,7 @@
         <validation-api.version>2.0.0.Final</validation-api.version>
         <hibernate-validator.version>6.0.10.Final</hibernate-validator.version>
         <glassfish-el.version>3.0.1-b09</glassfish-el.version>
+        <com.auth0.version>3.2.0</com.auth0.version>
```
sergii-iakovenko: Why do you specify the version for the group instead of the artifact?

niksv: Because 'java-jwt' is the only one artifact which is used with this group. And also it was made just by analogy with other version properties. Most of them has group version.

niksv: I renamed it

PR #1297, opened 2018-09-07
Propose REST API design guidelines.

sergii-iakovenko:


docs/design/api-guidelines/api-guidelines.md
```diff
@@ -0,0 +1,233 @@
...
+Responses concerning successful operations should be coded according to the following rules:
+
+* `200 Ok`: Successful execution of `GET`, `PUT`, `PATCH` request. An existing resource is updated synchronously.
+* `201 Created`: Response to a `POST` request that results in entity creation.
+* `202 Accepted`: Accepted request concerning a `POST`, `PUT`, `DELETE`, or `PATCH` request that will be processed asynchronously.
+* `204 No Content`: Response to a successful request that will return empty body (as a `DELETE` request)
+
+Pay attention to the use of authentication and authorization error codes:
+
+* `401 Unauthorized`: Request failed because user is not authenticated
```
abochkarev: I would add the article 'a' before 'user'.

docs/design/api-guidelines/api-guidelines.md
```diff
@@ -0,0 +1,233 @@
...
+* `403 Forbidden`: Request failed because user does not have authorization to access a specific resource
```
abochkarev: ditto

docs/design/api-guidelines/api-guidelines.md
```diff
@@ -0,0 +1,233 @@
...
+
+#### Create entity
+
+* Server returns the `HTTP 201 Created` status code
+* Response will also include the `Location` header that points to the URL of the new resource ([RFC 6570](http://tools.ietf.org/html/rfc6570))
+* Server returns the new entity
+
+#### List entities (collection)
+
+* Response collection is presented as a plain array, not a dictionary with the array as a value.
```
surabujin: You should choose something one:
```
#### Wrap collection in object
 Always return root element as an object. This way you can add extra fields to the response without compatibility breakdown.
```

From my point of view - you must return `object` instead of `list`. At least it will be much easier to add pagination, when you will need to.

sergii-iakovenko: Nice catch! These two for sure contradict each other. I'll fix it.

docs/design/api-guidelines/api-guidelines.md
sergii-iakovenko: Fixed.

docs/design/api-guidelines/api-guidelines.md
sergii-iakovenko: Fixed.

docs/design/api-guidelines/api-guidelines.md
```diff
@@ -0,0 +1,233 @@
...
+Provide the full resource representation (i.e. the object with all properties) whenever possible in the response.
+
+### Filtering
+
+* Use a query parameter for each field that supports filtering.
+* If you want to filter many values in one field, you should repeat the field name many times in URL adding different values.
+
+### Command pattern
+
+Mapping CRUD operations to semantics of HTTP `POST`, `PUT`, `DELETE` is easy. However that is not the case for more complex operations that do more
```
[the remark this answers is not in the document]
sergii-iakovenko: Fixed.

docs/design/api-guidelines/api-guidelines.md
sergii-iakovenko: Fixed.

PR #1306, opened 2018-09-11
Fixed the issue with missed ISL Failed event when both switches go down

### LINKED ISSUE #1269: Flow won't evacuate from a broken ISL if it broke while its switches were down
The issue is that Kilda is unable to properly detect link as 'failed' and evacuate all the flows from the link if it stopped working while switches are down and it happened in less than 'discovery timeout' seconds.
Steps:
1. Create a flow
2. Pick two neighboring switches of this flow and disconnect them from controller.
3. Break the link between these two switches (bring the ISL's port down or break it via intermediate switch)
4. Bring the switches back online (connect back to controller). 
Note: Steps 2-4 should happen quicker than '${discovery.timeout}' param
**Expected:** The target ISL is FAILED. Flow is evacuated from this ISL or flow status changes to Down (if there is no alternative path to evacuate to)
**Actual:** The target ISL is FAILED, but the flow remains on that ISL. It won't attempt to reroute, leaving the flow on the broken ISL and in the Up status.
Note: The ISL will appear as 'failed' in the database, but won't be handled as 'failed' by the rest of the system.
@nikitacherevko did some investigation and can provide some more technical details on this problem.

nikitacherevko:
Closes #1269


services/src/messaging/src/main/java/org/openkilda/messaging/model/DiscoveryLink.java
```diff
@@ -116,7 +117,7 @@ public DiscoveryLink(@JsonProperty("source") final NetworkEndpoint source,
             @JsonProperty("consecutive_failure") final int consecutiveFailure,
             @JsonProperty("consecutive_success") final int consecutiveSuccess,
             @JsonProperty("consecutive_failure_limit") final int consecutiveFailureLimit,
-            @JsonProperty("active") final boolean active) {
+            @JsonProperty("link_state") final boolean active) {
```
surabujin: It is still boolean... why not change it on LinkState?

services/src/test-library/src/main/java/org/openkilda/testing/service/aswitch/ASwitchVirtualImpl.java
```diff
@@ -65,4 +69,17 @@ public void portsUp(List<Integer> ports) {
     public void portsDown(List<Integer> ports) {
         ports.forEach(port -> mininet.portDown(ASWITCH_NAME, port));
     }
+
+    @Override
+    public void knockoutSwitch(String switchId) {
```
surabujin: Can we use `SwitchId` instead of `String`?

PR #1320, opened 2018-09-12
Add code meters into FL

### LINKED ISSUE #996: Add FL request processing time metrics
To be able to detect FL overload we need time metrics for commands being processed by FL. I.e we need to log how long each request have been processed (level info).

surabujin:
Use library metrics to collect and report code timings. This timings are
required to locate performance bottlenecs.

To see collected metrict you need visualvm tool with MBeans plugin.

Start floodlight container, detirmine local port for internal port 11011
with command `docker-compose port floodlight 11011`. Create new jmx
connection into visualvm and investigate MBeans tab.

Close #996

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/1320)


surabujin:
It is useful function, but we can't deny or approve the impact on performance. It can/should be restored when we will need this functionality againe.

PR #1322, opened 2018-09-13
Handle invalid packet-in cookie

### LINKED ISSUE #1321: Invalid cookie value in packet-in OF messages
In some circumstances switches send invalid(0x0) value in cookie field in OF packet-in messages. At this moment we use cookie value to distinguish which OF rule was matches. As result valid packet-in messages with invalid cookie value are ignored by FL.

Affected subsystems - `isl discovery`, `flow pings`.

surabujin:
In some circumstances switches send invalid(0x0) value in cookie field
in OF packet-in messages. Alter cookieMismatch check to eliminate impact
such "invalid" cookies on packe-in processing.

Close #1321

### DIFF
--- a/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/model/OfInput.java
+++ b/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/model/OfInput.java
@@ -24,6 +24,7 @@ import org.projectfloodlight.openflow.protocol.OFPacketIn;
 import org.projectfloodlight.openflow.protocol.OFType;
 import org.projectfloodlight.openflow.types.DatapathId;
 import org.projectfloodlight.openflow.types.U64;
+import org.slf4j.Logger;
 
 public class OfInput {
     private final long receiveTime;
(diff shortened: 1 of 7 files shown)


services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/model/OfInput.java
```diff
@@ -63,13 +64,20 @@ public U64 packetInCookie() {
...
-            return null;
-        }
-
         return cookie;
     }
 
+    /**
+     * Check current packet(PACKET_IN) cookie value is mismatched with expected value.
+     */
+    public boolean packetInCookieMismatch(U64 expected, Logger log) {
```
sergii-iakovenko: 1. It's not obvious why the method accept the logger.
2. Should the getter method decide whether to log the result or not?

I think the method "unwrap" (which calls "packetInCookieMismatch") can log the details by itself.

surabujin: The only reason to log "match" results here - is complex(relatively) log message and several(2 at this moment) usage of this check. To keep ability to distinguish exact caller - passed logger instance is used to log messages.

PR #1323, opened 2018-09-13
Func tests/test for switch port config

ylobankov:


rtretyak:
Going forward, this spec will have to be supplied with additional Floodlight verification, and possible logs verification

Issue #650, opened 2018-04-21
Multiple floodlight instances per region


pzakatov:
## Problem
We have one floodlight instance. It is a single point of failure and a bottleneck

## Proposal
To have multiple floodlights per region

## Subtasks
* [x] Unify kafka-produces in FL #1340 
* [x] Implement FLRouter topology #1341 (Due to Dec. 23)
* [x] Create 2 floodlight environment for testing purpose(local)  (Due to Dec. 21)
* [x] Implement Router Topology with Switch Discovery on board.(Multi-switch)(Due to Dec. 28) 
* [x] Implement Alive mechanism with WFM as request origin (Due to Dec. 28)
* [x] Implement Install/Uniinstall Switch Rules Validate/Sync over Router (Single-switch) (Due to Dec. 25)
* [ ] Implement FL dump feature #1342 (1 week)

surabujin:
Design proposal https://github.com/telstra/open-kilda/pull/1277

surabujin:
One of subtasks(#1342) will be much simpler if task #1343 goes prior of #1342. #1343 implementation will decrease complexity and bug amount in all planned improvement of event topology.

I propose to make it first.

PR #1328, opened 2018-09-13
Remove unused batch install command

surabujin:


surabujin:
Related to #650

surabujin:
Due to "python" layer I have not found usage of this command. It is used in "switch sync" call. My fail, it should stay, at least for now.

PS It affect/block #650

PR #1329, opened 2018-09-14
Coping ui properties to target dir

PinGwynn:


services/src/openkilda-gui/Dockerfile
```diff
@@ -81,6 +81,7 @@ RUN mv select2.min.js src/main/resources/lib/javascript/
 RUN mv jquery.multi-select.js src/main/resources/lib/javascript/
 RUN ls -la 
 RUN mvn clean install
-
+RUN cp /app/src/main/resources/application.properties /app/target/ && \
```
surabujin: Because the `cp` target is the same for both commands, you can do it in one call to `cp`.

PR #1349, opened 2018-09-18
Move topology.yaml to src/test/resources dir in functional tests

ylobankov:


rtretyak:
This is not an example file, this is a default file which is used for automatic testing for every pull request

rtretyak:
Before merging make sure that appropriate Jenkins jobs are updated according to the new location of topology.yaml

ylobankov:
Closed in favor of https://github.com/telstra/open-kilda/pull/1450.

PR #1350, opened 2018-09-18
Updated kafka-python to 1.4

PinGwynn:


nikitacherevko:
We've tested this version of kafka-python on staging and it works as expected.

Issue #197, opened 2018-01-22
Investigate Topology Health Check design and implementation.


carmine:
Ensure the design and implementation is simple, elegant, well tested.

The basics should be covered:
(1) Healthy elements show up as healthy
(2) Unhealthy elements show up as unhealthy
(3) We follow policies regarding how we determine healthy / unhealthy
(4) We can easily report on the healthy / unhealthy / **flapping** elements
(5) We have policies for flapping elements - ie how to treat those.

Each layer we add to the initial / raw check should be like a pipeline of filters.

IlyaMoiseev:
Fixed in #1357

Issue #50, opened 2017-11-02
northbound: health-check fail if we are running several instances


PinGwynn:
Test case:
docker-compose scale northbound=2

result:
On the first instance:
curl -s http://172.18.0.7:8080/api/v1/health-check

{"name":"Northbound","version":"1.0-SNAPSHOT","description":"Northbound API service provides API for Kilda controller","components":{"event-splitter-storm-topology":"operational","flow-storm-topology"
:"operational","event-wfm-storm-topology":"operational","cache-storm-topology":"non-operational","statistics-storm-topology":"operational"}}

On the second one:
curl -s http://172.18.0.8:8080/api/v1/health-check

{"name":"Northbound","version":"1.0-SNAPSHOT","description":"Northbound API service provides API for Kilda controller","components":{"event-splitter-storm-topology":"non-operational","flow-storm-topology"
:"non-operational","event-wfm-storm-topology":"non-operational","cache-storm-topology":"non-operational","statistics-storm-topology":"non-operational"}}

If we stop instance where health-check is ok, non-operational instance comes to the operational state.

This behaviour blocks rolling-updates of northbound on Amazon ECS an example. Because before stops healthy task we must be sure that second task (container) in a healthy state too.

surabujin:
My estimate 1-3d

sergii-iakovenko:
We should consider [Spring Endpoints](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-monitoring.html) as implementation approach for health-checks. It provides extendable and configurable way to expose metrics and health-check information.

IlyaMoiseev:
Fixed in #1357

PR #1357, opened 2018-09-19
Healthcheck refactor

timofei-durakov:


Settis:
Nice idea. But as I see, it keeps only last sending status. I think it will be useful to keep last 10 statuses or statistics about last 10 minutes.

IlyaMoiseev:
Closes #1000, #197, #50

PR #1358, opened 2018-09-19
tests: Delete flow while it is in the reroute window

rtretyak:


services/src/test-library/src/main/java/org/openkilda/testing/service/database/DatabaseNeoImpl.java
```diff
@@ -132,6 +132,16 @@ public int getIslCost(Isl isl) {
...
     }
 
+    @Override
+    public int countFlows() {
+        String query = "MATCH ()-[f:flow]-() RETURN count(f)";
+        StatementResult result;
+        try (Session session = neo.session()) {
+            result = session.run(query);
+        }
+        return result.single().get("count(f)").asInt();
```
ylobankov: Will it work properly when we have no flows? I mean result.single().get("count(f)") may return NULL and we will fail when we try to do .asInt(). We already had a similar bug in the getIslCost method.

rtretyak: Yes it will. It properly returns 0 event for a wiped empty database.

PR #1361, opened 2018-09-19
Add test to verify ISL cost is NOT increased due to connection loss

ylobankov:
Closes: #1316.

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/isl/IslCostSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/isl/IslCostSpec.groovy
@@ -1,0 +1,52 @@
+package org.openkilda.functionaltests.spec.isl
+
+import static org.junit.Assume.assumeTrue
+
+import org.openkilda.functionaltests.BaseSpecification
+import org.openkilda.functionaltests.helpers.Wrappers
+import org.openkilda.messaging.info.event.IslChangeType
+import org.openkilda.testing.model.topology.TopologyDefinition
+import org.openkilda.testing.service.aswitch.ASwitchService
+import org.openkilda.testing.service.aswitch.model.ASwitchFlow
+import org.openkilda.testing.tools.IslUtils
+
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.beans.factory.annotation.Value
+
+class IslCostSpec extends BaseSpecification {
+    @Autowired
+    TopologyDefinition topology
+    @Autowired
+    IslUtils islUtils
+    @Autowired
+    ASwitchService aswitch
+
+    @Value('${discovery.interval}')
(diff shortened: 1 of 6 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/isl/IslCostSpec.groovy
```diff
@@ -0,0 +1,71 @@
...
+        def switches = topology.getActiveSwitches()
+        List<List<PathNode>> allPaths = []
+        List<PathNode> aswitchPath
+
+        [switches, switches].combinations().findAll { src, dst -> src != dst }.find { Switch src, Switch dst ->
+            allPaths = topologyEngineService.getPaths(src.dpId, dst.dpId)*.path
+            aswitchPath = allPaths.find { pathHelper.getInvolvedIsls(it).find { it.aswitch } }
+            aswitchPath
+        } ?: assumeTrue("Wasn't able to find suitable switch pair", false)
+
```
rtretyak: Not clear why are you performing a lot of strange actions above. Why do you need paths between switches at all if you are not creating any flows. If you need an aswitch isl to break just do `def aswitchIsl = topology.islsForActiveSwitches.find { it.aswitch }`

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/isl/IslCostSpec.groovy
```diff
@@ -0,0 +1,71 @@
...
+        when: "Remove a-switch rules to break link between switches"
+        int islCost = pathHelper.getCost(aswitchPath)
```
rtretyak: you are naming it 'islCost', but taking the cost of the whole path. should be `int islCost = pathHelper.getCost(aswitchIsl)`

rtretyak: by the way, it is probably makes sense to move getCost(Isl isl) method to islUtils intead of pathHelper.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/isl/SwitchFailuresSpec.groovy
```diff
@@ -1,4 +1,4 @@
-package org.openkilda.functionaltests.spec
+package org.openkilda.functionaltests.spec.isl
```
rtretyak: we already have 'spec.links' package. I find it pretty much the same meaning as 'isl'. Can we leave either isls or links?

ylobankov: Let's have a separate discussion about this offline.

PR #1364, opened 2018-09-20
Resolve circular dependencies in FL modules

### LINKED ISSUE #1363: Resolve circular dependencies in FL modules
We have several circular dependencies between `org.openkilda.floodlight.kafka.KafkaMessageCollector` and `org.openkilda.floodlight.pathverification.PathVerificationService`. It lead to incorrect/unstable module initialization process.

We must split utility services from all other services and break down all circular dependencies between modules.

Block #1340

surabujin:
We have several circular dependencies between
org.openkilda.floodlight.kafka.KafkaMessageCollector and
org.openkilda.floodlight.pathverification.PathVerificationService. It
lead to incorrect/unstable module initialization process.

Remove ConfigService, because there is no simple way to join
configuration from serval modules. Also it force to make all-to-all
dependencies.

Extract utility services from KafkaMessageCollector into new KildaCore
module.

Move PingService from KafkaMessageCollector into PathVerificationService
- there is dependencies between them and they have similar purpose.

Close #1363


services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/pathverification/PathVerificationService.java
```diff
@@ -126,16 +125,17 @@
...
+                IPathVerificationService.class,
+                PingService.class);
     }
 
     @Override
     public Map<Class<? extends IFloodlightService>, IFloodlightService> getServiceImpls() {
-        Map<Class<? extends IFloodlightService>, IFloodlightService> map = new HashMap<>();
-        map.put(IPathVerificationService.class, this);
-        return map;
+        Map<Class<? extends IFloodlightService>, IFloodlightService> services = new HashMap<>();
```
nikitacherevko: Let's use ImmutableMap as in other FL "modules"

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/KildaCore.java
```diff
@@ -0,0 +1,78 @@
...
+import net.floodlightcontroller.core.module.FloodlightModuleException;
+import net.floodlightcontroller.core.module.IFloodlightModule;
+import net.floodlightcontroller.core.module.IFloodlightService;
+import net.floodlightcontroller.threadpool.IThreadPoolService;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+public class KildaCore implements IFloodlightModule {
```
nikitacherevko: Would be great to describe (add some docs?) the purpose of this class and perhaps modify classname to something like KildaCoreModule, because it is not really obvious for the first look.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/KafkaMessageCollector.java
```diff
@@ -92,26 +69,25 @@ public KafkaMessageCollector() {
...
         logger.info("Starting {}", this.getClass().getCanonicalName());
 
-        commandProcessor.init(moduleContext);
-        pingService.init(moduleContext);
+        ConfigurationProvider provider = ConfigurationProvider.of(moduleContext, this);
+        KafkaConsumerConfig consumerConfig = provider.getConfiguration(KafkaConsumerConfig.class);
 
-        KafkaConsumerConfig consumerConfig = configService.getConsumerConfig();
-        KafkaTopicsConfig topicsConfig = configService.getTopics();
+        KafkaTopicsConfig topicsConfig = moduleContext.getServiceImpl(KafkaMessageProducer.class).getTopics();
```
nikitacherevko: Can we get rid of dependency between KafkaMessageCollector and KafkaMessageProducer? Please correct me if I'm wrong, but as I see we might get list of topics from KafkaConsumerConfig.

surabujin: This is second step in FL reorganisation. I am going to make one module that contain kafka producer service and service - factory for kafka consumers.

FL propose to split code across services (implement `IFloodlightService`), also it provide API to lookup any loaded service. So service is the only correct way to provide implement some functions. Services can handle requests produced by FL or by other services.

So if some service need to lookup topic name it should lookup and call some other service that provide this info. That why TopicsConfig are part of KafkaMessageProvider(some service).

PS Even if we ignore kafka-topics lookup, we can't drop KafkaMessageProducer(service) from here because KafkaMessageConsumer use it to send responses in kafka.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/command/ping/PingCommand.java
```diff
@@ -31,19 +31,21 @@
...
+
     PingCommand(CommandContext context) {
         super(context);
 
         FloodlightModuleContext moduleContext = getContext().getModuleContext();
-        configService = moduleContext.getServiceImpl(ConfigService.class);
         pingService = moduleContext.getServiceImpl(PingService.class);
         kafkaProducer = moduleContext.getServiceImpl(KafkaMessageProducer.class);
+
+        topics = moduleContext.getServiceImpl(KafkaMessageProducer.class).getTopics();
```
sergii-iakovenko: kafkaProducer is already obtained a line above, no sense to repeat.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/KafkaMessageCollector.java
```diff
@@ -50,37 +47,17 @@
 public class KafkaMessageCollector implements IFloodlightModule {
```
sergii-iakovenko: Shouldn't we transform it to IService instead?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/command/ping/PingCommand.java
surabujin: Nice catch, thanks. Fixed.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/KafkaMessageCollector.java
surabujin: There is no simple answer here. From my point of view FL service (IFloodlightService) is something that can be used "outside" of it i.e. it provide some "function". FL module (IFloodlightModule) is a container for services and "consumer" of "funtions" provided by FL services.

So until there is no any service/module that need to call something implemented inside `KafkaMessageCollector` it can stay IFloodlightModule.

But from other point of view we can convert it into IService(IFloodlightService) just to keep everithing uniformly.

Issue #1265, opened 2018-09-05
Add ability to kill/restart Floodlight programmatically via API.


rtretyak:
As a tester, I want to have an ability to control Floodlight lifecycle from test framework via some API, so that I can test Kilda behavior when controller dies or restarts.
Probably, this functionality can be added to Lock-keeper service (staging).
If possible, this should have implementation both for Staging env and for local env.

PR #1366, opened 2018-09-20
Lock-keeper: add ability to control FL's lifecycle

rtretyak:
part of #1265

### DIFF
--- a/services/lock-keeper/lockkeeper.py
+++ b/services/lock-keeper/lockkeeper.py
@@ -108,19 +108,19 @@ def ports_down():
     return jsonify({'status': 'ok'})
 
 
-@app.route('/floodlight/stop', methods=['GET'])
+@app.route('/floodlight/stop', methods=['POST'])
 def fl_stop():
     docker.from_env().containers.get(FL_CONTAINER_NAME).stop()
     return jsonify({'status': 'ok'})
 
 
-@app.route('/floodlight/start', methods=['GET'])
+@app.route('/floodlight/start', methods=['POST'])
 def fl_start():
     docker.from_env().containers.get(FL_CONTAINER_NAME).start()
     return jsonify({'status': 'ok'})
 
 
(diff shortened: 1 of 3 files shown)


services/lock-keeper/lockkeeper.py
```diff
@@ -104,3 +106,21 @@ def ports_up():
 def ports_down():
     change_ports_state(request.get_json(), "down")
     return jsonify({'status': 'ok'})
+
+
+@app.route('/floodlight/stop', methods=['GET'])
```
ylobankov: Maybe it should be POST instead of GET?

services/lock-keeper/lockkeeper.py
```diff
@@ -104,3 +106,21 @@ def ports_up():
...
+def fl_stop():
+    docker.from_env().containers.get(FL_CONTAINER_NAME).stop()
+    return jsonify({'status': 'ok'})
+
+
+@app.route('/floodlight/start', methods=['GET'])
```
ylobankov: Ditto.

services/lock-keeper/lockkeeper.py
```diff
@@ -104,3 +106,21 @@ def ports_up():
...
+def fl_start():
+    docker.from_env().containers.get(FL_CONTAINER_NAME).start()
+    return jsonify({'status': 'ok'})
+
+
+@app.route('/floodlight/restart', methods=['GET'])
```
ylobankov: Ditto.

services/lock-keeper/lockkeeper.py
rtretyak: What? why?

ylobankov: Because GET is intended to take some information about but not doing any actions. I am not saying that it should be exactly POST but not GET for sure.
methods=['GET'] -> methods=['POST']

PR #1367, opened 2018-09-20
Added proper lockKeeper implementations to control FL.

### LINKED ISSUE #1265: Add ability to kill/restart Floodlight programmatically via API.

rtretyak:
Also renamed ASwitchService to LockKeeperService.
Depends on #1366 
Closes #1265

### DIFF
--- a/services/src/test-library/src/main/java/org/openkilda/testing/service/lockkeeper/LockKeeperServiceImpl.java
+++ b/services/src/test-library/src/main/java/org/openkilda/testing/service/lockkeeper/LockKeeperServiceImpl.java
@@ -101,19 +101,19 @@ public class LockKeeperServiceImpl implements LockKeeperService {
     }
 
     @Override
-    public void stopController() {
+    public void stopFloodlight() {
         restTemplate.exchange("/floodlight/stop", HttpMethod.POST,
                 new HttpEntity(buildJsonHeaders()), String.class);
     }
 
     @Override
-    public void startController() {
+    public void startFloodlight() {
         restTemplate.exchange("/floodlight/start", HttpMethod.POST,
                 new HttpEntity(buildJsonHeaders()), String.class);
     }
 
     @Override
-    public void restartController() {
+    public void restartFloodlight() {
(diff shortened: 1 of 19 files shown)


services/src/test-library/src/main/java/org/openkilda/testing/service/lockkeeper/LockKeeperServiceImpl.java
```diff
@@ -98,6 +100,24 @@ public void reviveSwitch(String switchId, String controllerAddress) {
                 "reviveSwitch operation for a-switch is not available on hardware env");
     }
 
+    @Override
+    public void stopController() {
```
surabujin: To keep our terminology, this method should be `stopSpeaker` or `stopFloodlight`.

services/src/test-library/src/main/java/org/openkilda/testing/service/lockkeeper/LockKeeperVirtualImpl.java
```diff
@@ -30,15 +35,29 @@
...
 
+/**
+ * Simulates the same functionality as {@link LockKeeperServiceImpl} but for virtual Mininet network.
+ */
 @Service
 @Profile("virtual")
 @Slf4j
-public class ASwitchVirtualImpl implements ASwitchService {
+public class LockKeeperVirtualImpl implements LockKeeperService {
+    private static final String FL_CONTAINER_NAME = "/floodlight";
```
surabujin: Can it be config option?

rtretyak: Need to investigate how this is being configured during deployment. If it is not configurable there, then why bother. We do have couple more places with similar hardcoded values (e.g. speaker address when creating mininet topology). Will do this as a separate pr

PR #1369, opened 2018-09-20
tests: Added additional verifications that Kilda is up

rtretyak:
This fixes an issue when tests on Jenkins were run before Kilda was completely up and ready


ylobankov:
Maybe it is better to create this verification as a separate test that will run before all tests as you did for spring context already?

PR #1374, opened 2018-09-21
Wait for bandwidth to settle down after removing flows

rtretyak:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/ThrottlingRerouteSpec.groovy
```diff
@@ -262,6 +262,7 @@ class ThrottlingRerouteSpec extends BaseSpecification {
     def cleanup() {
         northboundService.deleteLinkProps(northboundService.getAllLinkProps())
         db.resetCosts()
+        assert Wrappers.wait(5) { northboundService.getAllLinks().every { it.availableBandwidth == it.speed } }
```
ylobankov: It looks like it will not work if we have some other flows in the system as it often happens on the staging.

rtretyak: 90% of our tests will not work in this case. Not having any flows is a precondition before executing the suite. We should be adding some mechanisms that will ensure that this is true

ylobankov: Ok, let's do it later.

PR #1375, opened 2018-09-23
Increase timeout to make CI job more stable

ylobankov:


rtretyak:
Added this to #1377 
+5 was not enough for staging, I caught a fail couple times so I did `* 1.5`

Issue #580, opened 2018-04-14
Timestamp of packets on Novi (Latency measurement)


pzakatov:
# UseCase
As a user, I want to see ISL roundtrip latency measured from a single switch and stored in DB and OTSDB

# Logic
1 Send a discovery as a packet_out to switch A
2 Switch A put a timestamp `t1` to the discovery packet and send it to switch B
3 Switch B receives the discovery packet and does 2 things: send it to the controller, modify packet (UDP port) and send it back to switch A (via group tables)
4 Switch A receives his own message, add a timestamp `t2` to it and send it to a controller.
5 Controller use the discovery packet from Switch B for ISL discovery purposes
6 Controller use the discovery packet from Switch A for a roundtrip latency measurement purposes (`t2-t1`)

# Details
Accurately measuring latency within an OpenFlow network has caused challenges as the OpenFlow protocol itself does not have a standardised way to measure latency.  A common technique is to send a packet which has the timestamp embedded in the packet of when the controller sent the packet.  When the receiving switch sends the packet back to the controller the receive time by the controller is known and thus the latency can be calculated.

That mechanism for latency calculation has a challenge in that the latency between the Controller and the switch is not easy to account for.  A common practice is to add the average latency between Controller and Switch to the T0 (time PACKET_OUT was sent by the Controller) and subtract the latency from T1 (time PACKET_OUT was received by a controller).  In a large network with high latencies and jitter, those numbers can become meaningless quickly.  Couple that with the processing time of the PACKET_OUT/PACKET_IN messages within the switch and latency is not very accurate.

To combat that issue Noviflow has added an extension allowing the time a packet is received and transmitted by a switch to be appended to the packet.  That timestamp is based on a software mechanism and accuracy is likely to be +/- 1-2ms.  

Noviflow has also added an extension which allows the switch to append a hardware-based timestamp that is extremely accurate to all outgoing packets (matching a specific match rule).  However, as the timestamp is appended to the packet after the ethernet frame the packet is not a valid Ethernet packet and some equipment will drop the packet as being malformed.

[...]

PR #1376, opened 2018-09-24
Switch timestamps

jonvestal:
This adds the ability to have Noviflow switches put a timestamp into two new TLV's in the Discovery Payload.  The timestamps are inserted directly by the switch on sending the packet from switch_a and receiving the packet on switch_b.  Also added is a new property to enable/disable the feature with default to disable.

Latency calculation changed such that if the new timestamps are in the TLV they are used, else fallback to a legacy method of controller-based timestamps.

Does change timestamp to nanoseconds.

Related to #580 

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/1376)


rtretyak:
Logs are flooded with FL errors of type 
```
OFBadActionErrorMsgVer13(xid=1760, code=BAD_EXPERIMENTER, data=[unparsed: 04 0d 01 4e 00 00 06 e0 ff ff ff ff ff ff ff fd 00 38 00 00 00 00 00 00 ff ff 00 28 ff 00 00 02 ff 00 00 05 00 40 00 00 03 40 00 00 ff ff 0e 04 ff 00 00 02 ff ff 10 04 ff 00 00 02 00 00 00 00]) from switch OFSwitch DPID[00:00:00:00:00:00:00:03] in state net.floodlightcontroller.core.internal.OFSwitchHandshakeHandler$MasterState@77b8a2e0
```
You can get the logstash dumps from the artifacts of the jenkins build

jonvestal:
It works fine on my environment.  There are changes to Loxigen that are required and they have been committed a while back.  I’m wondering if you are picking those changes up?

jonvestal:
And I’ve tested it against live switches in my lab and they are discovered.  I think the build is not picking up the latest Loxigen.

jonvestal:
Sorry keep asking questions.   I’m wondering if the functional tests are failing because we are now using Noviflow specific features?  

Sorry I’m not sure how to get into Jenkins when I’m not in the office.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/pathverification/PathVerificationService.java
```diff
@@ -212,10 +215,23 @@ public Command call() throws Exception {
     protected List<OFAction> getDiscoveryActions(IOFSwitch sw, OFPort port) {
         // set actions
         List<OFAction> actions = new ArrayList<>();
+        actions.add(actionAddTxTimestamp(sw, 832));
```
surabujin: Is this offset in bits? Can we calculate it using our `discovery packet generator`? There is a risk that packet format will be altered and this "placeholder" will be moved.

From my point of view the best option is to put this 2 fields at the begin of payload, and shift LLDP on 16 bytes(2 timestamp fields) from the start of payload. I.e. 2 timestamp fields will precede LLDP and will be outside LLDP staff.

In this case we will guarantee constant timestamp placeholder's offsets. Also it is extremely simple to extract this fields and check their "existence". The only obstacle I see is the goal to keep "valid" LLDP packet (wrapped in UDP). Do we really need to keep valid LLD packet?

jonvestal:
I like the LLDP packet format which can be used by other systems in the future.  Why not stay with LLDP instead of creating our own packet format?  I know it is wrapped in UDP but if we strip the UDP then we can have any LLDP receiver receiving these packets and can take them off of the control plane.

Jon

jonvestal:
As previously mentioned there is value to keeping LLDP as it can be read by any LLDP daemon, once the UDP is stripped.  That being said, Carmine and I have been talking about changing the way discovery works to basically remove any business logic from Floodlight and to move most processing to the switch.  

While playing around with timestamps, if we can’t keep the switches in extremely tight clock lockstep then the switch timestamps are not that accurate.  NTP is not sufficient to keep the clocks in sync with less than 1 second accuracy which is not going to help with latency.  PTP (Precision Time Protocol) works but has some deployment and distance limitations and may not be deployable for us unless we have local GPS sync’d Master Clocks.

With that in mind we need to consider changing the way we are measuring latency to be round-trip instead of uni-directional and rely on the timestamps based on the initiating switch and thus we do not have a requirement for sub-millisecond precision across our switch clocks.  Utilising the copy_field feature of OF1.5 (and back ported to OF1.3 in Noviflow) we can have the switches add what they need to directly to the payload as it traverses the network.  For ISL latency measurement we can also off-load the response such that it does not end up at the controller but instead goes to a collection server that is responsible for sending them onto floodlight thus removing even more load from Floodlight.

A simple diagram looks like this:



Each switch then has 2 match rules:

1.  LATENCY_REQUEST_MATCH:
		eth_dst = 08:ED:02:E3:FF:FF
		eth_type = 0x803
		udp_src_port = 61231 
		udp_dst_port = 61231

2.  LATENCY_REPONSE_MATCH:
		eth_dst = 08:ED:02:E3:FF:FF
		eth_type = 0x803
		udp_src_port = 61232 
		udp_dst_port = 61232


This introduces a new concept which is an Latency Collector service that can grab these packets and send them on to Kafka, don’t process them there, instead process as normal in Storm which gives us horizontal scale.  Still using UDP as depending on the barrier network (whatever is between the two switches) ICMP could be rate-limited (thinking VPLS types of cores).

In this model if we continue to use LLDP as the payload then we can having a running start on a collector.  If we use a custom format, then we will need to write our own collector (not horrible but more work).

Comments?

Jon

carmine:
Is this PR still valid?

carmine:
@jonvestal  - looks like this needs to be re-based and/or get the conflicts eliminated (re-merge).

@surabujin - have you given this a LGTM?

surabujin:
@carmine From my point of view, we shoult keep this POC as separate branch (not merged). And use it during design/implementation of telemetry server (mentioned above service to collect `discovery` packets instead of FL).

pzakatov:
Code from this PR will be used as a part of #580

PR #1379, opened 2018-09-24
[TrivialFix] Moved floodlight exceptions to 'error' package

niksv:
- Exception classes were moved to 'exception' package
 - Fixed some checkstyle issues


niksv:
@surabujin Initially this patch was made just because half of all classes in switchmanager package are exceptions. I didn't saw error package on top level but now I think this is a good place to move all exceptions from floodlight module.

surabujin:
Resolve conflicts (rebase).

Issue #984, opened 2018-07-09
Deleting meters with -1 index leads to all meters being removed


rtretyak:
Steps:
1. Create flow (or two flows between same switches)
2. Get all meters for src/dst switch. Observe non-empty list
3. Run delete meters command with -1 index `/api/v1/switches/{switch_id}/meter/-1`
**Expected:** No meters removed.
**Actual:** All meters for target switch are removed

PR #1385, opened 2018-09-25
Fixed issue with deleting meter with -1 id

niksv:
- Now removing of meter with -1 id causes 400 BadRequest

Closes: #984


services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/NorthboundSteps.java
```diff
@@ -107,4 +107,10 @@ public void removeMeter(String meterAlias, String switchAlias) {
     public void removeMeterResponseIsSuccessful() {
         assertTrue(deleteMeterResponse.isDeleted());
     }
+
+    @And("^remove meter -1 from '(.*)'$")
```
nikitamarchenko: For what purpose you added that?

niksv: Oops.. It was old test implementation. I forgot to remove it

PR #1419, opened 2018-09-28
Synchronize ISL discovery messages in FL

### LINKED ISSUE #786: Event/wfm possible events lost during sync with FL
During sync with FL event/wfm wait `sync-end-marker` in input stream. And all events that come before process as `sync events`. But on FL side (due to multithreaded nature) "real" event can be produced and posted in the middle of sync `stream`.

So it will be processed as `sync event` i.e. it not processed (only event/wfm cache update). It can lead(depends from event) to "lost" ISLs/ports.

surabujin:
Move all code that produce ISL discovery messages(including network
dump) into SwitchEventCollector(and rename it into
SwitchTrackingService). Add synchronisation(lock) around all ISL
discovery message producing code.

As result FL will wait until whole "network dump" is sent before sending
any other ISL discovery messages.

Close #786


nikitacherevko:
Please resolve conflicts

Issue #1027, opened 2018-07-20
atdd-staging: add switch manufacturer/model to the topology model


npliashechnikov:
In order to test flow management reliably, we need to run tests on all models of the networking equipment present in the testing environment. Previously, reliability was achieved by selecting switches in a random fashion, but at the cost of the reproduceability. To achieve both, test should be aware of the equipment in the environment.

rtretyak:
For now decided to go with the switch description, since it usually has the manufacturer and even the firmware version

PR #1420, opened 2018-09-28
Added test permutations for all unique switch descriptions

rtretyak:
Related to #1027. Need to discus whether this approach closes that issue

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowPingSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowPingSpec.groovy
@@ -11,6 +11,7 @@ import org.openkilda.functionaltests.helpers.PathHelper
 import org.openkilda.functionaltests.helpers.Wrappers
 import org.openkilda.messaging.info.event.IslChangeType
 import org.openkilda.messaging.info.event.PathNode
+import org.openkilda.messaging.info.event.SwitchInfoData
 import org.openkilda.messaging.payload.flow.FlowState
 import org.openkilda.northbound.dto.flows.PingInput
 import org.openkilda.northbound.dto.flows.PingOutput.PingOutputBuilder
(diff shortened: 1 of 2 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowPingSpec.groovy
```diff
@@ -224,7 +224,30 @@ class FlowPingSpec extends BaseSpecification {
     }
 
     def getOfSwitchCombinations() {
-        return [topology.activeSwitches, topology.activeSwitches].combinations().findAll { src, dst -> src != dst }
-                .unique { it.ofVersion.sort() }
+        def nbSwitches = northboundService.getActiveSwitches()
```
npliashechnikov: Wouldn't it be better to use nonCentecSwitches() instead of getActiveSwitches()? Looks like Centecs are filtered out anyway.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowPingSpec.groovy
```diff
@@ -224,7 +224,30 @@ class FlowPingSpec extends BaseSpecification {
     }
 
     def getOfSwitchCombinations() {
```
npliashechnikov: If possible, please change method name to getOFSwitchCombinations(). Also, would be great to move code comments on the switches quirks to the groovydoc.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowPingSpec.groovy
rtretyak: The problem is that nonCentecSwitches returns list of Switch objects (topology.yaml format), while I need SwitchInfoData here (northbound format).

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowPingSpec.groovy
rtretyak: `getOFSwitchCombinations` is incorrect camel-case name. Won't do.
Will add a brief groovydoc describing what this method does.

rtretyak: please see this guide (we are following Google style guide on Kilda)
http://google.github.io/styleguide/javaguide.html#s5.3-camel-case
p.3

PR #1430, opened 2018-10-04
Use correct logger object in abstract bolts

surabujin:
Force abstract bolt successors to define logger object. It allow to
identify correct bolt class(using loggerName) for messages produced by
parent/abstract level.


services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/RouterBolt.java
```diff
@@ -83,4 +85,9 @@ public void declareOutputFields(OutputFieldsDeclarer declarer) {
         declarer.declareStream(StreamType.ISL.toString(), new Fields("request", "correlationId"));
         declarer.declareStream(StreamType.FLOW.toString(), new Fields("request", "correlationId"));
     }
+
+    @Override
+    protected Logger getLog() {
```
nikitacherevko: there are already defined loggers in nbworker topology, could you replace old ones?

surabujin: my fail, fixing it.

PR #1435, opened 2018-10-08
LabService implementation

### LINKED ISSUE #1133: Implement the LabService to enable testing on both virtual and physical environments
In order to create the automated test cases which will run seamlessly on either physical or virtual OpenKilda deployments using the same code paths, we are going to need a common interface for accessing them.

So far, there is a design document describing the LabService API here: https://github.com/telstra/open-kilda/tree/develop/docs/design/test-lab. The implementation of such service would be extremely useful for the test cases development and debugging, as the QA engineers wouldn't have to use the physical environment in the process of debugging.

IlyaMoiseev:
Proposal: https://github.com/telstra/open-kilda/blob/develop/docs/design/test-lab/test-lab.md
Closes #1133


services/src/test-library/src/main/java/org/openkilda/testing/config/DefaultServiceConfig.java
```diff
@@ -66,8 +67,10 @@ public RestTemplate traffExamRestTemplate() {
     }
 
     @Bean(name = "lockKeeperRestTemplate")
-    public RestTemplate lockKeeperRestTemplate(@Value("${lockkeeper.endpoint}") String endpoint) {
-        return buildLoggingRestTemplate(endpoint);
+    public RestTemplate lockKeeperRestTemplate(
+            @Value("${lab-api.endpoint}") String baseEndpoint,
+            LabService labService) {
+        return buildLoggingRestTemplate(baseEndpoint + "/api/" + labService.getLab().getLabId() + "/lock-keeper");
```
rtretyak: I don't understand how this will work for hardware environment at this point. If I execute a hardware profile it will create a virtual lab for me, won't it?

services/src/test-library/src/main/java/org/openkilda/testing/service/lockkeeper/LockKeeperService.java
```diff
@@ -15,7 +15,7 @@
 
 package org.openkilda.testing.service.lockkeeper;
 
-import org.openkilda.messaging.model.SwitchId;
+import org.openkilda.testing.model.topology.TopologyDefinition.Switch;
```
rtretyak: Don't quite understand this change. Why pass the whole Switch object?

services/src/test-library/src/main/java/org/openkilda/testing/service/lockkeeper/LockKeeperVirtualImpl.java
```diff
@@ -15,40 +15,48 @@
...
- * Simulates the same functionality as {@link LockKeeperServiceImpl} but for virtual Mininet network.
+ * Simulates the same functionality as {@link LockKeeperServiceImpl} but for virtual network.
  */
 @Service
 @Profile("virtual")
 @Slf4j
 public class LockKeeperVirtualImpl implements LockKeeperService {
     private static final String FL_CONTAINER_NAME = "/floodlight";
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(LockKeeperVirtualImpl.class);
```
rtretyak: Should be removed since we use lombok's `@Slf4j`

services/src/test-library/src/main/java/org/openkilda/testing/service/lockkeeper/LockKeeperVirtualImpl.java
```diff
@@ -61,46 +69,56 @@ public LockKeeperVirtualImpl() throws DockerCertificateException, DockerExceptio
...
     }
 
     @Override
-    public void reviveSwitch(SwitchId switchId) {
-        String switchName = topology.getSwitches().stream()
-                .filter(sw -> sw.getDpId().equals(switchId)).findFirst().get().getName();
-        mininet.revive(switchName, VIRTUAL_CONTROLLER_ADDRESS);
+    public void reviveSwitch(Switch switchDef) {
+        restTemplate.exchange("/reviveswitch", HttpMethod.POST,
+                new HttpEntity<>(new SwitchModify(switchDef.getName(), VIRTUAL_CONTROLLER_ADDRESS),
```
rtretyak: since lab-service will run on a remote server I suppose it won't be able to resolve the "tcp:kilda:6653" host. So we should instead pass executor's IP address (which will have a Floodlight and the whole Kilda running locally) from tests, which can then be used at the lab-service's side

services/src/atdd-staging/topology.yaml.example
```diff
@@ -61,6 +61,6 @@ trafgens:
       switch_port: 11
       status: active
 
-trafgen_config:
+traffgen_config:
```
rtretyak: As far as I understand, this should now also have a 'controller' section

services/src/test-library/src/main/java/org/openkilda/testing/model/topology/TopologyDefinition.java
```diff
@@ -57,6 +61,7 @@
     private List<TraffGen> traffGens;
     @NonNull
     private TraffGenConfig traffGenConfig;
+    private final String controller = VIRTUAL_CONTROLLER_ADDRESS;
```
rtretyak: This value should probably be handled via config file, because "tcp:kilda:6653" won't resolve on the LabService's side

rtretyak:
Is this version of LabService should be able to handle the hardware(staging) env?

services/lab-api/lab-api/api.py
```diff
@@ -0,0 +1,105 @@
...
+        cnt.remove()
+    except KeyError as ke:
+        logger.error(ke)
+        return 'No lab with id {}'.format(lab_id)
+    except Exception as ex:
+        logger.warn(ex)
+
+    return 'OK'
+
+
```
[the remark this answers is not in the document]
IlyaMoiseev: Added endpoint to return all active labs

services/src/test-library/src/main/java/org/openkilda/testing/service/lockkeeper/LockKeeperService.java
IlyaMoiseev: We need only switch name to manipulate its state. But I think it's more clear when we pass whole switch object

services/lab-api/lab-api/api.py
```diff
@@ -0,0 +1,162 @@
...
+        is_hw = request.headers.get('Hw-Env', False)
+        if is_hw:
+            if str(HW_LAB_ID) in labs:
+                return Response('Hardware lab already defined', status=409)
+            else:
+                labs[str(HW_LAB_ID)] = Lab(HW_LAB_ID, request.get_json())
+                return jsonify({'lab_id': HW_LAB_ID})
+
+        count = count + 1
+        labs[str(count)] = Lab(count, request.get_json())
```
surabujin: ...
use list for labs instead of dict
```
dummy = object()
labs.append(dummy)
index = labs.index(dummy)

labs[index] = Lab(index, reqiest.get_json())
```

On lab removal do `labs[index] = None`... no global variables and thread safe. If avoid passing id/index into Lab constructor - this schema will become much simpler(no need in dummy object).

services/lab-api/lab-api/api.py
```diff
@@ -0,0 +1,162 @@
...
+        if lab_id in labs:
+            labs[lab_id].destroy()
+            del labs[lab_id]
+        else:
+            return Response('No lab with id %s' % lab_id, status=404)
+    except Exception as ex:
+        logger.error(ex)
+        return Response(str(ex), status=500)
+
+    return 'OK'
```
surabujin: not a json... break down REST model

services/lab-api/lab-api/api.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+@app.route('/api/<lab_id>/definition', methods=['GET'])
+def get_lab_definition(lab_id):
+    return jsonify(labs[lab_id].lab_def)
```
surabujin: where error handling for "invalid" id?

services/lab-api/lab-api/api.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def proxy_request(host, port, path):
+    url = 'http://{}:{}/{}'.format(host, str(port), path)
```
surabujin: Use at least `urllib.parse.urlunparse` for proper encoding.

services/lab-api/lab-api/api.py
```diff
@@ -0,0 +1,162 @@
...
+    r = requests.request(
+        request.method, url, headers=request.headers, data=request.data
+    )
+    return Response(r.content, status=r.status_code)
+
+
+@app.route('/api/<lab_id>/lock-keeper/<path:to_path>', methods=['GET', 'POST', 'DELETE'])
+def lockkeeper_proxy(lab_id, to_path):
+    if int(lab_id) == HW_LAB_ID:
+        return proxy_request(HW_LOCKKEEPER_REST_HOST, '5001', to_path)
```
surabujin: Why not fetch address from container's network?

services/lab-service/lab-service/cmd.py
```diff
@@ -0,0 +1,56 @@
...
+    :param cmd: the command to run
+    :param sync: block and wait for command termination
+    :param return_code: return just cmd exit code. Work only for sync=True
+    :return:
+    Stdout+stderr, when sync=True,return_code=False.
+    Cmd exit code, when sync=True,return_code=True.
+    Subprocess.Popen instance, when sync=False.
+    """
+    logger.warn(cmd)
+    p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
```
surabujin: use list for cmd. And why the hell `shell=True` is doing here?

services/lab-service/lab-service/cmd.py
```diff
@@ -0,0 +1,56 @@
...
+    if sync:
+        output, error = p.communicate()
+        if return_code:
```
surabujin: extremely ugly design

services/lab-service/lab-service/main.py
```diff
@@ -0,0 +1,51 @@
...
+from topology import Topology
+from lockkeeper import run_server
+
+logger = logging.getLogger()
+
+LAB_ID = os.environ.get("LAB_ID", 1)
+API_HOST = os.environ.get("API_HOST", 'lab-api.pendev:8288')
+
+
+if __name__ == '__main__':
```
surabujin: Can we produce normal entry points using setuptools?

nevermind, just crying...

services/lab-service/lab-service/main.py
```diff
@@ -0,0 +1,51 @@
...
+    url = "http://{}/api/{}/definition".format(API_HOST, LAB_ID)
+    topo_def = requests.get(url).json()
```
surabujin: Getting it via stdin should/can be much simpler.

services/lab-service/lab-service/main.py
```diff
@@ -0,0 +1,51 @@
...
+    topo = Topology.create(topo_def)
+
+    logger.warn("Running topology")
```
surabujin: you do not setup logger

services/lab-service/lab-service/topology.py
```diff
@@ -0,0 +1,247 @@
...
+        self.add_route_flows(a_mappings)
+
+    def add_route_flows(self, mappings):
+        if mappings:
+            ofctl(['add-flow {sw} -O {of_ver} in_port={in_port},actions=output={out_port}'
+                       .format(sw=self.name, of_ver=self.of_ver, **flow) for flow in mappings])
+
+    def del_route_flows(self, mappings):
+        if mappings:
+            ofctl(['del-flows {sw} -O {of_ver} in_port={in_port}'
```
surabujin: We can use cookie(masked or not) to remove all "required" rules/flows.

services/lab-service/lab-service/cmd.py
IlyaMoiseev: Because it runs shell command, and without shell environment they don't working

services/lab-service/lab-service/main.py
IlyaMoiseev: Nope. In case of stdin it should  take care about json escaping when pass through docker arguments\env, and json string may be quite big

services/lab-service/lab-service/topology.py
IlyaMoiseev: It is ASwith class,  what handle only simple port to port routes. I doesn't see any profit of using addition cookie abstraction on these actions

rtretyak:
@IlyaMoiseev plz let me fully check this locally before merge. Will return the 'Ready To Merge' label when I'm done

rtretyak:
Currently this breaks 'hardware' profile and makes it impossible to run tests against our staging environment. What 'lab-api.endpoint' should I use for staging env? Should we deploy this on staging? Using local lab service disables me from accessing traffic generators in stage env, thus preventing the whole suite from starting

rtretyak:
In case of not enough ports on the switches in topology.yaml (max_port) to build all the required links, the lab-service won't throw/log any errors, the topology will just not appear. Took me quite some time to debug the issue, would be good if lab-service somehow informed me that it failed to build the topo

IlyaMoiseev:
Current issues:
Virtual topology creates asyncronicaly, so, if something go wrong the test suit will know nothing about that
Topology delete after tests ends - needed avoid that for debug proposes

docker-compose.yml
```diff
@@ -20,6 +20,21 @@ services:
...
+    - "8288:8288"
+    privileged: true
+    command: api
+    volumes:
+    - /var/run/docker.sock:/var/run/docker.sock
+    networks:
+      default:
+        aliases:
+        - lab-api.pendev
+
```
nikitamarchenko: add mem_limit

services/lab-service/lab/api/api.py
```diff
@@ -0,0 +1,209 @@
...
+HW_LAB_ID = 0
+count = itertools.count(start=1, step=1)
+labs = {}
+
+
+def make_container_name(id):
+    return 'lab_service-' + str(id)
+
+
+class Lab:
```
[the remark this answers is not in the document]
nikitamarchenko: no, python2 is dead

services/lab-service/lab/Makefile
```diff
@@ -0,0 +1,14 @@
+all: build
+
+build:
+	tox -e pep8,py35
+	.tox/py35/bin/python setup.py bdist_wheel
```
nikitamarchenko: remove wheel and use raw setup.py install, also your command install app into virtual env don't do that.

services/lab-service/lab/api/api.py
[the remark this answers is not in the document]
timofei-durakov: up 2 u, used to make things explicit here

services/lab-service/lab-service/main.py
surabujin: Why it can't be started via sh wrapper?

IlyaMoiseev: This code invoked in container by request from containerized Api, that itself gets JSON definition through http from test framework. So, chain is: Test framework -> Api (docker) -> Docker daemon -> **Topology** (docker). 
From my point of view it's easier to ask Api for topology definition by http than passing it through docker daemon some other way.

Issue #1115, opened 2018-08-07
Handle for change HTTP error codes (4xx, 5xx)


Levitan:
Change HTTP code for error responses described in Swagger UI.
400 - Invalid data
408 - Request timed out

PR #1436, opened 2018-10-09
Change 400 and 408 HttpStatus for same ErrorTypes

Levitan:
Issue #1115


rtretyak:
Which requests are affected by this change?

Levitan:
> Which requests are affected by this change?

400 - added for REQUEST_INVALID ErrorType, used for internal exceptions
408 - all requests which not receive message from kafka

Issue #1437, opened 2018-10-10
GUI container do not start in ALL in one build


surabujin:
When you try do build and run kilda using our guidelines in README files... you will not get working GUI. Because GUI does not start inside GUI container.

Steps to reproduce:
* clone/pull fresh code from develop branch
* `make build-latest`
* `make up-test-mode`

Try to open url "http://localhost:8010/openkilda" - get "connection refused response".

PS Why for we need separate docker-compose.yml into `services/src/openkilda-gui`?

PR #1439, opened 2018-10-10
Changed workdir for GUI

PinGwynn:
Changed workdir for Openkilda GUI (resolved issue with logback.xml file location, which should be in current dir)


gauravchug:
#1437 Issue fixed with this change and now GUI container will start in all in one build.

PR #1444, opened 2018-10-12
New API: delete switch

### LINKED ISSUE #328: Create a way to delete switch when it is no longer used
Check that switch is not a part of a flow
Check that switch have no ISLs
Check that switch is not Active
Delete switch from neo4j
Delete switch from a cache

Proposed design
---------------------
Add `DELETE` action to `/switches/{switch_id}` in NB API with optional flag `force` (default is false).
The action must be protected by `EXTRA_AUTH`.

Feature workflow when the event dispatched from the separate topology:
sendfromtopo

A detailed description of the checks:
- Check that the switch is deactivated. Read switch status from neo4j and check it. Otherwise, raise an error "Removable switch is active. It must be deactivated first.".
- Check that the switch is not a part of a flow. Do `/switches/{switch_id}/rules/validate` and check that no rules are presented. Otherwise, raise an error "Switch have assigned rules. It must be freed first.".
- Check that the switch has no active ISL's. Otherwise, raise an error "Switch have active links. Ensure that it the switch is unplugged first.".
- Check that the switch has no inactive ISL's. Otherwise, raise an error "Switch have inactive links. Remove them first.".

Steps to complete
-----------------------
- [ ] approve the desing
- [ ] implement NB API
- [ ] implement checks (can be a part of NB)
- [ ] implement deleting the switch from neo4j
- [ ] implement deleting the switch from cache #1074

niksv:
Closes #328

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/1444)

### DIFF
--- a/services/src/api/src/main/java/org/openkilda/northbound/dto/switches/DeleteSwitchResult.java
+++ b/services/src/api/src/main/java/org/openkilda/northbound/dto/switches/DeleteSwitchResult.java
@@ -1,4 +1,4 @@
-/* Copyright 2018 Telstra Open Source
+/* Copyright 2019 Telstra Open Source
  *
  *   Licensed under the Apache License, Version 2.0 (the "License");
  *   you may not use this file except in compliance with the License.
@@ -13,21 +13,18 @@
  *   limitations under the License.
  */
 
-package org.openkilda.persistence.repositories;
+package org.openkilda.northbound.dto.switches;
 
-import org.openkilda.model.Switch;
-import org.openkilda.model.SwitchId;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Value;
 
-import java.util.Optional;
+@Value
+public class DeleteSwitchResult {
 
-public interface SwitchRepository extends Repository<Switch> {
-    boolean exists(SwitchId switchId);
+    @JsonProperty("deleted")
+    private boolean deleted;
 
-    Optional<Switch> findById(SwitchId switchId);
-
-    Switch reload(Switch entity);
-
-    void lockSwitches(Switch... switches);
-
-    void forceDelete(SwitchId switchId);
+    public DeleteSwitchResult(@JsonProperty("deleted") boolean deleted) {
+        this.deleted = deleted;
+    }
 }
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchDeleteSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchDeleteSpec.groovy
@@ -1,0 +1,281 @@
+package org.openkilda.functionaltests.spec.switches
+
+import static org.openkilda.testing.Constants.WAIT_OFFSET
+
+import org.openkilda.functionaltests.BaseSpecification
+import org.openkilda.functionaltests.helpers.Wrappers
+import org.openkilda.messaging.info.event.IslChangeType
+import org.openkilda.messaging.info.event.IslInfoData
+import org.openkilda.messaging.payload.flow.FlowState
+import org.openkilda.model.SwitchId
+import org.openkilda.northbound.dto.links.LinkParametersDto
+import org.openkilda.testing.model.topology.TopologyDefinition.Switch
+import org.springframework.web.client.HttpClientErrorException
+
+class SwitchDeleteSpec extends BaseSpecification {
+    def "Cannot delete nonexistent switch"() {
+        given: "Nonexistent switch ID"
+        def switchId = new SwitchId("de:ad:be:ef:99:99:99:99")
+
+        when: "Try to delete nonexistent switch"
+        northbound.deleteSwitch(switchId, false)
+
+        then: "Got 404 NotFound"
+        def exc = thrown(HttpClientErrorException)
+        exc.rawStatusCode == 404
+    }
+
+    def "Cannot delete active switch"() {
+        given: "An active switch"
+        def switchId = topology.getActiveSwitches()[0].dpId
+
+        when: "Try to delete active switch"
+        northbound.deleteSwitch(switchId, false)
+
+        then: "Got 400 BadRequest because switch must be deactivated"
+        def exc = thrown(HttpClientErrorException)
+        exc.rawStatusCode == 400
+        exc.responseBodyAsString.contains("Switch '$switchId' is in 'Active' state")
+    }
+
+    def "Cannot delete inactive switch with active ISL"() {
+        requireProfiles("virtual")
+
+        given: "An inactive switch with active ISL"
+        def isl = northbound.activeLinks[0]
+        def switchId = isl.source.switchId
+        lockKeeper.knockoutSwitch(switchId)
+        Wrappers.wait(WAIT_OFFSET) {
+            !northbound.activeSwitches.any { it.switchId == switchId}
+        }
+
+        when: "Try to delete inactive switch with active ISL"
+        northbound.deleteSwitch(switchId, false)
+
+        then: "Got 400 BadRequest because switch has active ISL"
+        def exc = thrown(HttpClientErrorException)
+        exc.rawStatusCode == 400
+        exc.responseBodyAsString.matches(".*Switch '$switchId' has \\d+ active links\\. Unplug and remove them first.*")
+
+        and: "Cleanup: Activate the switch back"
+        lockKeeper.reviveSwitch(switchId)
+        Wrappers.wait(WAIT_OFFSET) {
+            northbound.activeSwitches.any { it.switchId == switchId }
+        }
+    }
+
+    def "Cannot delete inactive switch with inactive ISLs"() {
+        requireProfiles("virtual")
+
+        given: "An inactive switch with inactive ISLs"
+        def switchId = topology.getActiveSwitches()[0].dpId
+        // get all active ISLs on switch
+        def activeSwitchLinks = northbound
+                .getAllSwitchLinks(switchId)
+                .findAll { it.state == IslChangeType.DISCOVERED }
+        // deactivate all active ISLs on switch
+        downLinkPorts(activeSwitchLinks)
+        Wrappers.wait(WAIT_OFFSET) {
+            northbound.getAllSwitchLinks(switchId).every { it.state != IslChangeType.DISCOVERED }
+        }
+
+        // deactivate switch
+        lockKeeper.knockoutSwitch(switchId)
+        Wrappers.wait(WAIT_OFFSET) {
+            !northbound.activeSwitches.any { it.switchId == switchId }
+        }
+
+        when: "Try to delete inactive switch with inactive ISLs"
+        northbound.deleteSwitch(switchId, false)
+
+        then: "Got 400 BadRequest because switch has inactive ISLs"
+        def exc = thrown(HttpClientErrorException)
+        exc.rawStatusCode == 400
+        exc.responseBodyAsString.matches(".*Switch '$switchId' has \\d+ inactive links\\. Remove them first.*")
+
+        and: "Cleanup: Activate the switch back"
+        lockKeeper.reviveSwitch(switchId)
+        Wrappers.wait(WAIT_OFFSET) {
+            northbound.activeSwitches.any { it.switchId == switchId }
+        }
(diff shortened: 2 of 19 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchSpec.groovy
```diff
@@ -24,4 +33,118 @@ class SwitchSpec extends BaseSpecification {
...
+        given: "An active switch"
+        def sw = topology.getActiveSwitches()[0]
+
+        when: "Try to force delete active switch with ISLs"
+        def response = northboundService.deleteSwitch(sw.getDpId(), true)
+
+        then: "Check that switch is actually deleted"
+        response.deleted
+        Wrappers.wait(WAIT_OFFSET) {
+            !northboundService.getAllSwitches().stream().anyMatch({ s -> (s.switchId == sw.dpId) })
```
rtretyak: This is just fine, no need to change, but the more groovysh way would be to 
```
!northboundService.allSwitches.any { it.switchId == sw.dpId } 
//or .find { it.switchId == sw.dpId }
```

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchSpec.groovy
```diff
@@ -24,4 +33,118 @@ class SwitchSpec extends BaseSpecification {
...
+        }
+
+        and: "Cleanup: Restore the switch"
+        lockKeeperService.reviveSwitch(sw.getDpId())
+        Wrappers.wait(WAIT_OFFSET) {
+            northboundService.getAllSwitches().stream().anyMatch({ s -> (s.switchId == sw.dpId) })
+        }
+    }
```
rtretyak: missing positive case without force delete?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchSpec.groovy
```diff
@@ -24,4 +33,118 @@ class SwitchSpec extends BaseSpecification {
         def exc = thrown(HttpClientErrorException)
         exc.rawStatusCode == 400
     }
+
+    def "Delete nonexistent switch"() {
```
rtretyak: If possible, name the test with what expected behavior is being tested, instead of what actions are being done.
"Delete active switch" -> "Cannot delete active switch"
"Force switch delete" -> "Can delete active switch if using force delete"

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchSpec.groovy
```diff
@@ -24,4 +33,118 @@ class SwitchSpec extends BaseSpecification {
...
+        exc.rawStatusCode == 400
+    }
+
+    def "Delete inactive switch with ISLs"() {
+        requireProfiles("virtual")
+
+        given: "An active switch"
+        def sw = topology.getActiveSwitches()[0]
+
+        and: "Deactivate this switch"
```
rtretyak: Currently this better fits for a 'when' block description, not 'given'. 
Should probably be sth like
```
given: "Deactivated switch"
when: "Try to delete inactive switch with ISLs"
```
or
```
given: "An active switch"
when: "Deactivate this switch"
and: "Try to delete inactive switch with ISLs"
```

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchSpec.groovy
```diff
@@ -24,4 +33,118 @@ class SwitchSpec extends BaseSpecification {
...
+        def response = northboundService.deleteSwitch(sw.getDpId(), true)
+
+        then: "Check that switch is actually deleted"
+        response.deleted
+        Wrappers.wait(WAIT_OFFSET) {
+            !northboundService.getAllSwitches().stream().anyMatch({ s -> (s.switchId == sw.dpId) })
+        }
+
+        and: "Cleanup: Restore the switch"
+        lockKeeperService.reviveSwitch(sw.getDpId())
```
rtretyak: I'm wondering how this works. You didn't do `knockoutSwitch` in this test, so the switch theoretically is still connected to controller, so reviveSwitch should do nothing. Or does forced deleteSwitch operation also disconnects switch from controller?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchSpec.groovy
```diff
@@ -24,4 +33,118 @@ class SwitchSpec extends BaseSpecification {
...
+
+        and: "Deactivate this switch"
+        lockKeeperService.knockoutSwitch(sw.getDpId())
+
+        when: "Try to delete inactive switch with flow"
+        northboundService.deleteSwitch(sw.getDpId(), false)
+
+        then: "Got 400 BadRequest because switch has flow"
+        def exc = thrown(HttpClientErrorException)
+        exc.rawStatusCode == 400
```
rtretyak: It is not clear whether it is not removed due to existing flow or due to active ISLs being still present on the switch. In order to check that the reason is indeed the existing flow we can 
1. Verify the error message in case if it differs for isls/flows cases. It should state the 'flow' reason, not 'isl'
or
2. Eliminate the ISL factor by forcing all ISLs on this switch to down state.

same for the two below tests with flows

rtretyak:
Thanks for nice functional-tests :+1:

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchSpec.groovy
```diff
@@ -24,4 +33,118 @@ class SwitchSpec extends BaseSpecification {
...
+
+        and: "Cleanup: Activate the switch back and remove flow"
+        lockKeeperService.reviveSwitch(sw.getDpId())
```
rtretyak: I see the job https://jenkins.pn.telstra.com/view/Kilda/job/kilda-ATDD-runner/1735/console is failing on the next test because it cannot build flow through this switch. After reviving switch you should probably wait until it is 'discovered' and appears as ACTIVATED in NB, otherwise the next test will run too early, before the switch actually appears in the system

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchSpec.groovy
niksv: deleteSwitch just removes data from database. In real world switch should be disconnected from controller before delete. Force delete is just a way to delete switch without any restrictions. May be it will crash something in future. So it is not regular operation. Maybe reviveSwitch() do nothing on controller, but it restores the switch in database.  And it brings back topology on original state. This is the main purpose

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchSpec.groovy
niksv: yes( I'm still trying to find a way to remove inactive ISLs from DB and restore it after

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchSpec.groovy
niksv: I'll try to do it, but 
1. For some reasons `exc` has only "400 " error message. (Real http request returns correct error messages)
2. ISL in down state is not enough. I need to remove ISL from DB. At this moment I'm trying to find a way to remove and restore ISL in DB.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchSpec.groovy
rtretyak: Correct me if I'm wrong, but I think it should require the ISL to be in FAILED status (not necessary to remove it from DB entirely, just deactivate it). At least I would expect it to work this way

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchSpec.groovy
rtretyak: Found the updates in #328 . In this case you'll have to delete ISL from DB. We don't have such method atm, but this should be an easy fix in  org.openkilda.testing.service.database.DatabaseNeoImpl

rtretyak: 1. You should be able to do `exc.message == "whatever"`

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchSpec.groovy
```diff
@@ -24,4 +33,125 @@ class SwitchSpec extends BaseSpecification {
...
+        then: "Check that switch is actually deleted"
+        response.deleted
+        Wrappers.wait(WAIT_OFFSET) {
+            !northboundService.allSwitches.any { it.switchId == sw.dpId }
+        }
+
+        and: "Cleanup: Restore the switch"
+        lockKeeperService.reviveSwitch(sw.getDpId())
+        Wrappers.wait(WAIT_OFFSET) {
+            northboundService.allSwitches.any { it.switchId == sw.dpId }
```
rtretyak: This still affects subsequent tests. I'm wondering whether you should wait for it to have ACTIVATED status and wait until all related ISLs are DISCOVERED (this may require at most `kilda_discovery_interval` seconds to discover an ISL)

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchSpec.groovy
niksv: `exc.responseBodyAsString` helped! thanks

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchSpec.groovy
rtretyak: Discussed with @surabujin . The switch should not have **any** ISLs before deletion (just like it is implemented now). So for now the only way is to remove an ISL directly in DB

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchSpec.groovy
niksv: okay, lets wait ISLs

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchSpec.groovy
niksv: I rebased on Delete ISL pull request so now we can remove ISLs via API

niksv:
Work is stopped till PL #1448 will be merged

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchSpec.groovy
```diff
@@ -0,0 +1,294 @@
+package org.openkilda.functionaltests.spec.northbound.switches
```
ylobankov: You should rebase your tests on test code from the develop branch.

PR #1456, opened 2018-10-17
Bump java version up to 8u191

surabujin:
8u181 was removed from oracle servers and make our install process
corrupted. Bump version to 8u191 using sed magic.


surabujin:
Upstream have fixed java installer.

Issue #1451, opened 2018-10-16
Allow Path Computation to exclude a given graph


pzakatov:
Add ability to exclude a given graph from path computation algorithm. 
Exclusion graph could be:
- an ISL
- a switch
- a path (set of Switches and ISLs)

Scope:
- create a design
- implement it :)

Next tasks: 
- #1231 associate flows and make them diverse
- #1232 Introduce a concept of a backup path

Possible use-cases:
- Calculate a diverse path
- Evacuate flows from ISL/Switch (reroute with exclude)
- Allow user to specify exclusion list from API

IlyaMoiseev:
Closed because another way of graph manipulation will be used in dependent tasks

PR #1460, opened 2018-10-18
PCE exclude subgraph design

IlyaMoiseev:
Design for #1451


docs/design/solutions/pce-exclude-subgraph/pce-exclude-subgraph.md
```diff
@@ -0,0 +1,19 @@
...
+## Goals
+Add general excluding mechanics in Computation Engine. It will be used in scenarios like:
+- Calculate a diverse path
+- Evacuate flows from ISL/Switch (reroute with exclude)
+- Allow user to specify exclusion list from API
+
+## Solution overview
+There are two rules of excluding
+- Exclude specified ISL
+- Exclude entire switch
```
surabujin: Excluding switch you exclude all ISL attached to this switch. So switch exclusion is equal to ISL exclusion.

In other word we need to implement exclusion of N ISL from path computation.

IlyaMoiseev:
@surabujin I'm looking into current PCE with DAO branch, AvailableNetwork instance became internal there: https://github.com/telstra/open-kilda/blob/de9089f14c63db3d579b7efd3d77957ae4db3e05/services/src/kilda-pce/src/main/java/org/openkilda/pce/impl/PathComputerImpl.java#L69
Maybe it requires some refactoring for this feature

docs/design/solutions/pce-exclude-subgraph/pce-exclude-subgraph.md
pzakatov: I'd vote for keeping switch filter and do a performance test afterward, so we can decide later. Having this filtering logic inside PCE will simplify external business logic a bit.

surabujin: We can and should have "exclude switch" into NB API.

The possibility to convert "exclude switch" into "exclude N ISL" is our implementation details and external world should not worry/know about them.

IlyaMoiseev:
Diagrams updated

PR #1465, opened 2018-10-18
Hub & spoke storm implementation

surabujin:
Collect previous discussion materials about hub-and-spoke approach
implementation into storm.


surabujin:
Became obsolete because of  H&S release (mark7+).

PR #1466, opened 2018-10-19
The revised data model with encapsulation support.

sergii-iakovenko:
The changes in the model:
- Flow entity has been decomposed into 3 entities: Flow (basic flow information, e.g. src and dest endpoints, status, description, encapsulationType), UniDirectionFlow (represents a single direction of the flow, so it has a cookie in properteis), TransitVlan (a part of implementation of vlan-based encapsulation).
- Flow has encapsulationType property.
- FlowSegment has been renamed to PathSegment. It represents a flow path over the network without any encapsulation specific details.
- Single ISL entity represents a duplex link.


sergii-iakovenko:
Closing as the model has been updated and supplied with PoC in #2002.

PR #1469, opened 2018-10-19
Fixed issues with incorrect response for sync rules and delete link p…

nikitacherevko:
…rops

### DIFF
--- a/services/src/northbound/src/main/java/org/openkilda/northbound/service/impl/LinkServiceImpl.java
+++ b/services/src/northbound/src/main/java/org/openkilda/northbound/service/impl/LinkServiceImpl.java
@@ -15,6 +15,9 @@
 
 package org.openkilda.northbound.service.impl;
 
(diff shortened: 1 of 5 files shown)


services/src/northbound/src/main/java/org/openkilda/northbound/service/impl/LinkServiceImpl.java
```diff
@@ -108,72 +111,53 @@
...
-                    }
-                }
+            pendingRequest.add(messagingChannel.sendAndGetChunked(topologyEngineTopic, message));
+        }
+
+        return collectChunkedResponses(pendingRequest, LinkPropsResponse.class)
+                .thenApply(responses -> getDelLinkPropsResult(responses, new ArrayList<>()));
+    }
+
+    private BatchResults getDelLinkPropsResult(List<LinkPropsResponse> responses, List<String> errors) {
```
surabujin: this method is uses in "set" and in "del" operations, it should not containt "del" chunk in it.

PR #1476, opened 2018-10-22
Fixed deletion link props

nikitacherevko:


services/topology-engine/queue-engine/topologylistener/message_utils.py
```diff
@@ -369,16 +368,20 @@ def send_link_props_response(payload, correlation_id, chunked=False):
...
-    send_to_topic(
-        None, correlation_id, MT_INFO_CHUNKED,
-        destination='NORTHBOUND', topic=config.KAFKA_NORTHBOUND_TOPIC,
-        extra={'next_request_id': None})
+            extra={
+                'message_id': correlation_id,
+                'total_messages': 0
+            })
+    else:
+        for idx, payload in batch:
```
surabujin: it should be: `for idx, payload in enumerate(batch)`

PR #1480, opened 2018-10-22
Add test to verify default rules are installed on a new switch

ylobankov:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
```diff
@@ -0,0 +1,102 @@
...
+        when: "Connect the switch to the controller"
+        lockKeeperService.reviveSwitch(sw.dpId)
+
+        then: "The switch is really connected to the controller"
+        Wrappers.wait(WAIT_OFFSET) { sw.dpId in northboundService.getActiveSwitches()*.switchId }
+
+        and: "Default rules are installed on the switch"
+        def actualRulesSorted = northboundService.getSwitchRules(sw.dpId).flowEntries.sort { it.cookie }
+        actualRulesSorted.size() == defaultRulesSorted.size()
+        [0..actualRulesSorted.size() - 1].each { i ->
```
rtretyak: There is a more pretty way to loop through 2 lists
```
[actualRules, defaultRules].transpose().each { actual, expected ->
  actual.cookie == expected.cookie
  //etc.
}
```

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
```diff
@@ -0,0 +1,102 @@
...
+                    instructions == defaultRulesSorted[i].instructions
+                    flags == defaultRulesSorted[i].flags
+                }
+            }
+        }
+    }
+
+    def "Pre-installed rules are not deleted from a new switch when connecting it to the controller"() {
+        requireProfiles("virtual")
+        assumeTrue("Test is skipped because of the issue #1464. Also the current disconnection of the switch from " +
```
rtretyak: This is better to be described in `@Ignore("why?")`. Assumptions better fit for things that can only be computed at runtime

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
ylobankov: Thanks a lot!

Issue #1482, opened 2018-10-23
Refactor dublicate methods in FL


Levitan:
Use in SwitchManager method dpidToMac() from SwitchUtils

PR #1483, opened 2018-10-23
Refactor SwitchManager

Levitan:
Issue #1482


services/src/floodlight-modules/src/test/java/org/openkilda/floodlight/service/ping/PingServiceTest.java
```diff
@@ -45,26 +50,33 @@ public void setUp() throws Exception {
...
         moduleContext.addConfigParam(pathVerificationService, "hmac256-secret", "secret");
     }
 
     @Test
     public void wrapUnwrapCycle() throws Exception {
+        DatapathId dpIdAlpha = DatapathId.of(0xfffe000000000001L);
+        DatapathId dpIdBeta = DatapathId.of(0xfffe000000000002L);
+        MacAddress macBeta = MacAddress.of(0xfffe000000000002L);
+
+        moduleContext.getServiceImpl(ISwitchManager.class)
```
surabujin: em... what kind of magic is happening here?

Levitan: This part of test, need for simulate work method ` ISwitchManager.dpIdToMac(DatapathId)`.

surabujin: SwitchManager returned by moduleContext is mock object... I can suspect that your magic try to schedule method call... but without setup return value for some reason.

On the next line, you pass in expect real SwitchManager object and setup return value ... for "some" mock (switchManager is not mock in this case)... 

Whole manipulation is at least incorrect mock usage.

services/src/floodlight-modules/src/test/java/org/openkilda/floodlight/service/ping/PingServiceTest.java
```diff
@@ -45,26 +48,32 @@ public void setUp() throws Exception {
 
         moduleContext.addService(IOFSwitchService.class, createMock(IOFSwitchService.class));
         moduleContext.addService(InputService.class, createMock(InputService.class));
+        moduleContext.addService(ISwitchManager.class, createMock(ISwitchManager.class));
```
surabujin: Why not real SwitchManager?

Issue #1472, opened 2018-10-19
Add possibility to create one-switch flow for one port, but different vlans


nikitacherevko:
We should create another type of install rule command in case when user wants create a one-switch flow for one port. Here is some details:
https://mailman.stanford.edu/pipermail/openflow-discuss/2015-April/005636.html

NB: also need to extend validation to check that vlans differ from each other.

PR #1488, opened 2018-10-23
Added possibility to create one-switch flow for the same port

nikitacherevko:


services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -1036,6 +1040,17 @@ private OFAction actionSetOutputPort(final OFFactory ofFactory, final int output
         return actions.buildOutput().setMaxLen(0xFFFFFFFF).setPort(OFPort.of(outputPort)).build();
     }
 
+    /**
+     * Create an OFAction which sets the output port as in_port.
+     *
+     * @param ofFactory OF factory for the switch
+     * @return {@link OFAction}
+     */
+    private OFAction actionSetOutputInPort(final OFFactory ofFactory) {
```
surabujin: Too many duplication... I think it will be better to change type of second argument of `actionSetOutputPort` from `int` to `OFPort`. And use it for both cases.

surabujin:
There is one more issue here - controller do not remove rules from switch on flow remove. Altered rule do not match passed remove criteria.

It bring the question - do we really want to support "full match" on delete operations? We should be able to relly on cookie value only (perhaps we should dedicate several bits 2-4 for some kilda-identifier into cookie value to guarantee "kilda-rules only" match).

services/topology-engine/queue-engine/topologylistener/message_utils.py
```diff
@@ -189,15 +189,18 @@ def build_one_switch_flow_from_db(switch, stored_flow, output_action):
...
-    flow.switch_id = switch
-    flow.meter_id = meter_id
-    flow.criteria = {'cookie': cookie, 'in_port': in_port, 'in_vlan': in_vlan, 'out_port': out_port}
+    flow.flowid = str(node['flow_id'])
+    flow.cookie = node['cookie']
+    flow.switch_id = str(node['switch_id'])
+    flow.meter_id = node['meter_id']
+
+    flow.criteria = {'cookie': node['cookie'], 'in_port': node['in_port'], 'in_vlan': node['in_vlan']}
+    if node['in_port'] != node['out_port']:
```
surabujin: To avoid confusion from this check we should add switch_id equality check there.

services/topology-engine/queue-engine/topologylistener/message_utils.py
```diff
@@ -189,15 +189,18 @@ def build_one_switch_flow_from_db(switch, stored_flow, output_action):
     return flow
 
 
-def build_delete_flow(switch, flow_id, cookie, meter_id, in_port, in_vlan, out_port):
+def build_delete_flow(node):
```
surabujin: `node` bad name... there is no associations for it in our domain model. It should be  something like `flow_segments` or `path_segment`.

services/wfm/src/test/java/org/openkilda/wfm/topology/flow/FlowTopologyTest.java
```diff
@@ -933,7 +933,7 @@ private Flow createFlow(final String flowId) throws IOException {
         System.out.println("NORTHBOUND: Create flow");
         Flow flowPayload =
                 new Flow(flowId, 10000, false, "", new SwitchId("ff:00"), 1, 2,
-                        new SwitchId("ff:00"), 1, 2);
+                        new SwitchId("ff:01"), 1, 2);
```
surabujin: ?

nikitacherevko: that's because tests try to create flow between two same endpoints (switch dpid, port, vlan) and I've just changed destination dpid.

services/topology-engine/queue-engine/topologylistener/message_utils.py
nikitacherevko: I'll add a comment instead as discussed

nikitacherevko:
closes #1472

PR #1499, opened 2018-10-26
Changed the way mininet knocks out switch

rtretyak:
This way when knocking out switch all the rules will remain installed

### DIFF
--- a/services/mininet/kilda/mininet/flow_tool.py
+++ b/services/mininet/kilda/mininet/flow_tool.py
@@ -28,6 +28,8 @@ from functools import wraps
 
 logger = logging.getLogger()
 
+DUMMY_CONTROLLER = "tcp:127.0.1.2:20000"
+
 
 def log_to_logger(fn):
     '''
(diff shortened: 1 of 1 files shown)


services/mininet/kilda/mininet/flow_tool.py
```diff
@@ -131,7 +131,9 @@ def traffic_listener(traffic_goes_through, vlanid, link):
 @post("/knockoutswitch")
 @required_parameters("switch")
 def switch_knock_out(p):
-    result = os.system("ovs-vsctl del-controller %s" % p['switch'])
+    dummy_controller = "tcp:192.0.2.0:6666"
```
ylobankov: Probably it is better to place it as a constant at the top of this file

services/mininet/kilda/mininet/flow_tool.py
```diff
@@ -28,6 +28,8 @@
 
 logger = logging.getLogger()
 
+DUMMY_CONTROLLER = "tcp:192.0.2.0:6666"
```
surabujin: Better set something like "127.0.1.2:20000"

PR #1500, opened 2018-10-26
Functional tests/flow validation negative tests

npliashechnikov:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowValidationNegativeSpec.groovy
```diff
@@ -0,0 +1,169 @@
...
+        }
+        where:
+        rule << ["ingress", "transit", "egress"]
+        item << [0, 1, -1]
+    }
+
+    @Unroll
+    def "Both flow and switch validation should fail if flow rule is missing with #flowconfig flow configuration"() {
+        given: "A ${flowconfig} flow exists"
+        def (src, dest) = method()
```
rtretyak: wat

services/src/functional-tests/topology.yaml
```diff
@@ -1,57 +1,57 @@
 switches:
     - name: ofsw1
-      dp_id: 00:00:00:00:00:00:00:01
+      dp_id: 00:00:00:22:3d:6b:00:04
```
rtretyak: ?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowValidationNegativeSpec.groovy
```diff
@@ -0,0 +1,169 @@
...
+        Flow forward = database.getFlow(flow.id).left
+        def flowRules = findFlowRules(forward.cookie, involvedSwitches)
+
+        when: "A flow rule gets deleted"
+        def cookie = flowRules[flowRules.keySet()[0]][0].cookie.toString()
+        northbound.deleteSwitchRules(flowRules.keySet()[0], cookie)
+
+        then: "Flow and switch validation should fail"
+        !northbound.validateFlow(flow.id).every {it -> it.asExpected}
+        northbound.validateSwitchRules(flowRules.keySet()[0]).missingRules.size() > 0
```
rtretyak: both these verifications are very weak. why `> 0`? It should have exactly 1 missing rule, moreover you can even assert its cookie value

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowValidationNegativeSpec.groovy
```diff
@@ -0,0 +1,169 @@
...
+        cleanup: "Delete the flow"
+        if (flow) {
+            northbound.deleteFlow(flow.id)
+        }
+        where:
+        flowconfig << ["single switch", "neighbouring", "transit"]
+        method << [{ getSingleSwitch() }, { getNeighbouringSwitches() }, { getNonNeighbouringSwitches() }]
```
rtretyak: why all methods are wrapped into closure? The only reason for this would be to delay the method calls until test execution, but I don't really see any reason for that. These methods should run with no issues in `where` block directly

rtretyak: I don't really see the difference between neighboring and non-neighboring switches cases, since you always delete the rule on the first switch. So in third case you say 'transit', but that doesn't seem to be true

PR #1502, opened 2018-10-26
Toolset to detect switch features

### LINKED ISSUE #1418: Do not reuse "port up" events during switch activation
During switch activation FL produce "port up" event for each "enabled" port. In other words - we mix two complete different event - "port become up" and "port is up during switch connect".

As result we have an "extra" "port up" events that tangle up ISL/flow management logic. 

Switch activation event must include list of all switch port and their up/down status. And only real port up must produce port up event.

### LINKED ISSUE #1484: Implement switch features detector
To be able to use "optional" features available on some subset of switches we must "detect" switch features and report them into "management code" (in our case one of storm topology).

Most obvious place for feature detection code is FL - because it have all possible info about switch.

Part of #569

surabujin:
To propagate switch features new model objects(Switch, SwitchPort) are
added into messaging package.

Alter NetworkDumpSwitchData object. Now it carry Switch model object.
This object carry switch features and info about all physical ports.

SwitchInfoData was extened to carry additional switch model object.
Existing "protocol" between FL and event topology become more and more
unhandy. One of the problem - SwitchInfoData object are used far beyoud
discovery process and due to this usage there is too many limits for
possible changes in it. That why we do not replace SwitchInfoData object
with something else for "SWITCH ADD" notification.

Close #1484
Close #1418


surabujin:
Depends on #1705

services/src/messaging/src/main/java/org/openkilda/messaging/model/Switch.java
```diff
@@ -0,0 +1,58 @@
...
+import com.google.common.collect.ImmutableSet;
+import lombok.Builder;
+import lombok.Value;
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.Set;
+
+@Value
+public class Switch implements Serializable {
```
sergii-iakovenko: Can we avoid a conflict in naming of Switch in the general data model and in Floodlight model? Why not extend SwitchInfoData?

PR #1508, opened 2018-10-29
More verbose switch sync logging

surabujin:
Add __str__ method for object's produced by message_utils.py. Also add
test for switch sync request.

### DIFF
--- a/services/topology-engine/queue-engine/topologylistener/message_utils.py
+++ b/services/topology-engine/queue-engine/topologylistener/message_utils.py
@@ -37,8 +37,12 @@ MI_LINK_PROPS_RESPONSE = (
 
 
 class Abstract(model.JsonSerializable):
-    def to_json(self):
-        return json.dumps(self, cls=model.JSONEncoder)
+    def __str__(self):
+        return '<{}: {}>'.format(type(self).__name__, self.to_json(
+            sort_keys=True))
+
+    def to_json(self, sort_keys=False):
+        return json.dumps(self, cls=model.JSONEncoder, sort_keys=sort_keys)
 
 
 class Flow(Abstract):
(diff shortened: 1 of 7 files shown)


services/topology-engine/queue-engine/topologylistener/message_utils.py
```diff
@@ -37,8 +37,15 @@
 
 
 class Abstract(model.JsonSerializable):
-    def to_json(self):
-        return json.dumps(self, cls=model.JSONEncoder)
+    def __str__(self):
+        return '<{}: {}>'.format(type(self).__name__, self.to_json(
+            uniformly=True))
+
+    def to_json(self, uniformly=False):
```
timofei-durakov: why not to name this flag as sort_keys and pass it as is to json.dumps?

services/topology-engine/queue-engine/topologylistener/messageclasses.py
```diff
@@ -238,7 +238,7 @@ def handle_link_props(self):
                 raise exc.NotImplementedError(
                     'link props request {}'.format(self.get_message_type()))
         except CypherSyntaxError as e:
-            logger.exception('Invalid request: ', e)
+            logger.exception('Request produce invalid cypher query')
```
timofei-durakov: do we hide Cypher exception here and do not have traceback?

nikitamarchenko: according to docs https://docs.python.org/2/library/logging.html#logging.Logger.exception we will have all exception data and traceback

services/topology-engine/queue-engine/topologylistener/message_utils.py
surabujin: Good question. :) Will change it in proposed way.

PR #1523, opened 2018-10-31
Update 'healthcheck' tests

ylobankov:
This patch creates a uniform health check for virtual and hardware envs.

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/BaseSpecification.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/BaseSpecification.groovy
@@ -1,14 +1,11 @@
 package org.openkilda.functionaltests
 
 import static org.junit.Assume.assumeTrue
-import static org.openkilda.testing.Constants.WAIT_OFFSET
 
 import org.openkilda.functionaltests.extension.fixture.SetupOnce
 import org.openkilda.functionaltests.extension.healthcheck.HealthCheck
 import org.openkilda.functionaltests.helpers.FlowHelper
-import org.openkilda.functionaltests.helpers.Wrappers
 import org.openkilda.messaging.info.event.IslChangeType
-import org.openkilda.messaging.payload.flow.FlowState
 import org.openkilda.testing.model.topology.TopologyDefinition
 import org.openkilda.testing.service.floodlight.FloodlightService
 import org.openkilda.testing.service.northbound.NorthboundService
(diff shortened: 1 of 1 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/BaseSpecification.groovy
```diff
@@ -50,24 +47,25 @@ class BaseSpecification extends SpringSpecification implements SetupOnce {
...
         def links = northbound.getAllLinks()
         verifyAll {
-            northbound.activeSwitches.size() == topologyDefinition.activeSwitches.size()
-            northbound.allFlows.empty
-            northbound.allLinkProps.empty
             links.findAll {
                 it.state == IslChangeType.DISCOVERED
             }.size() == topologyDefinition.islsForActiveSwitches.size() * 2
             links.findAll { it.state == IslChangeType.FAILED }.empty
+            links.findAll { it.availableBandwidth != it.speed }.empty
```
rtretyak: This verification was intentionally put to the 'seconds phase of verifications', because this is a consequence of having flows in the system. My thought was, that it doesn't make sense to inform about bw!=speed if there a flows in the system (same about rules) -> second phase should be reported only after 'basic' things like flows in the system are fixed

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/BaseSpecification.groovy
```diff
@@ -77,22 +75,4 @@ class BaseSpecification extends SpringSpecification implements SetupOnce {
             }.empty
         }
     }
-
-    @HealthCheck
-    def "Create casual flow to ensure Kilda's up state"() {
```
rtretyak: I don't see these verifications to be replaced somehow. It used to create a flow and wait for `WAIT_OFFSET * 3` (longer than always) seconds to ensure that Kilda is completely Up after being deployed (**Jenkins case**).
On the other hand, this test was added when we tried to increase stability and I can't claim that it actually increased it. +it may no longer be required since we increased the sleep after kilda deploy from 1 to 3 minutes (but this is still under test and investigation)

ylobankov: I think it is just not needed now. In my opinion, if Kilda answers us successful responses when getting all links, switches, flows and link props (all of that are made above), it should be definitely UP. If it is not, so the corresponding bug should be filed I think.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/BaseSpecification.groovy
```diff
@@ -50,24 +47,31 @@ class BaseSpecification extends SpringSpecification implements SetupOnce {
...
-            northbound.allLinkProps.empty
+            links.findAll { it.state == IslChangeType.FAILED }.empty
             links.findAll {
                 it.state == IslChangeType.DISCOVERED
             }.size() == topologyDefinition.islsForActiveSwitches.size() * 2
-            links.findAll { it.state == IslChangeType.FAILED }.empty
         }
 
-        and: "More detailed inspection also reports no issues"
+        and: "No flows and link props are present"
```
rtretyak: It's fine, but it is not the same as it was before and I don't find it to be an improvement. Previously it would report you about all problems at once (except for second round of verifications which are not very typical to fail). Right now you will have to fix links, then rerun, then fix flows, then rerun again.

ylobankov: Fixed.

PR #1535, opened 2018-11-01
Error/failure output in tests vastly improved

rtretyak:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/Wrappers.groovy
```diff
@@ -34,33 +35,38 @@ class Wrappers {
...
                 }
             } catch (Throwable t) {
-                log.debug("Wait check threw exception: ${t.message}.\nSleeping for ${sleepTime} ms...")
+                thrown = t
                 sleep(sleepTime)
             }
         }
-        log.debug("Closure failed to succeed within ${timeout} seconds, aborting.")
-        return false
+        def message = "Condition didn't pass in $timeout seconds"
```
ylobankov: I would prefer such a variant:

Condition was not satisfied within ${timeout} seconds

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/internal/spec/WrappersSpec.groovy
```diff
@@ -0,0 +1,32 @@
...
+
+import spock.lang.Specification
+import spock.lang.Unroll
+
+@Unroll
+class WrappersSpec extends Specification {
+
+    def timeout = 0.001
+
+    def "Wrappers.wait throws if condition fails"() {
```
ylobankov: Wrappers.wait throws exception ...

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/internal/spec/WrappersSpec.groovy
```diff
@@ -0,0 +1,32 @@
+package org.openkilda.functionaltests.internal.spec
```
ylobankov: Let's use the more common name for such kind of tests.

internal -> unit

services/src/functional-tests/pom.xml
```diff
@@ -110,6 +110,18 @@
                             <skip>true</skip>
                         </configuration>
                     </execution>
+                    <execution>
+                        <id>tools-test</id>
```
ylobankov: IMHO 'unit' instead of 'tools' will be better to use.

PR #1536, opened 2018-11-01
Added random part to flow name to prevent flows with the same name

rtretyak:


services/src/test-library/pom.xml
```diff
@@ -22,6 +22,7 @@
         <nitorcreations-matchers.version>1.3</nitorcreations-matchers.version>
         <apache.httpclient.version>4.4.1</apache.httpclient.version>
         <docker.version>8.9.2</docker.version>
+        <faker.version>0.16</faker.version>
```
ylobankov: You are changing files in the functional-test module but dependency is added to test-library. Is it OK?

rtretyak: why not?

ylobankov: Ok, I was just curious.

PR #1433, opened 2018-10-05
Qa/kibana integration 2

npliashechnikov:
New effort to add ElasticSearch support to the test framework

PR #1538, opened 2018-11-01
Squash kibana integration commits

npliashechnikov:
Manually rebase and squash kibana integration commits with fixes from qa/kibana-integration-2 (https://github.com/telstra/open-kilda/pull/1433)

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/samples/KibanaSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/samples/KibanaSpec.groovy
@@ -1,0 +1,46 @@
+/* Copyright 2018 Telstra Open Source
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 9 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/samples/KibanaSpec.groovy
```diff
@@ -0,0 +1,46 @@
...
+    ElasticService elastic
+
+    def "Test framework should be able to extract logs from Storm Worker"() {
+        when: "Elastic Client is initialized"
+        assert elastic
+
+        then: "Retrieve all INFO+ level messages from Floodlight for last 5 minutes"
+        def logs = elastic.getLogs(new ElasticQueryBuilder().setAppId("kilda-floodlight").setTimeRange(300).build())
+        logs
+        logs?.hits?.total > 0
```
rtretyak: 1. Will I always have to access it in a null-safe manner? Why?
2. Is there ability to have at least top-level DTO object for the response? At this point I don't really understand how does the response look like and in order to implement a test I'll have to turn on debug mode and look into what is returned.

npliashechnikov: 1. As long as you check it once, it is no longer needed.
2. Unfortunately, Elastic doesn't have defined responce schema - the return format may vary for each component and even for each message. I will try and define some DTOs in next PR where I'll check logging for each Kilda component.

PR #1539, opened 2018-11-01
Fix ATDD checkstyle.

Settis:


services/src/atdd/src/main/java/org/openkilda/topo/TopoSlug.java
```diff
@@ -31,51 +27,91 @@
...
-    public static final String toString(Switch aSwitch) {
-        return "S." + aSwitch.getId();
+    /**
+     * Returns a string representation of the switch.
+     *
+     * @param theSwitch the switch.
+     * @return string representation.
+     */
+    public static final String toString(Switch theSwitch) {
+        return "S." + theSwitch.getId();
```
surabujin: A-Switch is our internal term. We do not have "the-switch" term.

services/src/atdd/src/main/java/org/openkilda/topo/MockController.java
```diff
@@ -13,30 +13,28 @@
...
-/**
- * 
- */
 package org.openkilda.topo;
 
 /**
- * @author carmine
+ * The mock controller.
  *
+ * @author carmine
```
surabujin: this is ... extra data. Remove it.

services/src/atdd/src/main/java/org/openkilda/topo/TopoSlug.java
Settis: Well, style checker ask to change `aSwitch` name. We need to think how to change style rules then.

Settis: The exception in stylecheck config was added.

PR #1558, opened 2018-11-07
Add helper method for flow deletion to functional-tests module

### LINKED ISSUE #1554: [Functional-tests] Add a helper method for deleting flow
We need a helper method for deleting a flow that will wait till all flow rules are deleted from switches. Also, it will be good to add some check to the existing helper method for creating switches where we will wait for rules that should be installed on switches. This should increase the stability of test executions.

ylobankov:
Closes #1554

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/FlowHelper.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/FlowHelper.groovy
@@ -1,20 +1,31 @@
 package org.openkilda.functionaltests.helpers
 
+import static org.openkilda.testing.Constants.RULES_DELETION_TIME
+
 import org.openkilda.messaging.payload.flow.FlowEndpointPayload
 import org.openkilda.messaging.payload.flow.FlowPayload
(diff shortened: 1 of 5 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/FlowHelper.groovy
```diff
@@ -57,6 +70,35 @@ class FlowHelper {
...
+     * Deletes flow with checking rules on source and destination switches.
+     * It is supposed if rules absent on source and destination switches, the flow is completely deleted.
+     */
+    FlowPayload deleteFlow(String flowId) {
+        def flowEntry = db.getFlow(flowId)
+
+        log.debug("Deleting flow '$flowId'")
+        def response = northbound.deleteFlow(flowId)
+
+        def cookie = [flowEntry.left.cookie, flowEntry.right.cookie]
```
rtretyak: Consider renaming to `cookies`, since it is an array and more natural to name it in plural

ylobankov: Fixed.

PR #1560, opened 2018-11-07
Rename 'reflow_on_switch_activation' feature

### LINKED ISSUE #1559: Misleading name of the feature that enables rerouting flows on ISL discovery
The current name of the feature that enables rerouting flows on ISL discovery is "reflow_on_switch_activation". This name doesn't reflect the proper meaning of this feature and should be reworked. If the feature is enabled, rerouting flows starts on ISL discovery, not on switch activation.

ylobankov:
The current name of the feature that enables rerouting flows on ISL discovery
is "reflow_on_switch_activation". This name doesn't reflect the proper meaning
of this feature and should be changed. If the feature is enabled, rerouting
flows starts on ISL discovery, not on switch activation. So this patch renames
the feature from "reflow_on_switch_activation" to "reroute_on_isl_discovery"
to get a more clear understanding what the feature does.

Closes #1559.


sergii-iakovenko:
Please, note that after per-H&s the feature toggles don’t work. See 
https://github.com/telstra/open-kilda/issues/1731

ylobankov:
Outdated.

PR #1578, opened 2018-11-12
Lowering limits for local containers

nikitamarchenko:
based on https://developers.redhat.com/blog/2017/03/14/java-inside-docker/


rtretyak:
Floodlight container is not starting in functional tests

PR #1581, opened 2018-11-12
Functional tests/logging tests

npliashechnikov:
DTOs for Elasticsearch and tests for Kilda logging facilities


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -1,9 +1,75 @@
...
+        try {
+            northbound.getFlow(flowId)
+        } catch (e) {
+            log.debug("Logging tests: omitting exception", e)
+        }
+        and: "Validate rules on a switch"
+        def switchId = topologyDefinition.switches[0].dpId
+        northbound.validateSwitchRules(switchId)
+        and: "Sleep for 60 seconds to allow Elastic to process logs"
+        sleep(60000)
```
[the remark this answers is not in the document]
npliashechnikov: Will implement the waiter for logs.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -1,9 +1,75 @@
...
+        setup: "Request a non-existent flow"
+        def flowId = "nonexistentflowid" + System.currentTimeMillis()
+        try {
+            northbound.getFlow(flowId)
+        } catch (e) {
+            log.debug("Logging tests: omitting exception", e)
+        }
+        and: "Validate rules on a switch"
+        def switchId = topologyDefinition.switches[0].dpId
+        northbound.validateSwitchRules(switchId)
```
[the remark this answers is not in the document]
npliashechnikov: We can do this, but I'd rather not, for a few reasons.
1. This will lead to scattering of the test coverage. Having a dedicated test scenario for logging and checking part of logging in another scenario that doesn't even mention logging looks kind of obscure.
2. Checking Elastic logs is a time consuming operation, and it will slow down "validate switch rules" test case. Here, it checks logs for 3 components at the same time to ensure minimal wait delays.

I will put clarification in the step decription.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -1,9 +1,75 @@
...
+
+        def nbResult = elasticService.getLogs(new ElasticQueryBuilder().setTags(nbTag).setTimeRange(300).
+                setLevel("ERROR").build())
+        def stormResult = elasticService.getLogs(new ElasticQueryBuilder().setTags(stormTag).setTimeRange(300).
+            setLevel("ERROR").build())
+        def tpResult = elasticService.getLogs(new ElasticQueryBuilder().setTags(tpTag).setTimeRange(300).
+                setLevel("INFO").build())
+
+        then: "Northbound should generate ERROR message about being unable to locate flow"
+        !nbResult.timedOut
```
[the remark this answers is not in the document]
npliashechnikov: Fixed by above :)

services/src/test-library/src/main/java/org/openkilda/testing/service/elastic/ElasticServiceImpl.java
```diff
@@ -53,50 +53,63 @@
...
+    public ElasticResponseDto getLogs(ElasticQuery query) {
         if (("".equals(query.appId) && "".equals(query.tags))) {
             throw new IllegalArgumentException("Either app_id or tags should be specified");
         }
-        Long time = System.currentTimeMillis();
+        //Uses multiple criteria bool filters, here we start to fill out criteria array
         String queryString = "";
 
-        if (!"".equals(query.appId)) {
+        if (query.appId  != null && !"".equals(query.appId)) {
```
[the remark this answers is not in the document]
npliashechnikov: Thanks, will use StringUtils

services/src/test-library/src/main/java/org/openkilda/testing/service/elastic/ElasticServiceImpl.java
```diff
@@ -53,50 +53,63 @@
...
+        //first criteria - multi-filter on _source field
+        ObjectNode fieldValueFilter = mapper.createObjectNode();
+        fieldValueFilter.set("query_string", mapper.createObjectNode().put("query", queryString)
+                .put("default_field", query.defaultField));
+        combinedQuery.add(fieldValueFilter);
+
         if (query.timeRange > 0) {
-            queryString += " AND timeMillis: [" + (time - query.timeRange * 1000) + " TO " + time + "]";
+            //second criteria - time range (if present)
+            //TODO: consider using timestamps instead of "now-time" approach, as it would simplify post-mortem analysis.
```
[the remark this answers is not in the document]
npliashechnikov: Will resolve in the next PR

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -1,9 +1,75 @@
...
+        def hits = result.hits.hits
+        hits.any {hit -> hit.source.message.toLowerCase().contains("push discovery package via")}
+    }
+
+    def "Check Northbound, Storm and Topology Engine logging"() {
+        setup: "Request a non-existent flow"
+        def flowId = "nonexistentflowid" + System.currentTimeMillis()
+        try {
+            northbound.getFlow(flowId)
+        } catch (e) {
```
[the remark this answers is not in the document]
rtretyak: you should catch HttpClientErrorException, not a generic `e` in order to not catch-ignore not-related problems. Also you may want to only ignore on 404 status code in the exception and rethrow for 5xx etc., but this may be an overkill

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -1,9 +1,74 @@
...
+        Wrappers.wait(timeout, 10) {
+            def nbLogs = elasticService.getLogs(new ElasticQueryBuilder().setTags(nbTag).setTimeRange(timeout * 2).
+                    setLevel("ERROR").build())
+            def stormLogs = elasticService.getLogs(new ElasticQueryBuilder().setTags(stormTag).
+                    setTimeRange(timeout * 2).setLevel("ERROR").build())
+            def tpLogs = elasticService.getLogs(new ElasticQueryBuilder().setTags(tpTag).setTimeRange(timeout * 2).
+                    setLevel("INFO").build())
+
+            if ([nbLogs, stormLogs, tpLogs].every {it?.hits?.total > 0})
+            {
```
rtretyak: please autoformat this. This is not a common braces style for Java.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -1,9 +1,74 @@
...
+                nbResult = nbLogs.hits.hits.any { hit -> hit.source.message.contains(flowId) }
+                stormResult = stormLogs.hits.hits.any { hit -> hit.source.message.contains(flowId) }
+                tpResult = tpLogs.hits.hits.any {hit -> hit.source.message.contains(switchId.toString())}
+            }
+            [nbLogs, stormLogs].every {it.hits.hits.any { hit -> hit.source.message.contains(flowId) }}
```
rtretyak: Sorry I don't understand why this operation is performed twice

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -1,9 +1,74 @@
...
+            return nbResult && stormResult && tpResult
+        }
+
+        then: "Northbound and Storm should generate ERROR message about being unable to locate flow"
+        nbResult
+        stormResult
+
+        and: "Topology Engine should post an INFO message with rule validation with the switch dpId"
+        tpResult
```
rtretyak: You should understand that these verifications make no sense, since your 'wait' block already succeeded at this point (otherwise it would have thrown WaitException). So these lines make no sense except for documentation purpose. This is a little bit not natural from my point of view, maybe this is personal, but I would expect your `when: "Northbound, Storm and Topology Engine logs for 5 minutes are retrieved"` to be rephrased and changed to a `then/expect:` block since it is actually performing the verification right now, it is not just getting logs.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
rtretyak: Moreover, the result of this operation is not saved and is literally ignored (this line does absolutely 100% nothing)

npliashechnikov: The line has slipped during refactoring

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
npliashechnikov: These blocks were initially intended to be part of documentation.
Moved the text to the asserts

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
npliashechnikov: Removed lambda completely

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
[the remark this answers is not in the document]
npliashechnikov: Fixed

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -1,9 +1,69 @@
...
+            def tpLogs = elasticService.getLogs(new ElasticQueryBuilder().setTags(tpTag).setTimeRange(timeout * 2).
+                    setLevel("INFO").build())
+
+            nbResult = nbLogs?.hits?.hits?.any { hit -> hit.source.message.contains(flowId) }
+            stormResult = stormLogs?.hits?.hits?.any { hit -> hit.source.message.contains(flowId) }
+            tpResult = tpLogs?.hits?.hits?.any {hit -> hit.source.message.contains(switchId.toString())}
+
+            return nbResult && stormResult && tpResult
+        }
+        assert nbResult: "Northbound should generate an error message about not being able to find a flow"
```
rtretyak: This is ok, just FYI you can put those asserts directly into Wrappers.wait and it will report the original failure reason in case of failue
```
Wrappers.wait {
//whatever
assert expression1
assert expression2, "custom message" //will still report this message in case of failure
assert expression3
//note that in this case there is no need to return something. 'wait' will digest a 'null' result as OK
}
```
in this case you can also get rid of lines 46-48

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -1,9 +1,69 @@
...
+
+
+    def "Check Floodlight logging"() {
+        when: "Retrieve floodlight logs for last 5 minutes"
+        def tagName = "kilda-floodlight"
+        def result = elasticService.getLogs(new ElasticQueryBuilder().setTags(tagName).setTimeRange(300).build())
+        assert result?.hits?.total > 0
+
+        then: "There should be discovery messages"
+        result.hits.hits.any {hit -> hit.source.message.toLowerCase().contains("push discovery package via")}
```
ylobankov: Maybe it is worth moving "push discovery package via" to some constants.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -1,9 +1,69 @@
...
+
+    def "Check Northbound, Storm and Topology Engine logging"() {
+        when: "A non-existent flow is requested"
+        def flowId = "nonexistentflowid" + System.currentTimeMillis()
+        try {
+            northbound.getFlow(flowId)
+        } catch (HttpClientErrorException e) {
+        }
+        and: "Rules on a switch are validated"
+        def switchId = topologyDefinition.switches[2].dpId
```
ylobankov: Our topology can have only 2 switches (or even 1). Probably we can use here something like this:
 def switchId = topologyDefinition.switches.first().dpId

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -1,9 +1,69 @@
...
+        northbound.validateSwitchRules(switchId)
+
+        then: "Northbound, Storm and Topology Engine should log these actions within 1 minute"
+        def nbTag = "kilda-northbound"
```
ylobankov: The same, we need to think about moving it to constants.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -1,9 +1,69 @@
...
+        def stormTag = "storm-worker_log"
+        def tpTag = "kilda-tpe"
+
+        boolean nbResult = false
+        boolean stormResult = false
+        boolean tpResult = false
+        int timeout = 60
```
ylobankov: ditto.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
ylobankov: You are in 'then' block. 'assert' is redundant. Also, consider changing ':' -> ','

npliashechnikov: Thanks, will use this in next tests

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
npliashechnikov: Thanks, forgot to fix this during refactoring

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
npliashechnikov: Agreed

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
npliashechnikov: Can we still use assert formatting without an actual assert?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
npliashechnikov: ?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
npliashechnikov: May well be.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
ylobankov: it is also related to moving some stuff to constants.

npliashechnikov: I see. It is pre-initialized variables rather than constants. Moved into setup section.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -1,9 +1,89 @@
...
 
+    @Autowired
+    ElasticService elasticService
+
+    String discoveryMessage = "push discovery package via"
+
+    def "Check Floodlight logging"() {
+        when: "Retrieve floodlight logs for last 5 minutes"
+        def result
+        if (this.profile.equalsIgnoreCase("virtual")) {
```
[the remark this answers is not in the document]
rtretyak: 1. Is this still required? I thought we have sorted out all inconsistencies on environment level
2. If this is still the case, then we should think towards implementing 2 versions of service (for virtual and for hardware profile) with the same interface but different implementations, instead of using if/else everywhere. But I more lean towards fixing this on logstash level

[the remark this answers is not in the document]
npliashechnikov: There is a point in this, but as the logging is now fixed on local env, I'd just delete the separate branch

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -1,9 +1,89 @@
...
+            result = elasticService.getLogs(new ElasticQueryBuilder().setField("type", KildaTypes.FLOODLIGHT).
+                    setTimeRange(300).build())
+        } else {
+            result = elasticService.getLogs(new ElasticQueryBuilder().setTags(KildaTags.FLOODLIGHT).
+                    setTimeRange(300).build())
+        }
+        assert result?.hits?.total > 0
```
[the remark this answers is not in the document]
npliashechnikov: Will do

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -1,9 +1,89 @@
...
+            northbound.getFlow(flowId)
+        } catch (HttpClientErrorException e) {
+        }
+
+        and: "Rules on a switch are validated"
+        def switchId = topologyDefinition.switches.first().dpId
+        northbound.validateSwitchRules(switchId)
+
+        then: "Northbound, Storm and Topology Engine should log these actions within 1 minute"
+        def result = Wrappers.wait(timeout, 10) {
```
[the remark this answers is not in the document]
npliashechnikov: But how do asserts get propagated from within closure?
Do they get unmuted after the timeout?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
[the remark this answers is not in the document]
rtretyak: Want to add my opinion here. I don't want for 'custom message' to become a must-have requirement for all our tests. The assertion should be self-descriptive. imo this assertion already shows what went wrong, but I'm not saying we shouldn't add description here, it actually doesn't matter for me unless we make it a requirement for all assertions

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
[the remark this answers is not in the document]
rtretyak: Yes they do. I tried to explain this in one of my comments too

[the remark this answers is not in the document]
ylobankov: No, they should not. If asserts are not satisfied, we will get a timeout exception with the original exception where the reason will be provided.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -1,9 +1,89 @@
...
+            result = elasticService.getLogs(new ElasticQueryBuilder().setField("type", KildaTypes.FLOODLIGHT).
+                    setTimeRange(300).build())
+        } else {
+            result = elasticService.getLogs(new ElasticQueryBuilder().setTags(KildaTags.FLOODLIGHT).
+                    setTimeRange(300).build())
+        }
+        assert result?.hits?.total > 0
+
+        then: "There should be discovery messages"
+        result.hits.hits.any {hit -> hit.source.message.toLowerCase().contains(discoveryMessage)}
```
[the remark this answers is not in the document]
ylobankov: Please, provide whitespace between '{' and 'hit'. The same for '}' at the end of the line.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -1,9 +1,89 @@
...
+    }
+
+    def "Check Northbound, Storm and Topology Engine logging"() {
+        setup:
```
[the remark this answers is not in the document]
ylobankov: This setup doesn't have any sense. Please move 

int timeout = 60

under 'when' block.

Also, pay attention that we use 'given' instead of 'setup'. There is no difference between them but let's have consistency.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -1,9 +1,67 @@
...
+                        setTimeRange(timeout * 2).setLevel("INFO").build())
+
+            assert nbLogs?.hits?.hits?.any { hit -> hit.source.message.contains(flowId) } :
+                    "Northbound should generate an error message about not being able to find a flow"
+            assert stormLogs?.hits?.hits?.any { hit -> hit.source.message.contains(flowId) } :
+                    "Storm should generate an error message about not being able to find a flow"
+            assert tpLogs?.hits?.hits?.any { hit -> hit.source.message.contains(switchId.toString()) } :
+                    "Topology Engine should generate an info message in case of a switch rules validation event"
+
+            return true
```
ylobankov: It looks like it can be removed.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -1,9 +1,65 @@
...
```
ylobankov: Why are you ignoring my previous comment? This 'return' is redundant and can be removed.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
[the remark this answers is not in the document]
npliashechnikov: done

PR #1589, opened 2018-11-14
#1571: Add Meters to unicast/broadcast verification rules

### LINKED ISSUE #1571: Add Meter to all rules that forward packets to the Controller
## Problem
In certain circumstances, there could be a huge amount of packets on a data plane that we usually send to the controller. This may cause DoS for the rest of the traffic for the switch and for Floodlight.

## Proposal
In order to protect the controller and switch CPU from overwhelming by data forwarded from data-plane, we need to put a meter for all rules that forward packets to the controller. The meter should be in packets per second and configurable per rule type.

nikitacherevko:
closes #1571

### DIFF
--- a/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
+++ b/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
@@ -249,9 +249,8 @@ class RecordHandler implements Runnable {
             meterId = allocateMeterId(
                     command.getMeterId(), command.getSwitchId(), command.getId(), command.getCookie());
 
-            context.getSwitchManager().installMeter(
-                    DatapathId.of(command.getSwitchId().toLong()),
-                    command.getBandwidth(), 1024, meterId);
+            installMeter(DatapathId.of(command.getSwitchId().toLong()), meterId, command.getBandwidth(),
+                    command.getId());
         } else {
             logger.debug("Installing unmetered ingress flow. Switch: {}, cookie: {}",
                     command.getSwitchId(), command.getCookie());
@@ -371,9 +370,8 @@ class RecordHandler implements Runnable {
             meterId = allocateMeterId(
                     command.getMeterId(), command.getSwitchId(), command.getId(), command.getCookie());
 
-            context.getSwitchManager().installMeter(
-                    DatapathId.of(command.getSwitchId().toLong()),
-                    command.getBandwidth(), 1024, meterId);
+            installMeter(DatapathId.of(command.getSwitchId().toLong()), meterId, command.getBandwidth(),
+                    command.getId());
         } else {
             logger.debug("Installing unmetered one switch flow. Switch: {}, cookie: {}",
                     command.getSwitchId(), command.getCookie());
@@ -414,18 +412,24 @@ class RecordHandler implements Runnable {
                 logger.warn("No rules were removed by criteria {} for flow {} from switch {}",
                         criteria, command.getId(), dpid);
             }
-
-            // FIXME(surabujin): QUICK FIX - try to drop meterPool completely
-            Long meterId = command.getMeterId();
-            if (meterId != null) {
-                switchManager.deleteMeter(dpid, meterId);
-            }
-
-            message.setDestination(replyDestination);
-            getKafkaProducer().sendMessageAndTrack(replyToTopic, message);
         } catch (SwitchOperationException e) {
             throw new FlowCommandException(command.getId(), ErrorType.DELETION_FAILURE, e);
         }
+
+        // FIXME(surabujin): QUICK FIX - try to drop meterPool completely
+        Long meterId = command.getMeterId();
+        if (meterId != null) {
+            try {
+                switchManager.deleteMeter(dpid, meterId);
+            } catch (UnsupportedOperationException e) {
+                logger.info("Skip meter {} deletion from switch {}: {}", meterId, dpid, e.getMessage());
+            } catch (SwitchOperationException e) {
+                logger.error("Failed to delete meter {} from switch {}: {}", meterId, dpid, e.getMessage());
+            }
+        }
+
+        message.setDestination(replyDestination);
+        getKafkaProducer().sendMessageAndTrack(replyToTopic, message);
     }
 
     /**
@@ -690,14 +694,18 @@ class RecordHandler implements Runnable {
 
         try {
             DatapathId dpid = DatapathId.of(request.getSwitchId().toLong());
-            long txId = context.getSwitchManager().deleteMeter(dpid, request.getMeterId());
+            context.getSwitchManager().deleteMeter(dpid, request.getMeterId());
 
-            DeleteMeterResponse response = new DeleteMeterResponse(txId != 0L);
+            boolean deleted = context.getSwitchManager().dumpMeters(dpid)
+                    .stream()
+                    .anyMatch(config -> config.getMeterId() == request.getMeterId());
+            DeleteMeterResponse response = new DeleteMeterResponse(deleted);
             InfoMessage infoMessage = new InfoMessage(response, System.currentTimeMillis(), message.getCorrelationId(),
                     replyDestination);
             producerService.sendMessageAndTrack(replyToTopic, infoMessage);
         } catch (SwitchOperationException e) {
-            logger.info("Meter deletion is unsuccessful. Switch {} not found", request.getSwitchId());
+            logger.error("Failed to delete meter {} from switch {}: {}", request.getMeterId(), request.getSwitchId(),
+                    e.getMessage());
             ErrorData errorData =
                     new ErrorData(ErrorType.DATA_INVALID, e.getMessage(), request.getSwitchId().toString());
             ErrorMessage error = new ErrorMessage(errorData,
@@ -820,6 +828,19 @@ class RecordHandler implements Runnable {
             allocatedId = meterPool.allocate(switchId, flowId, Math.toIntExact(meterId));
         }
         return allocatedId;
+    }
+
+    private void installMeter(DatapathId dpid, long meterId, long bandwidth, String flowId) {
+        try {
+            context.getSwitchManager().installMeter(dpid, bandwidth, 1024, meterId);
+        } catch (UnsupportedOperationException e) {
+            logger.info("Skip meter {} installation for flow {} on switch {}: {}",
+                    meterId, flowId, dpid, e.getMessage());
+        } catch (SwitchOperationException e) {
+            logger.error("Failed to install meter {} for flow {} on switch {}: {}", meterId, flowId, dpid,
+                    e.getMessage());
+        }
+
     }
 
     private void parseRecord(ConsumerRecord<String, String> record) {
--- a/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
+++ b/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
@@ -15,6 +15,7 @@
 
 package org.openkilda.floodlight.switchmanager;
 
+import static java.lang.String.format;
 import static java.util.Arrays.asList;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
@@ -43,6 +44,7 @@ import org.openkilda.messaging.error.ErrorMessage;
 import org.openkilda.messaging.error.ErrorType;
 import org.openkilda.messaging.payload.flow.OutputVlanType;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
@@ -58,6 +60,9 @@ import net.floodlightcontroller.core.module.IFloodlightModule;
 import net.floodlightcontroller.core.module.IFloodlightService;
 import net.floodlightcontroller.restserver.IRestApiService;
 import net.floodlightcontroller.util.FlowModUtils;
+import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.math.NumberUtils;
 import org.projectfloodlight.openflow.protocol.OFBarrierReply;
 import org.projectfloodlight.openflow.protocol.OFBarrierRequest;
 import org.projectfloodlight.openflow.protocol.OFErrorMsg;
@@ -67,10 +72,6 @@ import org.projectfloodlight.openflow.protocol.OFFlowMod;
 import org.projectfloodlight.openflow.protocol.OFFlowStatsEntry;
 import org.projectfloodlight.openflow.protocol.OFFlowStatsReply;
 import org.projectfloodlight.openflow.protocol.OFFlowStatsRequest;
-import org.projectfloodlight.openflow.protocol.OFLegacyMeterBandDrop;
-import org.projectfloodlight.openflow.protocol.OFLegacyMeterFlags;
-import org.projectfloodlight.openflow.protocol.OFLegacyMeterMod;
-import org.projectfloodlight.openflow.protocol.OFLegacyMeterModCommand;
 import org.projectfloodlight.openflow.protocol.OFMessage;
 import org.projectfloodlight.openflow.protocol.OFMeterConfig;
 import org.projectfloodlight.openflow.protocol.OFMeterConfigStatsReply;
@@ -146,6 +147,7 @@ public class SwitchManager implements IFloodlightModule, IFloodlightService, ISw
     private SwitchTrackingService switchTracking;
 
     private ConnectModeRequest.Mode connectMode;
+    private SwitchManagerConfig config;
 
     /**
      * Create an OFInstructionApplyActions which applies actions.
@@ -158,19 +160,6 @@ public class SwitchManager implements IFloodlightModule, IFloodlightService, ISw
             List<OFAction> actionList) {
         return ofFactory.instructions().applyActions(actionList).createBuilder().build();
     }
-
-    /**
-     * Returns legacy meter action.
-     *
-     * @param ofFactory OF factory for the switch
-     * @param meterId meter id
-     * @return {@link OFAction}
-     */
-    private static OFAction legacyMeterAction(final OFFactory ofFactory, final long meterId) {
-        return ofFactory.actions().buildNiciraLegacyMeter().setMeterId(meterId).build();
-    }
-
-    // IFloodlightModule Methods
 
     /**
      * {@inheritDoc}
@@ -216,7 +205,9 @@ public class SwitchManager implements IFloodlightModule, IFloodlightService, ISw
         switchTracking = context.getServiceImpl(SwitchTrackingService.class);
 
         ConfigurationProvider provider = ConfigurationProvider.of(context, this);
-        String connectModeProperty = provider.getConfiguration(SwitchManagerConfig.class).getConnectMode();
+        config = provider.getConfiguration(SwitchManagerConfig.class);
+        String connectModeProperty = config.getConnectMode();
+
         try {
             connectMode = ConnectModeRequest.Mode.valueOf(connectModeProperty);
         } catch (Exception e) {
@@ -505,15 +496,12 @@ public class SwitchManager implements IFloodlightModule, IFloodlightService, ISw
         List<OFMeterConfig> result = null;
         IOFSwitch sw = lookupSwitch(dpid);
         if (sw == null) {
-            throw new IllegalArgumentException(String.format("Switch %s was not found", dpid));
+            throw new IllegalArgumentException(format("Switch %s was not found", dpid));
         }
+
+        verifySwitchSupportsMeters(sw);
 
         OFFactory ofFactory = sw.getOFFactory();
-        if (ofFactory.getVersion().compareTo(OF_13) < 0) {
-            throw new UnsupportedSwitchOperationException(dpid,
-                    String.format("Dumping of meters is not supported on the requested switch %s.", dpid));
-        }
-
         OFMeterConfigStatsRequest meterRequest = ofFactory.buildMeterConfigStatsRequest()
                 .setMeterId(0xffffffff)
                 .build();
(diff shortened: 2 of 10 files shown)


services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -778,6 +775,27 @@ public void installDropFlow(final DatapathId dpid) throws SwitchOperationExcepti
         }
     }
 
+    private boolean isSwitchSupportsMeters(IOFSwitch sw) {
+        if (OVS_MANUFACTURER.equals(sw.getSwitchDescription().getManufacturerDescription())) {
+            logger.info("Meters are not supported on OVS switch {}", sw.getId());
```
surabujin: Producing log message here will increate mess in our logs. It must be exception... and caller must decide is it going to log it... to ignore it... or handle in some other way.

Imagine several call to isSwitchSupportsMeters during one operation/command.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -807,6 +825,40 @@ private long buildAndinstallMeter(final IOFSwitch sw, final DatapathId dpid, fin
         return pushFlow(sw, "--InstallMeter--", meterMod);
     }
 
+    @VisibleForTesting
+    long installMeterForDefaultRule(IOFSwitch sw, long cookie, boolean isBroadcast) throws OfInstallException {
+        OFFactory ofFactory = sw.getOFFactory();
+
+        if (isSwitchSupportsMeters(sw) && isSwitchSupportsPktpsFlag(sw)) {
```
surabujin: Lack of `isSwitchSupportsPktpsFlag()` support is not enough reason to not install meter on rule that direct packets into controller.

We can recalculate meter limit on kbps ... lets take packet size as 500 bytes so 1000pkts is (500 * 1000 / 1024 kbps)... so do not use pkts for switches that do not support it (we should dig further why they don't support it, according to spec they should support it).

(critical)

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -807,6 +825,40 @@ private long buildAndinstallMeter(final IOFSwitch sw, final DatapathId dpid, fin
         return pushFlow(sw, "--InstallMeter--", meterMod);
     }
 
+    @VisibleForTesting
+    long installMeterForDefaultRule(IOFSwitch sw, long cookie, boolean isBroadcast) throws OfInstallException {
```
surabujin: 1. We can reuse `.buildAndinstallMeter(...)`... it can be altered to accept flags and called from here.

2. `boolean isBroadcast`... this is extremely ugly way to define what we should do... Please do not spread this approach further. You can pass bandwidth value from `installVerificationRule` it is already have this ugly argument/condition... so let it be the only place we use it.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -713,23 +706,27 @@ public void installVerificationRule(final DatapathId dpid, final boolean isBroad
...
+        logger.debug("Installing verification rule for {}", dpid);
 
         Match match = matchVerification(sw, isBroadcast);
-        ArrayList<OFAction> actionList = new ArrayList<>(2);
+        ArrayList<OFAction> actionList = new ArrayList<>(3);
         actionList.add(actionSendToController(sw));
         actionList.add(actionSetDstMac(sw, dpidToMac(sw)));
+
+        long cookie = isBroadcast ? VERIFICATION_BROADCAST_RULE_COOKIE : VERIFICATION_UNICAST_RULE_COOKIE;
+        long meterId = installMeterForDefaultRule(sw, cookie, isBroadcast);
```
surabujin: It will produce error `meter already exist` on each switch reconnect.

(critical)

services/src/floodlight-modules/src/test/java/org/openkilda/floodlight/switchmanager/SwitchManagerTest.java
```diff
@@ -379,6 +387,53 @@ public void shouldDeleteDefaultRules() throws Exception {
...
+
+        mockFlowStatsRequest(cookie, DROP_RULE_COOKIE, VERIFICATION_BROADCAST_RULE_COOKIE,
+                VERIFICATION_UNICAST_RULE_COOKIE);
+
+        Capture<OFFlowMod> capture = EasyMock.newCapture(CaptureType.ALL);
+        expect(iofSwitch.write(capture(capture))).andReturn(true).times(5);
+
+        mockBarrierRequest();
+        mockFlowStatsRequest(cookie);
+        expectLastCall();
```
surabujin: O_O weird usage.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -534,82 +525,73 @@ public long installOneSwitchFlow(
...
-            return meterCommandXid;
-        }
+        OFMeterConfigStatsRequest meterRequest = ofFactory.buildMeterConfigStatsRequest()
+                .setMeterId(meterId)
+                .build();
 
-        IOFSwitch sw = lookupSwitch(dpid);
+        try {
+            ListenableFuture<OFMeterConfigStatsReply> future = sw.writeRequest(meterRequest);
+            OFMeterConfigStatsReply reply = future.get(1, TimeUnit.SECONDS);
```
surabujin: 1 second could be not enough for heavily loaded switch... with a lot of pkt_out/pkt_in commands. And we should make it using some constant or config value (I would like to see it as a class level constant).

Later we should replace this "sleep" calls with tools proposed in https://github.com/telstra/open-kilda/pull/1601.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -534,82 +525,73 @@ public long installOneSwitchFlow(
...
+            verifySwitchSupportsMeters(sw);
+
+            Set<OFMeterFlags> flags = ImmutableSet.of(OFMeterFlags.KBPS, OFMeterFlags.BURST);
+            buildAndInstallMeter(sw, flags, bandwidth, burstSize, meterId);
+            // All cases when we're installing meters require that we wait until the command is processed and
+            // the meter is installed.
+            sendBarrierRequest(sw);
+        } else {
+            throw new InvalidMeterIdException(dpid,
+                    format("Could not delete meter '%d' from switch '%s'. Meter id must be positive.",
```
surabujin: why "delete"? This is installMeter method.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -712,23 +707,38 @@ public void installVerificationRule(final DatapathId dpid, final boolean isBroad
...
 
-        Match match = matchVerification(sw, isBroadcast);
-        ArrayList<OFAction> actionList = new ArrayList<>(2);
+        long cookie = isBroadcast ? VERIFICATION_BROADCAST_RULE_COOKIE : VERIFICATION_UNICAST_RULE_COOKIE;
+        long meterId = cookie & PACKET_IN_RULES_METERS_MASK;
+        long meterRate = isBroadcast ? config.getBroadcastRateLimit() : config.getUnicastRateLimit();
+        if (!meterExists(sw, meterId, meterRate)) {
+            installMeterForDefaultRule(sw, meterId, meterRate);
+        } else {
+            logger.debug("Meter {} won't be installed on {} switch. It already exists", meterId, dpid);
```
surabujin: "reinstalled" instead of "installed" is more suitable here. And it must be at least info, this info can be useful in operation's investigations.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -712,23 +707,38 @@ public void installVerificationRule(final DatapathId dpid, final boolean isBroad
...
+        }
+
+        ArrayList<OFAction> actionList = new ArrayList<>(3);
         actionList.add(actionSendToController(sw));
         actionList.add(actionSetDstMac(sw, dpidToMac(sw)));
+
+        OFInstructionMeter meter = null;
+        //FIXME: we should install meters on all switches with OF > 1.2
+        if (isSwitchSupportsPktpsFlag(sw)) {
```
surabujin: Why we have this condition here?
We should catch exception `UnsupportedSwitchOperationException` from `installMeterForDefaultRule` and make appropriate decisions here and not reinvent internal logic of `installMeterForDefaultRule`.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -822,31 +849,32 @@ private long buildAndinstallMeter(final IOFSwitch sw, final DatapathId dpid, fin
...
-                .setFlags(flags)
-                .build();
-
-        return pushFlow(sw, "--InstallMeter", meterMod);
+        try {
+            verifySwitchSupportsMeters(sw);
+            if (isSwitchSupportsPktpsFlag(sw)) {
+                flags = ImmutableSet.of(OFMeterFlags.PKTPS, OFMeterFlags.STATS);
+            } else {
+                // FIXME: should be implemented the way to install meters with KBPS flag
```
surabujin: Is there some issues with implementation this meters via "kbps" meter?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -796,12 +806,29 @@ void installDropLoopRule(DatapathId dpid) throws SwitchOperationException {
...
+
+    // FIXME: centec switches can't recognize PKTPS flag for meters.
+    // Need to simplify detection if the switch don't support PKTPS flag.
+    private boolean isSwitchSupportsPktpsFlag(IOFSwitch sw) {
+        return !StringUtils.contains(sw.getSwitchDescription().getManufacturerDescription(), "Centec");
+    }
+
+    private void buildAndInstallMeter(IOFSwitch sw, Set<OFMeterFlags> flags, long bandwidth, long burstSize,
+                                      long meterId) throws OfInstallException {
+        logger.debug("Installing meter {} on switch {} width bandwidth {}", meterId, sw.getId(), bandwidth);
```
surabujin: info?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -1593,4 +1606,36 @@ private MacAddress getPortHwAddress(IOFSwitch sw, int portNumber) throws SwitchO
...
+            verifySwitchSupportsMeters(sw);
+            OFMeterConfig meterConfig = getMeter(sw.getId(), meter);
+            long expectedRate = isSwitchSupportsPktpsFlag(sw) ? rate : getRateForMeterInKbps(rate);
+
+            List<OFMeterBand> ofMeterBands = meterConfig.getEntries();
+            if (ofMeterBands == null || ofMeterBands.isEmpty()) {
+                return false;
+            }
+
+            OFMeterBand meterBand = ofMeterBands.get(0);
```
surabujin: Danger assumption... The method name tells about check for meter existence, but method itself not only check the existence, but also compare meter options. It will lead to problems - this method report "meter does not exist" and caller call setup meter with appropriate options, but get error from switch "meter ID already exists". Confusing?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -1593,4 +1606,36 @@ private MacAddress getPortHwAddress(IOFSwitch sw, int portNumber) throws SwitchO
...
+            }
+        } catch (SwitchOperationException e) {
+            logger.warn(format("Failed to verify the existence of meter %s", meter), e);
+        }
+
+        return false;
+    }
+
+    private long getRateForMeterInKbps(long ratePkts) {
+        return (ratePkts * config.getDiscoPacketSize()) / 1024L;
```
surabujin: Why it is into config? The best way is to "ask" it from `PathVerificationModule`, but it will(probably) lead to circular modules dependency.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
```diff
@@ -690,14 +690,17 @@ private void doDeleteMeter(CommandMessage message, String replyToTopic, Destinat
...
+            boolean deleted = context.getSwitchManager().dumpMeters(dpid)
+                    .stream()
+                    .anyMatch(config -> config.getMeterId() == request.getMeterId());
+            DeleteMeterResponse response = new DeleteMeterResponse(deleted);
             InfoMessage infoMessage = new InfoMessage(response, System.currentTimeMillis(), message.getCorrelationId(),
                     replyDestination);
             producerService.sendMessageAndTrack(replyToTopic, infoMessage);
         } catch (SwitchOperationException e) {
-            logger.info("Meter deletion is unsuccessful. Switch {} not found", request.getSwitchId());
+            logger.info(String.format("Failed to delete the meter from switch %s", request.getSwitchId()), e);
```
surabujin: I don't think we need traceback here. `SwitchOperationException` is our exception and we know when and why it happens.

PS Info level log record with traceback is weird (as for me).

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -538,78 +525,47 @@ public long installOneSwitchFlow(
...
+            verifySwitchSupportsMeters(sw);
+
+            Set<OFMeterFlags> flags = ImmutableSet.of(OFMeterFlags.KBPS, OFMeterFlags.BURST);
+            buildAndInstallMeter(sw, flags, bandwidth, burstSize, meterId);
+            // All cases when we're installing meters require that we wait until the command is processed and
+            // the meter is installed.
+            sendBarrierRequest(sw);
+        } else {
+            throw new InvalidMeterIdException(dpid,
+                    format("Could not install meter '%d' from switch '%s'. Meter id must be positive.",
```
surabujin: Maybe 'onto' or 'into' or even 'on'?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -696,8 +652,21 @@ public long deleteMeter(final DatapathId dpid, final long meterId)
...
+        List<Long> deletedRules = deleteRulesWithCookie(dpid, DROP_RULE_COOKIE, VERIFICATION_BROADCAST_RULE_COOKIE,
                 VERIFICATION_UNICAST_RULE_COOKIE, DROP_VERIFICATION_LOOP_RULE_COOKIE);
+
+        try {
+            verifySwitchSupportsMeters(sw);
+
+            deleteMeter(dpid, VERIFICATION_BROADCAST_RULE_COOKIE & PACKET_IN_RULES_METERS_MASK);
+            deleteMeter(dpid, VERIFICATION_UNICAST_RULE_COOKIE & PACKET_IN_RULES_METERS_MASK);
+        } catch (UnsupportedSwitchOperationException e) {
+            logger.warn("Failed to delete default rules' meters", e);
```
surabujin: This is not valid... If switch don't support meters... on switch activation we will add meters... so we should not try to remove them on switch deactivation (or on manual request in this case)... So if switch don't support meters we should log "skip meter removal do to lack of meter support by switch dpID" on info level.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
```diff
@@ -249,9 +249,13 @@ private void installIngressFlow(final InstallIngressFlow command) throws SwitchO
...
                     command.getMeterId(), command.getSwitchId(), command.getId(), command.getCookie());
 
-            context.getSwitchManager().installMeter(
-                    DatapathId.of(command.getSwitchId().toLong()),
-                    command.getBandwidth(), 1024, meterId);
+            try {
+                context.getSwitchManager().installMeter(DatapathId.of(command.getSwitchId().toLong()),
+                        command.getBandwidth(), 1024, meterId);
+            } catch (SwitchOperationException e) {
+                logger.info(String.format("Failed to install meter %s for flow %s on switch %s",
```
surabujin: We do not need traceback here.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -712,23 +681,27 @@ public void installVerificationRule(final DatapathId dpid, final boolean isBroad
...
-
-        Match match = matchVerification(sw, isBroadcast);
-        ArrayList<OFAction> actionList = new ArrayList<>(2);
+        logger.debug("Installing verification rule for {}", dpid);
+        ArrayList<OFAction> actionList = new ArrayList<>(3);
         actionList.add(actionSendToController(sw));
         actionList.add(actionSetDstMac(sw, dpidToMac(sw)));
+
+        long cookie = isBroadcast ? VERIFICATION_BROADCAST_RULE_COOKIE : VERIFICATION_UNICAST_RULE_COOKIE;
+        long meterId = cookie & PACKET_IN_RULES_METERS_MASK;
```
surabujin: Add comment near definition of `PACKET_IN_RULES_METERS_MASK` that it is used to determine meter ID for default/system rules (with example).

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -796,14 +769,32 @@ void installDropLoopRule(DatapathId dpid) throws SwitchOperationException {
...
+    }
+
+    private void buildAndInstallMeter(IOFSwitch sw, Set<OFMeterFlags> flags, long bandwidth, long burstSize,
+                                      long meterId) throws OfInstallException {
+        logger.info("Installing meter {} on switch {} width bandwidth {}", meterId, sw.getId(), bandwidth);
 
-        Set<OFMeterFlags> flags = new HashSet<>(asList(OFMeterFlags.KBPS, OFMeterFlags.BURST));
         OFFactory ofFactory = sw.getOFFactory();
 
+        // NB: some switches can't accept zero value for burst size
```
surabujin: ... it is confusing. "Some switches replace 0 burst size value with some predefined value" should be less confusing..

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -822,31 +813,23 @@ private long buildAndinstallMeter(final IOFSwitch sw, final DatapathId dpid, fin
...
-                .setMeterId(meterId)
-                .setCommand(OFLegacyMeterModCommand.ADD)
-                .setMeters(singletonList(bandBuilder.build()))
-                .setFlags(flags)
-                .build();
+        Set<OFMeterFlags> flags;
+        if (isSwitchSupportsPktpsFlag(sw)) {
+            flags = ImmutableSet.of(OFMeterFlags.PKTPS, OFMeterFlags.STATS);
+        } else {
+            flags = ImmutableSet.of(OFMeterFlags.KBPS, OFMeterFlags.STATS);
```
surabujin: Should we recalculate rate here?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -1290,6 +1258,45 @@ public IOFSwitch lookupSwitch(DatapathId dpId) throws SwitchNotFoundException {
...
+    @VisibleForTesting
+    OFInstructionMeter installMeterAndBuildInstruction(IOFSwitch sw, long meterId, long packetsLimit,
+                                                       ArrayList<OFAction> actionList) {
+        long rate = isSwitchSupportsPktpsFlag(sw) ? packetsLimit : getRateForMeterInKbps(packetsLimit);
+
+        OFMeterConfig meterConfig;
+        try {
+            meterConfig = getMeter(sw.getId(), meterId);
+        } catch (SwitchOperationException e) {
+            logger.warn(format("Meter %s won't be installed on the switch %s", meterId, sw.getId()), e);
```
surabujin: Use message provided by exception object. Also we don't need traceback here too.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -1290,6 +1258,45 @@ public IOFSwitch lookupSwitch(DatapathId dpId) throws SwitchNotFoundException {
                 .collect(Collectors.toList());
     }
 
+    @VisibleForTesting
+    OFInstructionMeter installMeterAndBuildInstruction(IOFSwitch sw, long meterId, long packetsLimit,
```
surabujin: You should mention "default" rules in method name. And `AndBuildInstruction` is extra/useless info, because or return value type.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -1290,6 +1258,45 @@ public IOFSwitch lookupSwitch(DatapathId dpId) throws SwitchNotFoundException {
...
+                logger.info("Meter {} will be reinstalled on {} switch. The rate has been changed from {} to {}",
+                        meterId, sw.getId(), meterBandDrop.getRate(), rate);
+
+                buildAndDeleteMeter(sw, sw.getId(), meterId);
+                installMeterForDefaultRule(sw, meterId, rate);
+            } else {
+                logger.info("Meter {} won't be reinstalled on {} switch. It already exists", meterId, sw.getId());
+            }
+        } catch (SwitchOperationException e) {
+            logger.warn(String.format("Failed to (re)install the meter %s on switch %s", meterId, sw.getId()), e);
```
surabujin: traceback is excess here

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -1290,6 +1245,55 @@ public IOFSwitch lookupSwitch(DatapathId dpId) throws SwitchNotFoundException {
...
+                flags = ImmutableSet.of(OFMeterFlags.KBPS, OFMeterFlags.STATS);
+                rate = (ratePkts * config.getDiscoPacketSize()) / 1024L;
+            }
+
+            if (meterBandDrop != null && meterBandDrop.getRate() == rate) {
+                logger.info("Meter {} won't be reinstalled on switch {}. It already exists", meterId, sw.getId());
+                return buildMeterInstruction(meterId, sw, sw.getOFFactory(), actionList);
+            }
+
+            if (meterBandDrop != null && meterBandDrop.getRate() != rate) {
```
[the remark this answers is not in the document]
nikitacherevko: agree, my bad

PR #1598, opened 2018-11-15
 Add tests to verify deleting switch rules by inPort, inVlan, outPort

ylobankov:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
```diff
@@ -156,6 +160,68 @@ class SwitchRulesSpec extends BaseSpecification {
...
+
+        where:
+        flow << [buildFlow()] * 4
+        data << [[description  : "inPort",
+                  switch       : srcSwitch,
+                  inPort       : flow.source.portNumber,
+                  inVlan       : null,
+                  outPort      : null,
+                  rulesDeleted : 1,
+                  rulesRemained: defaultRules.size() + flowRulesCount - 1
```
rtretyak: rulesDeleted and rulesRemained don't seem to be changing between iterations. It seems that you can add negative cases here by feeding non-existing port-vlans and changing these params, can't you?

services/src/test-library/src/main/java/org/openkilda/testing/service/northbound/NorthboundService.java
```diff
@@ -76,7 +76,8 @@
 
     List<Long> installSwitchRules(SwitchId switchId, InstallRulesAction installAction);
 
-    List<Long> deleteSwitchRules(SwitchId switchId, DeleteRulesAction deleteAction);
+    List<Long> deleteSwitchRules(SwitchId switchId, DeleteRulesAction deleteAction, Integer inPort, Integer inVlan,
+                                 Integer outPort);
```
rtretyak: Is there ability to delete by outVlan?

ylobankov: It looks like not. Before creating this patch I checked the swagger documentation and found nothing about this.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
rtretyak: No I don't insist, but I find the less code the better. Otherwise, if you are not planning to extend this, then you can probably remove params that are static

ylobankov: I agree, less code is better. But positive tests are positive. Negative tests are negative. Yep, I will delete static values.

PR #1601, opened 2018-11-16
Implement tools that allow to catch errors returned via OF

### LINKED ISSUE #1122: Check actual result of OFMessages sent to the switch
Due to ansynchronous nature of OpenFlow protocol we had to wait "some time" after sending some OFMessage to the switch for possible error response (switch do not send response of flowMod messages, only if there is an error it will send error response for specific messages).

The idea:
 - prepare all message that we are goint to send to the switch
 - collect all `xid` of this messages
 - send all prepared messages
 - send barrier message
 - wait for barrier response
If any of sent message lead to error - it will received before barrier reply message. i.e. if we have received barrier reply and there is no error message before it - all messages have been processed by switch successfully.

If there was error messages we can detect with message cause this error by "xid" fields.

This "write technique" was implemented and testes as part of flow ping feature. But due to too big impact on speed/performance it's usage in flow pings was banned.

surabujin:
Add new service named SessionService. It is responsible for "creation" and
tracking communication "session" with connected switches. The session itself is
a closable object that can/should be used into try-with-resources construction.
All messages send via session will be traced for possible error responses.
Session will send barrier request when it closed.

So when we receive barrier response we can treat all sent messages as
successfully processed (if we do not receive error response for them before).

Using this session we can catch and handle error responses have been sent by
switch on our OF requests.

Close #1122

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/1601)


services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/service/session/Session.java
```diff
@@ -0,0 +1,191 @@
...
+        } catch (SwitchWriteException e) {
+            SessionCloseException closeError = new SessionCloseException(sw.getId());
+            incompleteRequestsStream()
+                    .forEach(entry -> entry.completeExceptionally(closeError));
+            throw e;
+        }
+    }
+
+    void disconnect() {
+        // must be safe to be called multiple times
```
Settis: Is it make sense to avoid concurrency here?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/service/session/Session.java
```diff
@@ -0,0 +1,191 @@
...
+        completed = true;
+
+        SessionConnectionLostException e = new SessionConnectionLostException(sw.getId());
+        incompleteRequestsStream()
+                .forEach(entry -> entry.completeExceptionally(e));
+    }
+
+    boolean handleResponse(OFMessage message) {
+        CompletableFuture<Optional<OFMessage>> future;
+        synchronized (this) {
```
sergii-iakovenko: Should requestsByXid be ConcurrentHashMap instead?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/error/SwitchWriteException.java
```diff
@@ -18,14 +18,17 @@
 import org.projectfloodlight.openflow.protocol.OFMessage;
 import org.projectfloodlight.openflow.types.DatapathId;
 
-public class OfWriteException extends AbstractException {
-    public OfWriteException(String s) {
-        super(s);
-    }
+public class SwitchWriteException extends SwitchOperationException {
+    private final transient OFMessage ofMessage;
```
sergii-iakovenko: What's the purpose of ofMessage (as exception details) if we can't be sure that we get them (after serialization/deserial)?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/service/session/Session.java
Settis: Thanks for explanation.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/error/SwitchWriteException.java
surabujin: Because we do not expose FL exceptions anywhere, they are not subject to serialize/deserialize process. I don't see the reason to invent view for `OFMessage` object just because the exception object must be capable to be serializable.

PS It is handy to have access to object that case error response.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/service/session/Session.java
surabujin: From my point of view - it is overkill. By why not - change requestsByXid on ConcurrentHashMap.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/service/session/SessionService.java
```diff
@@ -0,0 +1,120 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class SessionService implements IService, IInputTranslator {
+    private static final Logger log = LoggerFactory.getLogger(SessionService.class);
+
+    private final Map<DatapathId, SwitchSessions> sessionsByDatapath = new HashMap<>();
```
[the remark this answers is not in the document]
surabujin: done

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/service/session/SwitchSessions.java
```diff
@@ -0,0 +1,67 @@
...
+package org.openkilda.floodlight.service.session;
+
+import net.floodlightcontroller.core.IOFSwitch;
+import org.projectfloodlight.openflow.protocol.OFMessage;
+
+import java.util.HashMap;
+import java.util.Map;
+
+class SwitchSessions {
+    private final Map<Long, Session> sessionsByXid = new HashMap<>();
```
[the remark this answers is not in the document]
surabujin: Better leave it with HashMap. FL design let us to leave some methods without synchronization (always called from single/master thread). Plus in case of `ConcurrentHashMap` will make double lock/synchronization in method `unbindSession`.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/service/session/Session.java
```diff
@@ -0,0 +1,183 @@
...
+        CompletableFuture<Optional<OFMessage>> future = new CompletableFuture<>();
+
+        long xid = message.getXid();
+        requestsByXid.put(xid, future);
+        group.bindRequest(this, xid);
+
+        return future;
+    }
+
+    private void actualWrite(CompletableFuture<Optional<OFMessage>> future, OFMessage message)
```
sergii-iakovenko: Actually this method doesn't need to know about the CompletableFuture instance as the logic which calls actualWrite can handle it. Moreover, the CompletableFuture doesn't require to be completed when as an exception is thrown.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/service/session/SessionService.java
```diff
@@ -0,0 +1,122 @@
...
+     *
+     * <p>Called from main FL thread, that why we can avoid `synchronization` here.
+     */
+    void switchActivate(DatapathId dpId) {
+        SwitchSessions group = new SwitchSessions();
+        SwitchSessions previous;
+        previous = sessionsByDatapath.put(dpId, group);
+
+        if (previous != null) {
+            log.error("Switch {} already registered (connect/disconnect race condition?)", dpId);
```
sergii-iakovenko: Should "previous.disconnect();" be called to complete all bound CompletableFutures?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/service/session/Session.java
```diff
@@ -0,0 +1,183 @@
...
+        OFBarrierRequest barrier = sw.getOFFactory().barrierRequest();
+        closingBarrier = prepareRequest(barrier);
+        try {
+            actualWrite(closingBarrier, barrier);
+        } catch (SwitchWriteException e) {
+            SessionCloseException closeError = new SessionCloseException(sw.getId());
+            incompleteRequestsStream()
+                    .forEach(entry -> entry.completeExceptionally(closeError));
+            throw e;
+        }
```
sergii-iakovenko: Shouldn't we validate here that the session is empty / all expected responses received?

sergii-iakovenko:
In general looks good. Left minor comments.

PR #1616, opened 2018-11-20
Slight refactor of logs in NB

nikitacherevko:
- changed log level for all important actions like create/update/delete flow or operations with switch's rules
- changed log level for the message that tells new generated correlationId (from warning to debug)
- removed unused BasicService

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/1616)


services/src/northbound/src/main/java/org/openkilda/northbound/utils/RequestCorrelationFilter.java
```diff
@@ -50,7 +50,7 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
         String correlationId = request.getHeader(CORRELATION_ID);
         if (StringUtils.isBlank(correlationId)) {
             correlationId = UUID.randomUUID().toString();
-            LOGGER.warn("CorrelationId was not sent, generated one: {}", correlationId);
+            LOGGER.debug("CorrelationId was not sent, generated one: {}", correlationId);
```
nikitamarchenko: why do you lower debug level?

nikitacherevko: I'm not sure that it is really warning and we need such kind of information. Currently there are too many such messages that make no sense and we might missed really some warning messages.

PR #1621, opened 2018-11-20
Add flow crud test with negative (forbidden) scenarios

### LINKED ISSUE #1614: Create test that checks that single switch same port-vlan flow is forbidden
Verify that user is not able to create a single-switch flow with the src port and vlan both matching the dst port and vlan.
Test should be part of FlowCrudSpec.

rtretyak:
Also added groovy extension with convenience methods
Closes #1614

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowCrudSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowCrudSpec.groovy
@@ -7,6 +7,8 @@ import org.openkilda.functionaltests.extension.fixture.rule.CleanupSwitches
 import org.openkilda.functionaltests.helpers.FlowHelper
 import org.openkilda.functionaltests.helpers.PathHelper
 import org.openkilda.functionaltests.helpers.Wrappers
+import org.openkilda.messaging.error.MessageError
+import org.openkilda.messaging.model.SwitchId
 import org.openkilda.messaging.payload.flow.FlowPayload
 import org.openkilda.messaging.payload.flow.FlowState
 import org.openkilda.testing.model.topology.TopologyDefinition
@@ -20,6 +22,8 @@ import org.openkilda.testing.tools.FlowTrafficExamBuilder
 import groovy.transform.Memoized
 import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.http.HttpStatus
+import org.springframework.web.client.HttpClientErrorException
 import spock.lang.Narrative
 import spock.lang.Shared
 import spock.lang.Unroll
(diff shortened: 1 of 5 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowCrudSpec.groovy
```diff
@@ -151,6 +155,95 @@ class FlowCrudSpec extends BaseSpecification {
...
+        flow.destination.vlanId = flow.source.vlanId
+
+        when: "Try creating such flow"
+        northbound.addFlow(flow)
+
+        then: "Error is returned, stating a readable reason"
+        def error = thrown(HttpClientErrorException)
+        error.statusCode == HttpStatus.BAD_REQUEST
+        error.responseBodyAsString.to(MessageError).errorMessage ==
+                "Could not create flow: It is not allowed to create one-switch flow for the same ports and vlans"
```
ylobankov: Is there sense to move such kind of messages to special error message constants? In the future, we may have lots of such kind of tests.

rtretyak: I understand what you mean, but my thought process was that it is very unlikely that we will ever use this text in any other test and I didn't want to spread-out parts of the test to different places (away from the test/spec itself). We can move this and others to some Errors class that will contain all errors and I'm not too hell-bent to keeping current implementation. Do you want to change this?

ylobankov: Yes, you are right that it is very unlikely that we will ever use this text in any other test again. But when such messages are located in some single place, it is very simple to make the corresponding changes if response messages are changed. Also, it is easy to understand what test cases are covered, what not, when we have all messages at a single place.

rtretyak: 1. Since we agreed that this will never be reused, then when response changes in both cases we will have to change only in 1 place, regardless.
2. There is no global Errors class with constants at the dev code to compare with our class, so I don't really understand how can we asses coverage by looking at Errors class. We do have specs and features that represent our test scenarios and coverage. Can you elaborate on how having errors in one place helps with understanding coverage? Do you mean that it will be like a short summary of what negative tests we have? This may be misleading / not representative

fyi, I already agree to move this to a separate class, just debating.
Can I/we do this in a separate patch? I guess this will have to do with refactoring and reviewing all existing tests to make it consistent. I can make an issue for it

ylobankov: Yes, I meant a short summary of what negative tests we have. Ok, let's do it in a separate patch. It will be better.

rtretyak: created #1622

PR #1627, opened 2018-11-21
Fix functional test for verifying switch rules deletion

ylobankov:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
```diff
@@ -38,13 +38,16 @@ class SwitchRulesSpec extends BaseSpecification {
...
+    @Shared
+    List dstSwDefaultRules
     @Shared
     int flowRulesCount = 2
 
     def setup() {
         (srcSwitch, dstSwitch) = topology.getActiveSwitches()[0..1]
-        defaultRules = northboundService.getSwitchRules(srcSwitch.dpId).flowEntries
+        srcSwDefaultRules = northboundService.getSwitchRules(srcSwitch.dpId).flowEntries
+        dstSwDefaultRules = northboundService.getSwitchRules(srcSwitch.dpId).flowEntries
```
Settis: Maybe you need to use `dstSwitch` here.

ylobankov: Definitely YES! Thanks!

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
```diff
@@ -38,13 +38,16 @@ class SwitchRulesSpec extends BaseSpecification {
...
+    @Shared
+    List dstSwDefaultRules
     @Shared
     int flowRulesCount = 2
 
     def setup() {
         (srcSwitch, dstSwitch) = topology.getActiveSwitches()[0..1]
-        defaultRules = northboundService.getSwitchRules(srcSwitch.dpId).flowEntries
+        srcSwDefaultRules = northboundService.getSwitchRules(srcSwitch.dpId).flowEntries
+        dstSwDefaultRules = northboundService.getSwitchRules(dstSwitch.dpId).flowEntries
```
rtretyak: I'm not sure whether it's just easier to pull certain rules for certain switch directly in the test rather than pull rules for both switches for every test in the spec. I bet in most cases you only need one of them or neither

ylobankov: Yes, you are right. But I don't think that is a heavy operation. Let's have such a variant of fixing the test until I come up with a new solution.

PR #1637, opened 2018-11-23
Extend test for verifying switch port config actions (port DOWN/UP)

### LINKED ISSUE #1378: Add a test to verify switch port config actions for all kinds of hardware switches

ylobankov:
Closes #1378.

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchPortConfigSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchPortConfigSpec.groovy
@@ -6,6 +6,8 @@ import org.openkilda.functionaltests.BaseSpecification
 import org.openkilda.functionaltests.helpers.Wrappers
 import org.openkilda.messaging.info.event.IslChangeType
 import org.openkilda.testing.model.topology.TopologyDefinition
+import org.openkilda.testing.model.topology.TopologyDefinition.Isl
+import org.openkilda.testing.model.topology.TopologyDefinition.Switch
 import org.openkilda.testing.service.northbound.NorthboundService
 import org.openkilda.testing.service.otsdb.OtsdbQueryService
 import org.openkilda.testing.tools.IslUtils
@@ -13,6 +15,7 @@ import org.openkilda.testing.tools.IslUtils
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.beans.factory.annotation.Value
 import spock.lang.Narrative
+import spock.lang.Unroll
 
 @Narrative("Verify that Kilda allows to properly control port state on switches (bring ports up or down).")
 class SwitchPortConfigSpec extends BaseSpecification {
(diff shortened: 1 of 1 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchPortConfigSpec.groovy
```diff
@@ -46,7 +47,7 @@ class SwitchPortConfigSpec extends BaseSpecification {
             assert islUtils.getIslInfo(islUtils.reverseIsl(isl)).get().state == IslChangeType.FAILED
         }
 
-        and: "Port failure is logged in OTSDB"
+        and: "Port failure is logged in OpenTSDB"
         def statsData = [:]
         Wrappers.wait(WAIT_OFFSET) {
             statsData = otsdb.query(portDownTime, "pen.switch.state",
```
npliashechnikov: May be a good idea for the future refactoring to move OpenTSDB metrics names to a designated place (as there were some discussions about possible renaming of them)

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchPortConfigSpec.groovy
```diff
@@ -65,34 +66,54 @@ class SwitchPortConfigSpec extends BaseSpecification {
...
+        and: "Port UP event is logged in OpenTSDB"
         Wrappers.wait(WAIT_OFFSET) {
             statsData = otsdb.query(portUpTime, "pen.switch.state",
                     [switchid: isl.srcSwitch.dpId.toOtsdFormat(), port: isl.srcPort]).dps
             assert statsData.size() == 1
         }
         statsData.values().first() == otsdbPortUp
+
+        where:
+        isl << uniqueIsls
```
npliashechnikov: Do we need to do the same checks on destination switch, or the source switch should be enough?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchPortConfigSpec.groovy
rtretyak: I have pretty much the same opinion as with moving out the error messages: we can/will do it if majority will decide so, but my vote will go for keeping test data as close to the test itself as possible (since these strings will never be reused anywhere else. Otherwise, of course moving them out is a must)

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchPortConfigSpec.groovy
rtretyak: Source and destination switches here are used in a bit different context, since we are not working with 'flows'. For ISLs it actually doesn't matter src or dst switch is used, since there is always a reversed version of that ISL where src and dst are reversed. Or am I missing some idea?

ylobankov: It is not important whether it is a source switch or not. Here we are collecting ISLs where sources switches will be unique switches depending on manufacturer and firmware version.  We can collect the same ISLs where such switches will be destination switches and bring ports DOWN/UP on them.

rtretyak: > We can collect the same ISLs where such switches will be destination switches and bring ports DOWN/UP on them.

...but this makes no sense, at least for me

ylobankov: Yes, I tried to say the same.

Issue #1573, opened 2018-11-11
Put STATS flag for all meters and collect this stats to OTSDB


pzakatov:
http://flowgrammable.org/sdn/openflow/message-layer/metermod/

**NOTE: We assume that each meter has only one meterBand.**

There two types of meters on switch at this moment:
1. Meters for flows.
2. Meters for default rules.

**Metrics for flows:**

- pen.flow.meter.bits        - meterBand bits
- pen.flow.meter.bytes     - meterBand bytes
- pen.flow.meter.packets - meterBand packets

Tags: cookie, switchid, flowid, direction, meterid

**Metrics for default rule meters:**

- pen.switch.flow.system.meter.packets - meterBand packets
- pen.switch.flow.system.meter.bytes     - meterBand bytes
- pen.switch.flow.system.meter.bits        - meterBand bits

Tags: cookieHex, switchid, meterid

pzakatov:
@niksv, please make a change to metermod with stats flag enabled as a separate PR ASAP.

surabujin:
@pzakatov 
> Tags: cookie (in Hex), switchid, meterid

It breaks current stats representation - now tag "cookie" store cookie value in decimal(signed long) representation.

As for me it is better to make 2 tags - `cookie` for decimal representation and `cookieHex` for hex representation (If we accept this "format" we should add "cookieHex" tag for all other records with "cookie" tag).

pzakatov:
> As for me it is better to make 2 tags - `cookie` for decimal representation and `cookieHex` for hex representation (If we accept this "format" we should add "cookieHex" tag for all other records with "cookie" tag).

Ok. Let it be 2 tags

PR #1641, opened 2018-11-26
Added STATS flag to meters

niksv:
Added `STATS` flag to meter bands to be able to collect statistics for these meters.

Part of #1573


surabujin:
And... why you decide to split implementation of #1573 on several PRs?

pzakatov:
> And... why you decide to split implementation of #1573 on several PRs?

@surabujin that was a request from Jon

Issue #1636, opened 2018-11-23
Throttling reroute sometimes won't renew timeout windows when different flow reroutes are issued


rtretyak:
**Steps**
1. Create 2 flows
2. Init reroute of the first flow
3. Right before the throttling windows should end and the flow reroute should be performed - issue a reroute for the second flow
**Expected:** reroute of *both* flows is delayed for a 'rerouteTimeout' time
**Actual:** reroute of the first flow is not additionally delayed. Reroute of the second flow does not influence the reroute time for the first flow.
Note: looks like a floating issue, may not reproduce from the first time.
Functional test: ThrottlingRerouteSpec#"Reroute timer is refreshed even if another flow reroute is issued"

timofei-durakov:
Reroute throttling should be capable for suspending
reroute of all affected flows in case of network flapping.
However due to parallelism of ThrotthlingBolt=2 and
field grouping by flow id it's not the case. Since only
50% of flows are treated by one worker.
See  #1642

PR #1642, opened 2018-11-26
Change ThrotthlingBolt parallelism to 1

timofei-durakov:
Reroute throtthling should be capable for suspending
reroute of all affected flows in case of network flapping.
However due to parallelism of ThrotthlingBolt=2 and
field grouping by flow id it's not the case. Since only
50% of flows are threated by one worker.
This patch change this level of parallelism as a hotfix.
While it should be refactored so it will be possible to
set level of parallelism more than 1.

RelatedTo: #1636


confd/templates/wfm/topology.properties.tmpl
```diff
@@ -1,6 +1,6 @@
 environment.naming.prefix = {{ getv "/kilda_environment_naming_prefix" }}
 
-parallelism.new = 2
+parallelism.new = 1
```
surabujin: Too abstract option... Because it is hotfix, we should hardcode 1 in correct place or add separate option for parallelism of reroute throttling bolt. I vote for hardcoded value with corresponding comment in code.

PS I saw that this option is used only for throttling bolt.

timofei-durakov: Agree. Updated PR.

PR #1648, opened 2018-11-27
Fix code style violations in FlowValidationNegativeSpec file

ylobankov:


ylobankov:
Violations will be fixed while refactoring the spec in the nearest future.

PR #1651, opened 2018-11-27
Statistics service optimizations

niksv:
- mapper methods were moved to separate class
- removed unused exceptions
- added missing getter to data class
- added tests for mapping


services/src/messaging/src/main/java/org/openkilda/messaging/info/stats/FlowStatsReply.java
```diff
@@ -36,6 +36,10 @@ public FlowStatsReply(@JsonProperty("xid") long xid, @JsonProperty("entries") Li
         this.entries = entries;
     }
 
+    public long getXid() {
```
surabujin: Em?.. Why for?

PR #1656, opened 2018-11-27
On-demand flow pings hang if using big timeout

### LINKED ISSUE #1416: On-demand flow pings hang if using big timeout
1. Create a flow
2. Break half of the flow by removing rules somewhere on the intermediate switch
3. Issue a flow ping with timeout set to 4000+ ms
*Expected:* response showing that flow is partially broken
*Actual:* no response, NB returns timeout error
Note: this seems to be an issue only for a partially-working flow. Healthy flows seem to work fine

IlyaMoiseev:
Closes #1416

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/1656)

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/ping/bolt/GroupCollector.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/ping/bolt/GroupCollector.java
@@ -28,7 +28,9 @@ import org.apache.storm.tuple.Fields;
 import org.apache.storm.tuple.Tuple;
 import org.apache.storm.tuple.Values;
 
+import java.util.List;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 public class GroupCollector extends Abstract {
     public static final String BOLT_ID = ComponentId.GROUP_COLLECTOR.toString();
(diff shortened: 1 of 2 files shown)


services/wfm/src/main/java/org/openkilda/wfm/topology/ping/bolt/GroupCollector.java
```diff
@@ -80,7 +89,9 @@ private void collect(Tuple input) throws AbstractException {
     private CollectorDescriptor saveCurrentRecord(Tuple input) throws AbstractException {
         final PingContext pingContext = pullPingContext(input);
 
-        long expireAt = System.currentTimeMillis() + expireDelay;
+        // expiring is only a memory leakage protection in this place
+        // all timeout responses must be fired in upstream
```
sergii-iakovenko: It's better to add a comment with explanation of what this sum of timeouts and delays means.

surabujin: Timeout error produced by `TimeoutManager`. This bolt is responsible for collectiong ping pairs (one flow ping NB-request produce 2 pings - one for each flow direction).

If timeout is reached, `TimeoutManager` fill error field in pingContext object and send it to results processors. If we do not emit this timeout's or lost them somewhere, we should check this execution branch.

PS What is the reason to change this expiration time?

IlyaMoiseev: Good point, updated

IlyaMoiseev: What if on-demand ping timeout is bigger than expireDelay constant? `TimeoutManager` doesnt emit timeout yet, but in `GroupCollector` it's allready expired and dropped
Updated formula is: expireAt = current time + ping context delay ( allow `TimeoutManager` emit timeouts) + some delay for internal storm processing (for tick tuples emitting and bolt communication)

PR #1660, opened 2018-11-27
Bfd session management on FL side

### LINKED ISSUE #1485: Implement BFD session management on FL side
Noviflow provide "experimental" OFAction to configure BFD session. Also we need to setup OFMatch rule to redirect BFD traffic into switch CPU.

New FL commands:
* SetupBFDSession
* RemoveBFDSession

Part of #569

surabujin:
Close #1485


surabujin:
Depends on #1601

PR #1668, opened 2018-11-28
Include sample ElasticSearch credentials in kilda.properties

npliashechnikov:


services/src/atdd-staging/kilda.properties.example
```diff
@@ -10,4 +10,8 @@ floodlight.endpoint=http://localhost:8081
 floodlight.username=username
 floodlight.password=password
 
+elasticsearch.endpoint = http://elasticsearch.pendev:9200
```
ylobankov: Let's use 'localhost' instead of 'elasticsearch.pendev' as it is done for other properties.

services/src/functional-tests/kilda.properties.example
```diff
@@ -16,6 +16,10 @@ neo.uri=bolt://localhost:7687
 lab-api.endpoint=http://localhost:8288
 
 otsdb.endpoint=http://localhost:4242
+elasticsearch.endpoint = https://elasticsearch.pendev:9200
```
ylobankov: Ditto.

services/src/functional-tests/kilda.properties.example
```diff
@@ -16,6 +16,10 @@ neo.uri=bolt://localhost:7687
...
+elasticsearch.username = username
+elasticsearch.password = password
+
```
ylobankov: Please, remove the excess line.

services/src/atdd-staging/kilda.properties.example
npliashechnikov: sure

surabujin:
Squash commits.

PR #1673, opened 2018-11-30
Set meter burst as 10% of requested rate

### LINKED ISSUE #1671: Set meter burst as 10% of requested rate
new config options:
- flow-meter-burst-coefficient - This coefficient is used to calculate burst size for flow meters. Burst size will be equal to 'coefficient * bandwidth'. Value '0.1' means that burst size will be equal to 10% of flow bandwidth.
- flow-meter-min-burst-size-in-kbits - Minimum possible burst size for flow meters. It will be used instead of calculated flow meter burst size if calculated value will be less than value of this option. 
- system-meter-burst-size-in-packets - This is burst size for default rule meters in packets.

niksv:
Closes #1671

### DIFF
--- a/confd/templates/floodlight/floodlightkilda.properties.tmpl
+++ b/confd/templates/floodlight/floodlightkilda.properties.tmpl
@@ -80,3 +80,6 @@ org.openkilda.floodlight.switchmanager.SwitchManager.connect-mode=AUTO
 org.openkilda.floodlight.switchmanager.SwitchManager.broadcast-rate-limit=200
 org.openkilda.floodlight.switchmanager.SwitchManager.unicast-rate-limit=200
 org.openkilda.floodlight.switchmanager.SwitchManager.disco-packet-size=250
+org.openkilda.floodlight.switchmanager.SwitchManager.flow-meter-burst-coefficient=0.1
+org.openkilda.floodlight.switchmanager.SwitchManager.flow-meter-min-burst-size-in-kbits=1024
+org.openkilda.floodlight.switchmanager.SwitchManager.system-meter-burst-size-in-packets=4096
--- a/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManagerConfig.java
+++ b/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManagerConfig.java
@@ -16,7 +16,11 @@
 package org.openkilda.floodlight.switchmanager;
 
 import com.sabre.oss.conf4j.annotation.Configuration;
+import com.sabre.oss.conf4j.annotation.Default;
+import com.sabre.oss.conf4j.annotation.Description;
 import com.sabre.oss.conf4j.annotation.Key;
+
+import javax.validation.constraints.Min;
 
 @Configuration
 public interface SwitchManagerConfig {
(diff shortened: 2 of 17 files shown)


services/src/floodlight-modules/src/test/java/org/openkilda/floodlight/kafka/ReplaceInstallFlowTest.java
```diff
@@ -136,7 +136,7 @@ protected void initScheme() {
     public void installOneSwitchNoneFlow() throws IOException, InterruptedException {
         String value = Resources.toString(getClass().getResource("/install_one_switch_none_flow.json"), Charsets.UTF_8);
         InstallOneSwitchFlow data = (InstallOneSwitchFlow) prepareData(value);
-        OFMeterMod meterCommand = scheme.installMeter(data.getBandwidth(), 1024, data.getMeterId());
+        OFMeterMod meterCommand = scheme.installMeter(data.getBandwidth(), data.getBandwidth() / 10, data.getMeterId());
```
nikitacherevko: I'd like to suggest to move such "coefficient" to one common variable (or even configurable option), so we won't have to change it in every method call like you did.

niksv: good point

confd/templates/floodlight/floodlightkilda.properties.tmpl
```diff
@@ -80,3 +80,5 @@ org.openkilda.floodlight.switchmanager.SwitchManager.connect-mode=AUTO
 org.openkilda.floodlight.switchmanager.SwitchManager.broadcast-rate-limit=200
 org.openkilda.floodlight.switchmanager.SwitchManager.unicast-rate-limit=200
 org.openkilda.floodlight.switchmanager.SwitchManager.disco-packet-size=250
+org.openkilda.floodlight.switchmanager.SwitchManager.flow-meter-burst-coefficient=0.1
+org.openkilda.floodlight.switchmanager.SwitchManager.system-rule-meter-burst-coefficient=0.1
```
surabujin: `-rule` is excess piece in this name. To make both names alike, is should be `system-meter-burst-coefficient`.

(minor)

services/src/floodlight-modules/src/test/java/org/openkilda/floodlight/Constants.java
```diff
@@ -27,6 +27,6 @@
     public static final int inputVlanId = 300;
     public static final long bandwidth = 10000;
     public static final long meterId = 1;
-    public static final long burstSize = 1024;
+    public static final long burstSize = bandwidth / 10;
```
surabujin: There is only 1 usage, we can drop this "constant".

(minor)

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManagerConfig.java
```diff
@@ -31,4 +33,19 @@
 
     @Key("disco-packet-size")
     int getDiscoPacketSize();
+
+    @Key("flow-meter-burst-coefficient")
+    @Default("0.1")
+    @Description("This coefficient is used to calculate burst size for flow meters. " +
+                 "Burst size will be equal to 'coefficient * bandwidth'. " +
+                 "Value '0.1' means that burst size will be equal to 10% of flow bandwidth.")
+    double getFlowMeterBurstCoefficient();
```
surabujin: We should set `min` limit.

niksv: Unfortunately @Min annotation doesn't work for double and float:

```
 * Note that {@code double} and {@code float} are not supported due to rounding errors
 * (some providers might provide some approximative support).
```

niksv: Is there some other way to do it?

confd/templates/floodlight/floodlightkilda.properties.tmpl
```diff
@@ -80,3 +80,6 @@ org.openkilda.floodlight.switchmanager.SwitchManager.connect-mode=AUTO
 org.openkilda.floodlight.switchmanager.SwitchManager.broadcast-rate-limit=200
 org.openkilda.floodlight.switchmanager.SwitchManager.unicast-rate-limit=200
 org.openkilda.floodlight.switchmanager.SwitchManager.disco-packet-size=250
+org.openkilda.floodlight.switchmanager.SwitchManager.flow-meter-burst-coefficient=0.1
+org.openkilda.floodlight.switchmanager.SwitchManager.flow-meter-min-burst-size-in-kbits=1024
+org.openkilda.floodlight.switchmanager.SwitchManager.system-meter-burst-size-in-kbits=1024
```
surabujin: system rules rate are set in pktps... it will be more natural to set burls in pktps for them too.

niksv: will fix

PR #1674, opened 2018-11-30
Add message serializer/deseriazer for storm spout and kafka bolt

nikitacherevko:
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/1674)

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/AbstractTopology.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/AbstractTopology.java
@@ -19,6 +19,7 @@ import static java.lang.String.format;
 
 import org.openkilda.config.KafkaConfig;
 import org.openkilda.config.naming.KafkaNamingStrategy;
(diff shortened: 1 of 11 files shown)


services/wfm/src/main/java/org/openkilda/wfm/topology/AbstractTopology.java
```diff
@@ -202,27 +205,64 @@ public final T getConfig() {
      *
      * @param topic Kafka topic
      * @return {@link KafkaSpout}
+     * @deprecated should be replaced by {@link AbstractTopology#buildKafkaSpout(String, String)}.
      */
+    @Deprecated
```
surabujin: We need this method, at least for "stats". They need to call `.setFirstPollOffsetStrategy(...)` by their own.

I like builder approach more than serveral set of hardcoded configs.

nikitacherevko: you are right, I'll fix it

Issue #905, opened 2018-06-19
Atdd-staging: Cover switch rules CRUD endpoints in NB


rtretyak:
Most of the cases exist in ATDD framework under NorthboundRun.feature
- [x] get switch rules
- [ ] delete switch rules:
  - [x] DROP_ALL
  - [ ] DROP_ALL_ADD_DEFAULTS
  - [x] IGNORE_DEFAULTS
  - [ ] OVERWRITE_DEFAULTS
  - [ ] REMOVE_DROP
  - [ ] REMOVE_BROADCAST
  - [ ] REMOVE_UNICAST
  - [ ] REMOVE_DEFAULTS
  - [ ] REMOVE_ADD_DEFAULTS
  - [x] delete by cookie
  - [x] delete by in-port
  - [x] delete by in-vlan
  - [x] delete by priority
  - [x] delete by out-port
  - [x] combinations of the above
- [ ] validate switch rules:
  - [x] missing rules on egress switch
  - [x] missing rules on ingress switch
  - [x] missing rules on transit switch
  - [ ] missing default rules
  - [x] missing non-default rules
- [x] sync switch rules

rtretyak:
Moved from TODO to Backlog. This is at lower priority for now

PR #1690, opened 2018-12-03
Add tests to verify switch rules deletion by various delete actions

ylobankov:
Part of #905.

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
@@ -117,27 +117,26 @@ class SwitchRulesSpec extends BaseSpecification {
     }
 
     @Unroll
-    def "Able to delete #data.description rules from a switch"() {
+    def "Able to delete rules from a switch (delete-action=#data.deleteRulesAction)"() {
         given: "A switch with some flow rules installed"
         def flow = flowHelper.randomFlow(srcSwitch, dstSwitch)
         flowHelper.addFlow(flow)
 
-        when: "Delete #data.description rules from the switch"
+        when: "Delete rules from the switch"
+        def expectedRules = data.getExpectedRules(srcSwitch.dpId, srcSwDefaultRules)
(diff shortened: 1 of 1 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
```diff
@@ -115,49 +116,79 @@ class SwitchRulesSpec extends BaseSpecification {
...
         given: "A switch with some flow rules installed"
         def flow = flowHelper.randomFlow(srcSwitch, dstSwitch)
         flowHelper.addFlow(flow)
 
-        when: "Delete #data.description rules from the switch"
+        when: "Delete rules from the switch"
+        def flowRules = getFlowRules(srcSwitch.dpId)
         def deletedRules = northboundService.deleteSwitchRules(srcSwitch.dpId, data.deleteRulesAction)
+        def expectedRules = srcSwDefaultRules
+        switch (data.deleteRulesAction) {
```
rtretyak: I don't really like all these `if`s. Test is getting too complicated. You can put the `expectedRules` variable inside your `where` block instead of calculating it for different cases inside the test. This will make your `where` block as well as test itself to be more readable, because you will have a clear expected result directly in where for every delete action(every iteration), instead of calculating it in two different places in test for different scenarios. Since you cannot directly calculate it in 'where'(because you don't know the exact flowRules), you can supply a 'getExpectedRules' closure in the `where`, that will be called during test execution and take current flowRules as param (similar to `"Unable to create flow with conflicting vlans or flowIds"` test)   
Also, why on line 142, for the DROP_ALL action, you can't compare actualRules with an empty array? For DROP_ALL the expectedRules is an empty array, isn't it? I mean it would be at least less painful if we have all our `if`s in one place (inside this `switch`) instead of having multiple places where we split the logic. In this case you can completely remove the `rulesRemained` variable as well, because you will run 'compareRules' for all cases and it takes size into account
I don't heavily insist on these changes, just shared my point of view

ylobankov: Fixed.

PR #1707, opened 2018-12-05
Add tests to verify deleting default switch rule by delete action

ylobankov:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
```diff
@@ -177,6 +178,56 @@ class SwitchRulesSpec extends BaseSpecification {
         ]
     }
 
+    @Unroll
+    def "Able to delete default rule from a switch (delete-action=#data.deleteRulesAction)"() {
+        assumeFalse("Unable to run the test because an OF_ 12 switch has one broadcast rule as the default",
+                dstSwitch.ofVersion == "OF_12" && data.cookie != DefaultRule.VERIFICATION_BROADCAST_RULE.cookie)
```
rtretyak: Does it make sense to be more specific on which switches we use for test? Under certain circumstances it is possible that the dstSwitch will always be OF_12 and we will only test 1 scenario out of 4. Though this is not too important, because we virtual runs are tested with OF_13

ylobankov: I will fix it in the next patches. In the future, I think we can take unique firmware switches as we do in some of our tests.

Issue #1711, opened 2018-12-05
Ensure new rules installed before removing old ones on reroute


sergii-iakovenko:
Ensure new switch rules installed before removing old ones on flow reroute, in order to eliminate cases 
when a flow becomes broken due to unsuccessful rule setup.

PR #1712, opened 2018-12-05
Design for multi-stage process of switch rule installation

sergii-iakovenko:
Design proposal for #1711

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/1712)


nikitamarchenko:
Could you add a case with checkForExpiredCommands when fl is not responding?

PR #1716, opened 2018-12-07
hotfix for flow validator

timofei-durakov:
flow validator mixed up source and destination endpoints
and blocked flow creation in false-positive way.
This patch fixes that issue

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/flow/validation/FlowValidator.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/flow/validation/FlowValidator.java
@@ -83,9 +83,15 @@ public class FlowValidator {
 
 
         Optional<String> conflictedFlow = conflictsOnSource.stream()
-                .filter(flow -> !requestedFlow.getFlowId().equals(flow.getFlowId())
-                        && (flow.getSrcVlan() == 0 || flow.getSrcVlan() == requestedFlow.getSrcVlan()
+                .filter(flow -> !requestedFlow.getFlowId().equals(flow.getFlowId()))
+                .filter(flow -> (flow.getSrcSwitch().getSwitchId().equals(requestedFlow.getSrcSwitch().getSwitchId())
+                            && flow.getSrcPort() == requestedFlow.getSrcPort()
+                            && (flow.getSrcVlan() == requestedFlow.getSrcVlan() || flow.getSrcVlan() == 0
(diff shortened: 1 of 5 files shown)


services/wfm/src/main/java/org/openkilda/wfm/topology/flow/validation/FlowValidator.java
```diff
@@ -83,9 +83,15 @@ void checkFlowForEndpointConflicts(Flow requestedFlow) throws FlowValidationExce
 
 
         Optional<String> conflictedFlow = conflictsOnSource.stream()
-                .filter(flow -> !requestedFlow.getFlowId().equals(flow.getFlowId())
-                        && (flow.getSrcVlan() == 0 || flow.getSrcVlan() == requestedFlow.getSrcVlan()
+                .filter(flow -> !requestedFlow.getFlowId().equals(flow.getFlowId()))
```
surabujin: I am sorry, but this is piece of crap - extremely overcomplicated logic plus copy & pasted twice.

We have 2 point (endpoint) we need to check is there any flow that use same endpoints. So take first endpoint (flow source) and lookup this endpoint in persistent storage (it can be source or end endpoint of existing flow) if find - report conflict, if not take second endpoint and lookup it in persistent storage - report conflict if find.

So we need to translate stream of flows loaded from persistent storage into stream of endpoints, than lookup match.

services/wfm/src/main/java/org/openkilda/wfm/topology/flow/validation/FlowValidator.java
```diff
@@ -105,9 +111,15 @@ void checkFlowForEndpointConflicts(Flow requestedFlow) throws FlowValidationExce
 
 
         conflictedFlow = conflictsOnDest.stream()
-                .filter(flow -> !requestedFlow.getFlowId().equals(flow.getFlowId())
-                        && (flow.getDestVlan() == 0 || flow.getDestVlan() == requestedFlow.getDestVlan()
-                        || requestedFlow.getDestVlan() == 0))
+                .filter(flow -> !requestedFlow.getFlowId().equals(flow.getFlowId()))
```
carmine: With these expanded set of conflict tests, the code becomes harder to read. There is no comment that explains what is going on in plain english. I find it is best to include a simple description of what the function does .. ie Endpoint Conflicts .. and in this case, how it does it - ie what are the conditions.

PR #1718, opened 2018-12-07
Add positive tests for creating flows with different vlan combinations

rtretyak:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchFailuresSpec.groovy
```diff
@@ -134,4 +140,55 @@ class SwitchFailuresSpec extends BaseSpecification {
...
+        def flow = flowHelper.randomFlow(srcSwitch, dstSwitch)
+        flowHelper.addFlow(flow)
+        def currentPath = PathHelper.convert(northboundService.getFlowPath(flow.id))
+
+        and: "There is a more preferable alternative path"
+        def alternativePaths = allPaths.findAll { it != currentPath }
+        def preferredPath = alternativePaths.first()
+        def uniqueSwitch = pathHelper.getInvolvedSwitches(preferredPath).find {
+            !pathHelper.getInvolvedSwitches(currentPath).contains(it)
+        }
```
ylobankov: I would rather add assumeTrue() here with some message instead of asserting.

rtretyak: agree

PR #1725, opened 2018-12-10
Re-implement flow filtering for periodic pings via FlowRepository.

sergii-iakovenko:
The changes remove excessive reads from database by Ping topology: now it filters flows before fetching them.


services/src/kilda-core/kilda-persistence-neo4j/src/main/java/org/openkilda/persistence/repositories/impl/Neo4jFlowRepository.java
```diff
@@ -80,6 +81,15 @@ public boolean exists(String flowId) {
         return buildFlowPairs(findAll());
     }
 
+    @Override
+    public Collection<FlowPair> findFlowPairsWithPeriodicPingsEnabled() {
```
IlyaMoiseev: It will be good to create unit test for new repository method

PR #1728, opened 2018-12-10
Avoid extra fetching of entity relations with OGM load strategy depth 1.

sergii-iakovenko:
The changes prevent extra fetching of entity relations from database. These reads happen because Neo4j OGM handles load strategy"depth" improperly: when a relation entity is being loaded, OGM fetches ALL relations of start and end nodes of the requested relation. Even with the "depth" = 1.


services/src/kilda-core/kilda-model/src/main/java/org/openkilda/model/FlowSegment.java
```diff
@@ -53,9 +54,11 @@
 
     @NonNull
     @Property(name = "flowid")
+    @Required
     private String flowId;
 
     @Property(name = "cookie")
+    @Required
```
IlyaMoiseev: jfyi, https://github.com/neo4j/neo4j-ogm/blob/master/core/src/main/java/org/neo4j/ogm/annotation/Required.java#L28

PR #1737, opened 2018-12-11
Add missing properties to confd configuration

ylobankov:


npliashechnikov:
Could you please also add a more detailed description to mode.topology.single variable in kilda.properties.example? It is misleading now.

npliashechnikov:
Also, it looks like the mode.topology.single property is missing from kilda.properties.tmpl
Please make sure it gets there, as it is necessary to start tests using hardware profile.

PR #1738, opened 2018-12-11
Anti-flap for port down/up.

### LINKED ISSUE #1729: Anti-flap for port down/up
# Problem
Sometimes port with ISL on it bouncing within a short period of time which is not a real problem. We initiate ISL_down immediately.

# Proposal
Implement similar antiflapping approach as for reroute. Introduce a logic where we wait for a "Port UP" event after a "Port Down" for "throttling_delay_seconds_min" seconds. if the last event in a series is "UP", then do not initiate "ISL_Down".

# Details 
The bolt has states: `Nothing`, `WarmingUp`, and `CoolingDown` for each port.
Initially, the bolt has state `Nothing`.
If the bolt receives "Port Up" in `Nothing` state it must emit "Port Up" without state change.
If the bolt receives "Port Down" in `Nothing` state it changes state to `WarmingUp`.
If the bolt does not receive "Port Up" after `delay_seconds_min`, it emits "Port Down" event and goes to `CoolingDown` state.
If In `CoolingDown` state the bolt receives event, it calculates current port state and restarts `seconds_cool_down` timer. 
After `seconds_cool_down` of not receiving any events, state changes to `Nothing`. If the current port state is Up the bolt emit "Port Up" event.

In case with `WarmingUp` state, when delays between every "Port Down" / "Port Up" were less then `delay_seconds_min` and the last "Port Down" / "Port Up" event was during last `delay_seconds_min` period of `delay_seconds_warm_up,` the bolt emit "Port Down" event and go to `CoolingDown` state.

Defaults should be:
- kilda_port_up_down_throttling_delay_seconds_min: 1
- kilda_port_up_down_throttling_delay_seconds_warm_up: 2
- kilda_port_up_down_throttling_delay_seconds_cool_down: 5

Settis:
Closes #1729

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/event/service/PortEventThrottlingService.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/event/service/PortEventThrottlingService.java
@@ -1,0 +1,217 @@
+/* Copyright 2018 Telstra Open Source
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+package org.openkilda.wfm.topology.event.service;
+
+import org.openkilda.messaging.info.event.PortChangeType;
+import org.openkilda.messaging.info.event.PortInfoData;
+import org.openkilda.model.SwitchId;
+
+import lombok.Value;
+import lombok.extern.slf4j.Slf4j;
+
+import java.time.Clock;
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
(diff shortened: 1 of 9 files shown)


services/wfm/src/main/java/org/openkilda/wfm/topology/event/service/PortEventThrottlingService.java
```diff
@@ -0,0 +1,217 @@
...
+     * Processing an event of PortInfoData.
+     * Store it internally for Throttling or return true if Port in UP state and no case is opened for the Port.
+     *
+     * @param data the PortInfoData object.
+     * @return true if the event can be sent further.
+     */
+    public boolean processEvent(PortInfoData data, String correlationId) {
+        SwitchPort switchPort = new SwitchPort(data.getSwitchId(), data.getPortNo());
+        PortState portState = portStats.get(switchPort);
+        if (data.getState() == PortChangeType.UP && portState == null) {
```
surabujin: Should we pass "first" PORT_DOWN without delay too? Event more - we should not start throttling mechanism on PORT_UP event.

It look... natural to pass first PORT_DOWN and it port will continue flapping - drop following UP/DOWN events. When we detect end of flapping - we should emit finite port state.

PS We should not emit 2 PORT_DOWN event in a row.

Settis: According design - no. We want to hide port flapping. In case when event topology receive `Port Down` and after 300 ms  `Port Up`. The bolt will do nothing, just ignore this.
If it waits `Port Up` for too long, it fire only one `Port Down` and have behavior like you describe above.

PR #1746, opened 2018-12-12
Add test to verify symmetric flow creation

ylobankov:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/PathHelper.groovy
```diff
@@ -37,8 +37,10 @@ class PathHelper {
      */
     void makePathMorePreferable(List<PathNode> morePreferablePath, List<PathNode> lessPreferablePath) {
         def morePreferableIsls = getInvolvedIsls(morePreferablePath)
-        def islToAvoid = getInvolvedIsls(lessPreferablePath).find { !morePreferableIsls.contains(it) &&
-                !morePreferableIsls.contains(islUtils.reverseIsl(it)) }
+        def islToAvoid = getInvolvedIsls(lessPreferablePath).find {
+            !morePreferableIsls.contains(it) &&
```
npliashechnikov: That formatting is kind of strange

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/PathHelper.groovy
```diff
@@ -87,8 +89,8 @@ class PathHelper {
     /**
      * Converts FlowPathPayload path representation to a List<PathNode> representation
      */
-    static List<PathNode> convert(FlowPathPayload pathPayload) {
-        def path = pathPayload.forwardPath
+    static List<PathNode> convert(FlowPathPayload pathPayload, pathToConvert = "forwardPath") {
```
npliashechnikov: If I understand correctly, this method previously converted the both flows. Are you sure it won't affect existing tests? Maybe it'd be better to create a separate method?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowCrudSpec.groovy
```diff
@@ -370,6 +375,42 @@ class FlowCrudSpec extends BaseSpecification {
...
+        def (Switch srcSwitch, Switch dstSwitch) = [switches, switches].combinations()
+                .findAll { src, dst -> src != dst }.find { Switch src, Switch dst ->
+            possibleFlowPaths = db.getPaths(src.dpId, dst.dpId)*.path.sort { it.size() }
+            possibleFlowPaths.size() > 1
+        } ?: assumeTrue("No suiting switches found", false)
+
+        and: "Make the shorter forward path not preferable"
+        pathHelper.getInvolvedIsls(possibleFlowPaths[0]).each { db.updateLinkCost(it, Integer.MAX_VALUE) }
+
+        and: "Make the longer reverse path not preferable"
```
npliashechnikov: So we have cheap longer forward path and cheap short reverse path, correct?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/PathHelper.groovy
ylobankov: This method previously converted the forward path only. Please take a look at L93. Now it can convert both paths but the forward path is converted by default. So it shouldn't break existing tests.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/PathHelper.groovy
ylobankov: Ok, will fix.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowCrudSpec.groovy
ylobankov: Cleanup is done at L411.

PR #1777, opened 2018-12-14
Propose logging conventions.

sergii-iakovenko:
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/1777)


surabujin:
> Endpoint (switch ID and port) must be logged using underscore as a separator, e.g.: 00:00:00:00:00:00:00:01_2

If we will use object(actually we already have `NetworkEndpoint` such object) representing sw+port, we will be able to use `.toString()` method to produce "same" representation.

PS I like `SW0000010203040506-2` more than current and than proposed representations.

PR #1783, opened 2018-12-17
Add test to verify conflict occurs when updating flow

### LINKED ISSUE #1677: Tests: Test flow conflicts during flow update
New test should verify the same things as `"Unable to create flow with conflicting vlans or flowIds"` does, but during flow update. Additional valuable verifications are welcomed if make sense

ylobankov:
Closes #1677.

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowCrudSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowCrudSpec.groovy
@@ -141,69 +141,73 @@ class FlowCrudSpec extends BaseSpecification {
 
         where:
         data << [
-            [
-                description: "same switch-port but vlans on src and dst are swapped",
-                getSecondFlow: { FlowPayload existingFlow ->
-                    return getFlowHelper().randomFlow(findSwitch(existingFlow.source.datapath),
-                            findSwitch(existingFlow.destination.datapath)).tap {
-                        it.source.portNumber = existingFlow.source.portNumber
-                        it.source.vlanId = existingFlow.destination.vlanId
-                        it.destination.portNumber = existingFlow.destination.portNumber
-                        it.destination.vlanId = existingFlow.source.vlanId
-                    }
-                }
-            ],
-            [
-                description: "same switch-port but vlans on src and dst are different",
-                getSecondFlow: { FlowPayload existingFlow ->
-                    return getFlowHelper().randomFlow(findSwitch(existingFlow.source.datapath),
-                            findSwitch(existingFlow.destination.datapath)).tap {
-                        it.source.portNumber = existingFlow.source.portNumber
-                        it.source.vlanId = existingFlow.source.vlanId + 1
-                        it.destination.portNumber = existingFlow.destination.portNumber
-                        it.destination.vlanId = existingFlow.destination.vlanId + 1
-                    }
-                }
-            ],
-            [
-                description: "vlan-port of new src = vlan-port of existing dst (but different switches)",
-                getSecondFlow: { FlowPayload existingFlow ->
-                    //src for new flow will be on different switch not related to existing flow
-                    //thus two flows will have same dst but different src
-                    def newSrcSwitch = getTopology().activeSwitches.find {
-                        ![existingFlow.source.datapath, existingFlow.destination.datapath].contains(it.dpId)
-                    }
-                    return getFlowHelper().randomFlow(newSrcSwitch, findSwitch(existingFlow.destination.datapath)).tap {
-                        it.source.vlanId = existingFlow.destination.vlanId
-                        it.source.portNumber = existingFlow.destination.portNumber
-                        it.destination.vlanId = existingFlow.destination.vlanId + 1
-                    }
-                }
-            ],
-            [
-                description: "vlan-port of new dst = vlan-port of existing src (but different switches)",
-                getSecondFlow: { FlowPayload existingFlow ->
-                    def newSrcSwitch = getTopology().activeSwitches.find {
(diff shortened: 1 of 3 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowCrudSpec.groovy
```diff
@@ -280,99 +280,37 @@ class FlowCrudSpec extends BaseSpecification {
...
+        then: "Error is returned, stating a readable reason of conflict"
+        def error = thrown(HttpClientErrorException)
+        error.statusCode == HttpStatus.CONFLICT
+        error.responseBodyAsString.to(MessageError).errorMessage == data.getError(flow1, "update")
+
+        and: "Cleanup: delete flows"
+        [flow1, flow2].each { flowHelper.deleteFlow(it.id) }
+
+        where:
+        data << getConflictingData()[0..-2]
```
rtretyak: Not critical, but a bit flaky solution, since someone may add additional cases to the end of getConflictingData without knowing that it is sequence-dependent (by default it is not common behavior for 'data provider'). It is better to have a base 'conflictingData' which is the same between tests and add a missing 'same flow id' case directly to the 'create' test.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowCrudSpec.groovy
```diff
@@ -280,99 +280,37 @@ class FlowCrudSpec extends BaseSpecification {
...
-                    "Could not create flow: Flow $flowToError.id already exists"
-                }
-            ]
-        ]
+        where:
+        data << getConflictingData()
+    }
+
+    @Unroll("Unable to update flow with #data.conflict")
+    def "Unable to update flow with conflicting vlans or flow IDs"() {
```
rtretyak: During update there is also a possibility to have initial 2 flows on different switches and then update the src or dst _switch_ so that they conflict. Does it make sense?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowCrudSpec.groovy
```diff
@@ -280,99 +280,37 @@ class FlowCrudSpec extends BaseSpecification {
...
+        given: "Two potential flows"
+        def (Switch srcSwitch, Switch dstSwitch) = topology.activeSwitches
+        def flow1 = flowHelper.randomFlow(srcSwitch, dstSwitch)
+        def flow2 = flowHelper.randomFlow(srcSwitch, dstSwitch)
+
+        def conflictingFlow2 = flowHelper.randomFlow(srcSwitch, dstSwitch)
```
npliashechnikov: If possible, please rename conflictingFlow2 to a "conflictingFlow". It is a bit confusing now.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowCrudSpec.groovy
```diff
@@ -560,4 +499,104 @@ class FlowCrudSpec extends BaseSpecification {
...
+                    flowToConflict.destination.portNumber = dominantFlow.destination.portNumber
+                    flowToConflict.destination.vlanId = 0
+                },
+                getError: { FlowPayload flowToError, String operation = "create" ->
+                    portError(operation, flowToError.destination.portNumber, flowToError.destination.datapath,
+                            flowToError.id)
+                }
+            ],
+            [
+                conflict: "vlan vs no vlan on the same port on src switch",
```
npliashechnikov: Given that a top-level flow basically consists of forward and reverse flows, do we really need both "vlan vs no vlan" and "no vlan vs vlan" test cases?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowCrudSpec.groovy
ylobankov: Done.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowCrudSpec.groovy
ylobankov: Done.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowCrudSpec.groovy
ylobankov: Maybe not but these tests are very fast and I would prefer to keep them. They don't almost take any time.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowCrudSpec.groovy
ylobankov: Probably yes, I am not sure. Anyway, I think it's better to do with another patch.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/flows/FlowCrudSpec.groovy
rtretyak: I added those cases in panic when there was a prod defect with vlans to exclude the possibility of any missing scenarios.

PR #1785, opened 2018-12-17
Add a helper method to update flow

### LINKED ISSUE #1779: Add a method to flow helper to update a flow
_Originally posted by @ylobankov in https://github.com/telstra/open-kilda/pull/1693_

ylobankov:
Closes #1779.

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/FlowHelper.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/FlowHelper.groovy
@@ -114,6 +114,26 @@ class FlowHelper {
     }
 
     /**
+     * Updates flow with checking flow status and rules on source and destination switches.
+     * It is supposed if rules are installed on source and destination switches, the flow is completely updated.
+     */
+    FlowPayload updateFlow(String flowId, FlowPayload flow) {
+        def flowEntryBeforeUpdate = db.getFlow(flowId)
+
+        log.debug("Updating flow '${flow.id}'")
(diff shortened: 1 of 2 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/FlowHelper.groovy
```diff
@@ -108,7 +109,39 @@ class FlowHelper {
...
+    FlowPayload updateFlow(String flowId, FlowPayload flow) {
+        def flowEntryBeforeUpdate = db.getFlow(flowId)
+
+        log.debug("Updating flow '${flow.id}'")
+        def response = northbound.updateFlow(flowId, flow)
+        Wrappers.wait(WAIT_OFFSET) { assert northbound.getFlowStatus(flow.id).status == FlowState.UP }
+
+        def flowEntryAfterUpdate = db.getFlow(flowId)
+
+        if (flowEntryBeforeUpdate.left.sourceSwitch != flowEntryAfterUpdate.left.sourceSwitch) {
```
rtretyak: After update the flow will get new cookies regardless of whether the switch was changed, so you can remove `if` condition. Another thing is that today we released the multi-stage rules installation process, so I guess Kilda should take care of installing new rules first before removing the old ones (you can assume this in your code and check only the latest operation)

ylobankov: Let's keep all checks until we add a new test to verify this functionality.

PR #1786, opened 2018-12-18
Add test for port anti-flap feature

### LINKED ISSUE #1771: Add tests to verify anti-flap behavior for port down/up actions
As a QA operator, I would like to have a certain suite of tests to verify anti-flap behavior for port down/up actions. Tests should go to the functional-tests module. Please refer #1729 for details.

rtretyak:
Closes #1771


confd/templates/atdd-staging/kilda.properties.tmpl
```diff
@@ -27,4 +27,7 @@ reroute.delay={{ getv "/kilda_reroute_throttling_delay_min" }}
 reroute.hardtimeout={{ getv "/kilda_reroute_throttling_delay_max" }}
 discovery.interval={{ getv "/kilda_discovery_interval" }}
 discovery.timeout={{ getv "/kilda_discovery_timeout" }}
+antiflap.min={{ getv "/kilda_port_up_down_throttling_delay_seconds_min" }}
```
npliashechnikov: cosmetics: would be nice to rename 'min' to something like
antiflap.throttling_delay

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/BaseSpecification.groovy
```diff
@@ -26,7 +26,7 @@ class BaseSpecification extends SpringSpecification implements SetupOnce {
     @Autowired
     FlowHelper flowHelper
     @Autowired
-    TopologyDefinition topologyDefinition
```
npliashechnikov: This may break things. We'll have to watch carefully when merging.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/thread/PortBlinker.groovy
```diff
@@ -0,0 +1,49 @@
...
+    synchronized void start() {
+        super.start()
+        timeStarted = new Date()
+    }
+
+    @Override
+    void run() {
+        running = true
+        while(running) {
+            northbound.portDown(sw.dpId, port)
```
npliashechnikov: It may make sense to issue portDown and portUp asynchronously, so that minimum delay between these operations gets lower. IIRC, when testing against staging from my local machine, delay between portDown() and portUp() exceeded 1 second.
Also, to further speed up things, we may have to consider something faster, like directly managing the ports on the switches thru LabService, to avoid NB/Storm/Kafka/Floodlight traversal.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/thread/PortBlinker.groovy
```diff
@@ -0,0 +1,49 @@
...
+class PortBlinker extends Thread {
+    private NorthboundService northbound
+    private volatile boolean running
+
+    final Switch sw
+    final int port
+    Date timeStarted
+    Date timeStopped
+
+    PortBlinker(NorthboundService northbound, Switch sw, int port) {
```
npliashechnikov: Controllable delay between port up/down events would be nice.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/PortAntiflapSpec.groovy
```diff
@@ -0,0 +1,105 @@
...
+import org.springframework.beans.factory.annotation.Value
+import spock.lang.Issue
+import spock.lang.Narrative
+
+import java.util.concurrent.TimeUnit
+
+@Slf4j
+@Narrative("""
+Anti-flap feature is aimed to address the problem of blinking ports. Sometimes it happens that certain port on a switch
+begins to blink for a very short period of time. If it is a short-time behavior, in this case we don't want to init a
```
npliashechnikov: "In this case" looks a bit redundant here

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/PortAntiflapSpec.groovy
```diff
@@ -0,0 +1,105 @@
...
+
+        and: "ISL remains down even after cooldown timeout"
+        TimeUnit.SECONDS.sleep(antiflapCooldown + 1)
+        islUtils.getIslInfo(isl).get().state == IslChangeType.FAILED
+
+        when: "Port stops flapping, ending in UP state"
+        blinker.stop(true)
+
+        then: "Right before cooldown timeout runs out the ISL remains down"
+        sleep(untilCooldownEnds() - 100)
```
npliashechnikov: 100 ms safety delay may be a bit low if we run test against h/w environment

rtretyak: The 'safety' here goes to the opposite direction. It is impossible for this event to occur earlier, but it may occur later due to internal Kilda operation time and network delays. The `untilCooldownEnds` is calculated locally and is a pretty accurate time. It of course will not be the same on the server, but it for sure won't be less than this.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/thread/PortBlinker.groovy
rtretyak: Agree. I thought about this after push, but it is currently not needed anywhere. The class can be extended with new functionality pretty easy, so this can be added on-demand later

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/PortAntiflapSpec.groovy
rtretyak: ok I see a problem here, need to investigate

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/thread/PortBlinker.groovy
rtretyak: Tried the async nb calls, but unfortunately it resulted in even slower port switches. Will investigate the FL switch possibility later

confd/templates/atdd-staging/kilda.properties.tmpl
rtretyak: All 3 of them are `throttling_delay`s. I mean if you read all three vars they only differ by 'min'-'warm_up'-'coold_down'

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/PortAntiflapSpec.groovy
rtretyak: cannot test this properly due to #1790 issue which is caught by the test

rtretyak: changed to -1000 for now

PR #1789, opened 2018-12-18
Labservice improvements

IlyaMoiseev:
- Correct message when controller host resolution fails
- Create only required ports on ASwitch
- Documentation actualization


services/lab-service/lab/service/topology.py
```diff
@@ -94,10 +101,15 @@ def add_controller(self, controller, batch=True):
...
-class ASwitch(Switch):
-    def setup(self, a_mappings):
-        self.add_route_flows(a_mappings)
+class KildaSwitch(Switch):
+    @classmethod
+    def create(cls, sw_def):
+        sw = Switch.create(sw_def)
+        sw.define_ports(range(1, sw_def['max_port'] + 1))
+        return sw
+
```
timofei-durakov: ouch. there are 2 options implement `constructor` for that __init__()
or create a classmethod in Switch class

IlyaMoiseev: updated

PR #1791, opened 2018-12-18
Update flow helper to avoid creating flow requests with the same ports

ylobankov:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/FlowHelper.groovy
```diff
@@ -131,13 +145,17 @@ class FlowHelper {
      */
     private FlowEndpointPayload getFlowEndpoint(Switch sw, List<Integer> allowedPorts,
                                                 boolean useTraffgenPorts = true) {
-        def port = allowedPorts[random.nextInt(allowedPorts.size())]
+        allowedPorts -= getSwitchPortsMap()[sw.dpId]
```
npliashechnikov: It'd look better to just use variable directly instead of getter method

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/FlowHelper.groovy
```diff
@@ -110,6 +121,9 @@ class FlowHelper {
 
         checkRulesOnSwitches(flowEntry, RULES_DELETION_TIME, false)
 
+        switchPortsMap[flowEntry.left.sourceSwitch] -= flowEntry.left.sourcePort
```
npliashechnikov: Given that this thing is managed by Spring and can survive between test cases, I'd rather add appropriate cleanup to the BaseSpecification.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/FlowHelper.groovy
ylobankov: Unfortunately, it is impossible. We need to init a map where the keys are switch IDs and values are ports occupied by flows. It is not impossible to do at L41 because topology is still null and we fail at this point. So this workaround allows us to avoid failures.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/FlowHelper.groovy
ylobankov: Agree.

rtretyak:
I don't like the whole solution.
1. flow map being cleared in spec cleanup is extremely strange. You never explicitly touched it in any test but you are cleaning it up in test cleanup. These two things are absolutely unrelated and coupling them is a clear sign of bad design. I think spring should provide ways to configure your `@Component` to create a new instance each time you autowire it.
2. What if I remove the flow or update the flow? What if I request two random flows and then update them in the spec the way I need (e.g. I update the src port on the second flow before creating it because I need a conflict in my test) -> the flow map instantly gets outdated, holding absolutely irrelevant information. I don't have a quick solution to this off the top of my head.
3. What if I need more flows than there're switch ports available? It is not a fictional situation, there is already a ChaosSpec that works with big amount of flows in the system. I think by default the helper should only guarantee the 'no conflict' flows (avoid same switch same port same vlan situation), but not just avoid using same ports. The fact that your test can't work with same ports is a pretty narrow situation, I would say spec/test/case-specific so I don't see it as a global helper requirement.

Thoughts?

ylobankov:
Temporarily closed.

PR #1811, opened 2018-12-20
Use real SwitchManagerConfig in unit tests

niksv:
We have implementation SwitchManagerConfig for tests with hardcoded values. The problem is that this config is used as expected config and as actual config. Some time ago flowMeterBurstCoefficient was changed from 0.1 to 1.05 but tests still use old 0.1 value. To fix it missed @Default annotations were added to the real config class and now this class is used in tests.

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/1811)

### DIFF
--- a/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
+++ b/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
@@ -1602,9 +1602,4 @@ public class SwitchManager implements IFloodlightModule, IFloodlightService, ISw
                 .findFirst()
                 .orElse(null);
     }
-
-    @VisibleForTesting
-    void setConfig(SwitchManagerConfig config) {
-        this.config = config;
-    }
 }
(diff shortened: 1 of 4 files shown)


services/src/floodlight-modules/src/test/java/org/openkilda/floodlight/config/Config.java
```diff
@@ -0,0 +1,55 @@
...
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+package org.openkilda.floodlight.config;
+
+import org.openkilda.floodlight.switchmanager.SwitchManagerConfig;
+
+public class Config implements SwitchManagerConfig {
```
sergii-iakovenko: Please, add Test prefix/suffix to distinguish from production code.

niksv: sure

Issue #1820, opened 2018-12-24
Evolve PCE with weights computation strategies


IlyaMoiseev:
For now PCE uses only `cost` field as graph edges weights

While path computing logic going to be more complicated - we need to take several factors\costs to compute final edge\node weight, and keep this things maintainable

carmine:
Refactor or evolve?

carmine:
Can we get a concrete example of a weights strategy that we want to put into place?

Arguably, if there is some other numeric field / attribute that we'd like to use, the method signature could change (ie pass in the attribute to use).  If it is more complex, and not calculated a prior, then pass in a function / object.

IlyaMoiseev:
Signature will be `int getFinalCost(Edge edge)`. Initially it will simple calls `getCost` getter. In near future, after implementing maintenance and diverse features, it going to be sum of 4-5 cost numbers, what is arguable.

I think all needed attributes will be accessible from Edge instance. AvailableNetwork graph will be linked, and Edge instance may considered as graph entrypoint. But hope in strategy we will only look at source and dest edge switches, and don't needed to unwind it deeper.

PR #1824, opened 2018-12-24
Evolve PCE with weights computation strategies design

IlyaMoiseev:
Design for #1820

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/1824)


docs/design/solutions/pce-weights-strategies/pce-weights-strategies.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Goals
+For now PCE uses only cost field as graph edges weights
+
+While path computing logic going to be more complicated - we need to take several factors\costs to compute final edge\node weight, and keep this things maintainable.
+Also built AvailableNetwork should be logged, to make strange computed paths investigation possible.
+
+## Proposed changes
+- Introduce PCE data model instead of reusing DAO: for storing graph weighs by features and more convenient logging
+- Separate AvailableNetwork creation: it isn't responsibility of PathComputer
```
carmine: I think that the PathComputer can have an interface where it either uses the default AvailableNetwork or is given an AvailableNetwork. I believe the code is structured like that already. The benefit of using a default is that it doesn't require every PC client to call the same method to get available network; IOW, it makes sense to provide a simplified interface, and provide a more complex one for those that need it.

But, why do this as part of PCE diversity? What's the relationship, other than you're working on this code? It could be better to lodge this as a separate improvement proposal, to be discussed independent of diversity.

docs/design/solutions/pce-weights-strategies/pce-weights-strategies.md
```diff
@@ -0,0 +1,19 @@
+# Refactor PCE with weights computation strategies
+
+## Goals
```
carmine: I suggest starting with an example, and possibly a better one than below. The current set of Goals are okay at a very high level, but there isn't sufficient information on either the complete set of problems that need to be solved, or the set of solutions that would work.

carmine:
Based on conversation in skype, this feels more like anti-affinity - ie given 2 flows, either make them share zero ISLs, or have them share the fewest possible. I'll update #1820 accordingly.

docs/design/solutions/pce-weights-strategies/pce-weights-strategies.md
IlyaMoiseev: Ok, now I am also decline to keep single PCE interface and do all work inside.

Diversity is here, because this design was introduced in debates about how it should be implemented, and it going to be the first feature what will use it.
Aim is to have one computer class what will use strategies to calculate graph edges "weights" at runtime. The simplest strategy is "ISL cost getter", what is equal with the current PCE implementation.

docs/design/solutions/pce-weights-strategies/pce-weights-strategies.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Proposed changes
+- Introduce PCE data model instead of reusing DAO: for storing graph weighs by features and more convenient logging
+- Separate AvailableNetwork creation: it isn't responsibility of PathComputer
+- Introduce weights computing strategies: like weight getters above data model, defines what weigh features will be used in path computation and in that proportions
+
+### Weight strategy example
+Suppose that we have `sw1 -> isl -> sw2` graph edge in diversity scenario, then weight strategy may looks like:
+`sw1 diversity cost + isl cost + isl diversity cost + sw2 diversity cost`.
+
```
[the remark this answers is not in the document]
IlyaMoiseev: We doesn't needed entire set of paths, it's only a new flow feature for now, and it will be diversed only with one\several other flows.
Fore this case we are going to change "diversity flag" to some configurable big constant, gets it as a part of final edge weight with the proper strategy, and keep current algo working. If there is a path around used segment - it will be returned, if not - solution allows partial diversity, for cases like bottlenecks between graph parts.

docs/design/solutions/pce-weights-strategies/pce-weights-strategies.md
IlyaMoiseev: I see. Added some explanation about.

docs/design/solutions/pce-weights-strategies/pce-weights-strategies.md
```diff
@@ -0,0 +1,19 @@
+# Refactor PCE with weights computation strategies
+
+## Goals
+For now PCE uses only cost field as graph edges weights
+
+While path computing logic going to be more complicated - we need to take several factors\costs to compute final edge\node weight, and keep this things maintainable.
+Also built AvailableNetwork should be logged, to make strange computed paths investigation possible.
```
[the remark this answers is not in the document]
IlyaMoiseev: Logging excluded from scope of this PR to #1864

PR #1825, opened 2018-12-24
Reimplemented path finder: fixed equality of forward and reverse paths

nikitacherevko:
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/1825)


nikitacherevko:
Perhaps, we might leave path finder as is and take from this path only changes regarding loading only symmetric links from neo4j.

nikitacherevko:
Since there is no need to re-implement dijkstra algo I closed this PR and will create another one with only difference in the way how we load links from neo4j

Issue #1827, opened 2018-12-24
Port-antiflap leaves port down even after UP event


sergii-iakovenko:
Port-antiflap leaves a port down even if FL receives UP event as the last one.

Steps to reproduce:
- Start blinking a port. 
- The port goes into the cooling down state. Event topology receives DOWN event.
- During the cooling down state, stop blinking with UP as the last event. Hint: the previous DOWN event must comes to FL with the same timestamp in the log.
- Wait until the cooling down state is over for the port.

Expected: 
- Event topology receives UP event.

Actual:
- PortEventThrottlingService registers DOWN event as the last and doesn't forward UP to Event topology.

sergii-iakovenko:
Related to #1729

rtretyak:
Test reproducing the issue available here: #1835 
Please unignore when committing a fix

rtretyak:
Tested in #1835 
Fixed by #1738 #1705 and #1855

PR #1810, opened 2018-12-20
Move AvailabilityIssues.feature along with tools to functional-tests

rtretyak:
- Existing tests removed from ATDD module (AvailabilityIssues, Hearbeat)
- Similar test added to functional-tests
- Related tools moved from 'atdd' to 'test-library'

Closes #1805 
Closes #1799

PR #1835, opened 2018-12-26
#1827: Add test that blinks port super fast

rtretyak:
The PR should be repointed to develop after #1810 merge


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/Wrappers.groovy
```diff
@@ -34,6 +36,23 @@ class Wrappers {
...
+     * Executes given closure in a loop for certain amount of time.
+     * If time is less or equal to 0, the closure won't be executed even once.
+     * Loop may take a bit longer than 'timeout' by maximum of closure execution time, since check is performed
+     * before the closure call.
+     *
+     * @param timeout For how long to execute the given closure in a loop
+     * @param unit In which unit the 'timeout' value is specified
+     * @param closure code to repeat in a loop
+     */
+    static void timedLoop(long timeout, TimeUnit unit = TimeUnit.SECONDS, Closure closure) {
```
ylobankov: 1. Can we rename it to just 'loop' to have consistency with 'wait' and 'retry'? I don't insist, it is just a question.
2. I am not sure about timeout as long.  For example, I will want to execute my closure for 3.5 mins. I think we need to use double here.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/PortAntiflapSpec.groovy
```diff
@@ -105,4 +118,46 @@ timeout"() {
...
+        }
+        log.debug("Generated $messagesGenerated port up/down messages in ${antiflapWarmup + 1} seconds")
+
+        then: "Related ISL is FAILED"
+        Wrappers.wait(WAIT_OFFSET / 2.0) {
+            assert islUtils.getIslInfo(isl).get().state == IslChangeType.FAILED
+        }
+
+        and: "After the port cools down the ISL is discovered again"
+        TimeUnit.SECONDS.sleep(antiflapCooldown - 1)
```
ylobankov: Could you please explain why we need to subtract -1?

ylobankov: Also, why not move it to wait?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/Wrappers.groovy
rtretyak: TimeUnit only understands long. You'll have to pass 3500 milliseconds instead. Otherwise, if I change to double I'll have to remove the ability to select 'unit' and always require seconds. Do you prefer to change this way?

rtretyak: A self-made min->sec->millis converter for double values would be an overkill I think

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/PortAntiflapSpec.groovy
rtretyak: It doesn't make sense to even start 'waiting' and disturbing northbound before the cooldown runs out. Subtracting `-1` to start actually 'waiting' right before the cooldown ends

ylobankov: About -1. I thought so. Thanks.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/helpers/Wrappers.groovy
ylobankov: I don't think that we need to leave the possibility to choose a unit for this method. As far as I understand we can pass milliseconds, seconds, minutes, etc. But milliseconds unlikely will be used. Minutes can be used but only integer values. So my point is to use just seconds as in wait.

rtretyak: Changed similar to 'wait' to always accept seconds

PR #1844, opened 2018-12-28
Add test for perf loading Storm

rtretyak:
- also remove old mininet classes
- also add proper config for kafka naming strategy


services/src/functional-tests/kilda.properties.example
```diff
@@ -22,6 +22,7 @@ elasticsearch.endpoint = http://localhost:9200
 elasticsearch.username = username
 elasticsearch.password = password
 
+environment.naming.prefix=
```
ylobankov: Please, provide an example value for this property.

ylobankov: Ok. So I think we can leave a small comment that will explain that for virtual env it should be empty, for staging it should be "staging".

rtretyak: all staging info is 'internal' for us, so should not be pushed to public

PR #1848, opened 2018-12-31
Delete Bitbucket files.

Settis:


Settis:
It was made for test. It's not needed any more.

PR #1872, opened 2019-01-14
Evolve PCE with weights computation strategies

### LINKED ISSUE #1820: Evolve PCE with weights computation strategies

IlyaMoiseev:
Closes #1820


rtretyak:
Are there any functional changes that need to be additionally tested?

IlyaMoiseev:
No. This patch is refactoring only, and shouldn't affect any functionality

PR #1879, opened 2019-01-14
Fixed 'LinkSpec.Cannot delete nonexistent link' functional test

niksv:
Root cause of test fail: error message was changed


niksv:
Invalid

PR #1899, opened 2019-01-17
Fix functional test for logging verification

ylobankov:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/logging/CheckLoggingSpec.groovy
```diff
@@ -45,8 +45,8 @@ class CheckLoggingSpec extends BaseSpecification {
         def switchId = topology.activeSwitches.first().dpId
         northbound.validateSwitchRules(switchId)
 
-        then: "Northbound, Storm and Topology Engine should log these actions within 1 minute"
-        Wrappers.wait(timeout, 10) {
```
npliashechnikov: I'd rather not reduce the retry intervals, as spamming ElasticSearch with queries might actually slow the test down.

ylobankov: Fixed.

PR #1915, opened 2019-01-22
Add better logging of skipped tests

### LINKED ISSUE #1902: Add extension that will report the list of skipped tests
As a QA engineer, I want my framework to report the list of skipped tests at the end of the test run, so that I'm aware of which tests were skipped and for what reason. This is required due to large amount of 'assumptions' we use in tests, which may lead to certain tests being skipped. Engineer who runs tests should have means of inspecting the list of skipped tests and check whether all skips are 'legal', as well as better understand which places are left uncovered.
AC:
1. List of test names for skipped test is logged in separate place or printed out at the end of test run

rtretyak:
Closes #1902


rtretyak:
No suggestions from Spock, going with alt solution of creating a separate log file for skipped tests

PR #1917, opened 2019-01-22
Add test to verify resetting traffic counters for ingress rule on flow reroute

### LINKED ISSUE #1877: Test that counters are reset on flow reroute
See details under #1198

ylobankov:
Closes #1877.

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
@@ -9,13 +9,20 @@ import static org.openkilda.testing.Constants.WAIT_OFFSET
 import static spock.util.matcher.HamcrestSupport.expect
 
 import org.openkilda.functionaltests.BaseSpecification
+import org.openkilda.functionaltests.helpers.PathHelper
 import org.openkilda.functionaltests.helpers.Wrappers
 import org.openkilda.messaging.command.switches.DeleteRulesAction
 import org.openkilda.messaging.command.switches.InstallRulesAction
+import org.openkilda.messaging.info.event.IslChangeType
+import org.openkilda.messaging.info.event.PathNode
 import org.openkilda.messaging.info.rule.FlowEntry
+import org.openkilda.messaging.payload.flow.FlowEndpointPayload
 import org.openkilda.messaging.payload.flow.FlowPayload
+import org.openkilda.messaging.payload.flow.FlowState
 import org.openkilda.model.SwitchId
+import org.openkilda.northbound.dto.flows.PingInput
 import org.openkilda.testing.Constants.DefaultRule
+import org.openkilda.testing.model.topology.TopologyDefinition.Isl
 import org.openkilda.testing.model.topology.TopologyDefinition.Switch
 
 import spock.lang.Narrative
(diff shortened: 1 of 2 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
```diff
@@ -421,6 +427,61 @@ class SwitchRulesSpec extends BaseSpecification {
         flowHelper.deleteFlow(flow.id)
     }
 
+    def "Traffic counters for ingress rules are reset on flow rerouting"() {
+        given: "Two active switches and two possible flow paths at least"
+        def switches = topology.getActiveSwitches()
```
npliashechnikov: I think we really need to have some kind of library which would handle these flow route selection situations for us. Because this same code is used in a lot of test cases and if something changes in how the Kilda works - it'd be a lot to fix.
Something like def findPaths(int minDistance = 1, int minAlternatePaths, int maxAlternatePaths) would be great.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
```diff
@@ -457,4 +518,20 @@ class SwitchRulesSpec extends BaseSpecification {
 
         return rules
     }
+
+    void checkTrafficCountersForRules(Switch sw, FlowEndpointPayload flowEndpoint) {
```
npliashechnikov: I think it'd be nice to add boolean parameter "isTrafficExpected". This way you can also use this function on line 472 instead of duplicating the code.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
```diff
@@ -421,6 +427,61 @@ class SwitchRulesSpec extends BaseSpecification {
...
+        [(srcSwitch): flow.source, (dstSwitch): flow.destination].each {
+            def rules = northbound.getSwitchRules(it.key.dpId).flowEntries
+            def ingressRule = filterRules(rules, it.value.portNumber, it.value.vlanId, null)[0]
+
+            assert ingressRule.packetCount == 0
+            assert ingressRule.byteCount == 0
+        }
+
+        and: "Revive the ISL back (bring switch port up) and delete the flow"
+        northbound.portUp(islToFail.srcSwitch.dpId, islToFail.srcPort)
```
npliashechnikov: I don't know if it is safe. If the previous verification fails for some reason, the environment will end up with damaged topology. Maybe it'd be better to perform port up right after the successful reroute?

ylobankov: Your variant is safer but still, we may get a damaged topology if we fail inside Wrappers.wait(). Ideally, it should be in cleanup but we use such an approach everywhere since we have the "no cleanup" convection.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
```diff
@@ -457,4 +513,20 @@ class SwitchRulesSpec extends BaseSpecification {
...
+
+    void checkTrafficCountersInRules(FlowEndpointPayload flowEndpoint, trafficThroughIngressRule = true) {
+        def rules = northbound.getSwitchRules(flowEndpoint.datapath).flowEntries
+        def ingressRule = filterRules(rules, flowEndpoint.portNumber, flowEndpoint.vlanId, null)[0]
+        def egressRule = filterRules(rules, null, null, flowEndpoint.portNumber).find {
+            it.instructions.applyActions.fieldAction.fieldValue == flowEndpoint.vlanId.toString()
+        }
+
+        assert ingressRule.flags.contains("RESET_COUNTS")
+        assert (trafficThroughIngressRule ? ingressRule.packetCount > 0 : ingressRule.packetCount == 0)
```
npliashechnikov: It can actually be simplified to
assert trafficThroughIngressRule == (ingressRule.packetCount > 0)

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
```diff
@@ -421,6 +427,56 @@ class SwitchRulesSpec extends BaseSpecification {
...
+            assert northbound.getFlowStatus(flow.id).status == FlowState.UP
+            assert PathHelper.convert(northbound.getFlowPath(flow.id)) != flowPath
+        }
+
+        and: "Traffic counters in ingress rule on source and destination switches are reset"
+        checkTrafficCountersInRules(flow.source, false)
+        checkTrafficCountersInRules(flow.destination, false)
+
+        and: "Revive the ISL back (bring switch port up) and delete the flow"
+        northbound.portUp(islToFail.srcSwitch.dpId, islToFail.srcPort)
```
rtretyak: You should wait for ISL to become discovered

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
```diff
@@ -457,4 +513,20 @@ class SwitchRulesSpec extends BaseSpecification {
 
         return rules
     }
+
+    void checkTrafficCountersInRules(FlowEndpointPayload flowEndpoint, trafficThroughIngressRule = true) {
```
rtretyak: This is personal, but I find that the `trafficThroughIngressRule` should not be optional, since method signature without this param looks unintuitive

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
ylobankov: Thanks, fixed.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
ylobankov: Thanks, fixed.

Issue #1765, opened 2018-12-13
A flow may be created with asymmetric paths under certain conditions in a certain topology


ylobankov:
**Environment:**
Virtual topology like 

A == B == C

where "=" represents two parallel links from one switch to another. So we have the following flow paths:

A -- B -- C
A _ B _ C
A -- B _ C
A _ B -- C

**Steps to reproduce:**
A: 
1. Make cost for the **forward** path A -- B -- C very high. It can be made via DB, for example.
2. Make cost for the **reverse** path  A _ B _ C  very high as well.
3. Try to create a flow from A to C.

B:
1. Make all cost for both links the same.
2. Try to create a flow from A to C.

**Expected result:**
Forward and backward paths are symmetric.

**Actual result:**
Forward and backward paths are asymmetric.
A _ B _ C - forward path
A -- B -- D - reverse path

PR #1926, opened 2019-01-24
Fix case with asymmetric reverse path for flow

Levitan:
Closes: #1765

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/1926)


Levitan:
> Add tests that confirms the issue is fixed, add meaningful commit message

@timofei-durakov test was added

timofei-durakov:
still lgtm

PR #1927, opened 2019-01-24
Do not sent port events for bdf ports

niksv:


confd/vars/main.yaml
```diff
@@ -95,6 +95,8 @@ kilda_port_up_down_throttling_delay_seconds_min: 1
 kilda_port_up_down_throttling_delay_seconds_warm_up: 5
 kilda_port_up_down_throttling_delay_seconds_cool_down: 5
 
+kilda_bfd_port_offset: 100
```
timofei-durakov: replace to 200

services/wfm/src/main/java/org/openkilda/wfm/topology/event/OFEventWfmTopologyConfig.java
```diff
@@ -55,6 +56,10 @@ default String getKafkaTopoRerouteTopic() {
     @Key("port.up.down.throttling.delay.seconds.cool.down")
     int getPortUpDownThrottlingDelaySecondsCoolDown();
 
+    @Key("bfd.port.offset")
+    @Default("100")
```
timofei-durakov: the same

PR #1946, opened 2019-01-29
Add tests to verify rules installation on switches

### LINKED ISSUE #1941: New tests: cover install rules API
Add new tests for install rules endpoint `PUT /api/v1/switches/$swId/rules`.
Cover all possible InstallRulesAction-s

ylobankov:
Closes #1941.

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchRulesSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchRulesSpec.groovy
@@ -22,7 +22,6 @@ import org.openkilda.messaging.payload.flow.FlowEndpointPayload
 import org.openkilda.messaging.payload.flow.FlowPayload
 import org.openkilda.messaging.payload.flow.FlowState
 import org.openkilda.model.Cookie
-import org.openkilda.model.SwitchId
 import org.openkilda.northbound.dto.flows.PingInput
 import org.openkilda.testing.model.topology.TopologyDefinition.Switch
 
@@ -52,7 +51,7 @@ class SwitchRulesSpec extends BaseSpecification {
 
     @Unroll("Default rules are installed on an #sw.ofVersion switch(#sw.dpId)")
     def "Default rules are installed on switches"() {
-        expect: "Default rules are installed on the #sw.ofVersion switch"
+        expect: "Default rules are installed on the switch"
         def cookies = northbound.getSwitchRules(sw.dpId).flowEntries*.cookie
         cookies.sort() == sw.defaultCookies.sort()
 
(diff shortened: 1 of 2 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
```diff
@@ -519,9 +577,11 @@ class SwitchRulesSpec extends BaseSpecification {
         flowHelper.randomFlow(srcSwitch, dstSwitch)
     }
 
-    List<FlowEntry> getFlowRules(SwitchId switchId) {
+    List<FlowEntry> getRules(SwitchId switchId, defaultRules = true) {
```
rtretyak: This will be conflicting with my patch about new bfd rule and should be resolved to my advantage, since default rules are different across actons, centecs and noviflows

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
```diff
@@ -114,6 +114,64 @@ class SwitchRulesSpec extends BaseSpecification {
...
+                         }
+                 ],
+                 [
+                         installRulesAction: InstallRulesAction.INSTALL_DEFAULTS,
+                         rulesInstalled    : srcSwDefaultRules.size(),
+                         getExpectedRules  : { switchId -> getRules(switchId) }
+                 ]
+        ]
+    }
+
```
rtretyak: Consider scenario:
1. switch with both default and flow rules
2. drop rules
3. call install rules (without flag)
expect: both default and flow rules are installed

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
ylobankov: Fixed.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/northbound/switches/SwitchRulesSpec.groovy
ylobankov: For now, it doesn't work. Only default rules are installed.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchRulesSpec.groovy
```diff
@@ -114,6 +109,86 @@ class SwitchRulesSpec extends BaseSpecification {
...
+                                expectedRules     : { rules ->
+                                    rules.findAll {
+                                        it.cookie == Cookie.CATCH_BFD_RULE_COOKIE
+                                    }
+                                }
+                        ],
+                        [
+                                installRulesAction: InstallRulesAction.INSTALL_DEFAULTS,
+                                rulesInstalled    : { rulesCount -> rulesCount },
+                                expectedRules     : { rules -> rules }
```
rtretyak: This iteration brings a lot of complication to the test. If you exclude this iteration you could drop both closures (expectedRules would be just a Cookie.SOMETHING, rulesInstalled could be removed completely). This will simplify the 'where' block and the test code overall. Plus when testing the `INSTALL_DEFAULTS` you use the same `INSTALL_DEFAULTS` in the cleanup which is excess. I think it's better to duplicate some simple code and have a separate test for `INSTALL_DEFAULTS` over making a single test which is harder to understand.
Though it can still work as it is, I don't insist

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchRulesSpec.groovy
```diff
@@ -177,53 +252,50 @@ class SwitchRulesSpec extends BaseSpecification {
...
-    def "Able to delete default rule from a switch (delete-action=#data.deleteRulesAction)"() {
-        assumeFalse("Unable to run the test because an OF_ 12 switch has one broadcast rule as the default",
-                dstSwitch.ofVersion == "OF_12" && data.cookie != Cookie.VERIFICATION_BROADCAST_RULE_COOKIE)
-
-        given: "A switch with some flow rules installed"
-        def flow = flowHelper.randomFlow(srcSwitch, dstSwitch)
-        flowHelper.addFlow(flow)
+    def "Able to delete default rule from an #sw.ofVersion switch (#sw.dpId, delete-action=#data.deleteRulesAction)"() {
+        assumeFalse("Unable to run the test because an OF_12 switch has one broadcast rule as the default",
+                sw.ofVersion == "OF_12" && data.cookie != Cookie.VERIFICATION_BROADCAST_RULE_COOKIE)
```
rtretyak: I think this is fine and it's the way I do it too, but I just thought that it is more correct to filter these 'iterations' directly in data provider and use assumptions only for 'runtime' things that we cannot predict (like topology unsuitable for the test). Just thoughts for the future

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchRulesSpec.groovy
ylobankov: Fixed, thanks.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchRulesSpec.groovy
ylobankov: Yes, probably in the future we will fix it.

Issue #1231, opened 2018-08-30
Associate flows and make them diverse (ie Anti-Affinity)


pzakatov:
example: 
Customer wants to have reduncancy on the 2 flows so they reside on different switches and/or ISLs

amount of associated flows could be more than 2

Key Use Cases:

1. Give me a path B that has no overlap with path A.
2. Give me a path B that has minimal overlap with path A (ie zero overlap preferred, followed by fewest intersection)
3. Allow grooming to take place in order to satisfy use case 1 or 2 (ie what if we moved some paths around, could that allow use case 1, or have a better story for 2)

shusyaM:
- related flows (need to define how the relation would be calculated) should reside on different inital and reroute paths: isl and switches

timofei-durakov:
- [ ] define model updates to build flow groups
- [ ] define API changes
- [ ] update domain model objects and DTOs, including converters
- [ ] introduce changes in WFM Flow topology(create, update, reroute)

carmine:
I've updated the Description with the following:

Key Use Cases:

1. Give me a path B that has no overlap with path A.
2. Give me a path B that has minimal overlap with path A (ie zero overlap preferred, followed by fewest intersection)
3. Allow grooming to take place in order to satisfy use case 1 or 2 (ie what if we moved some paths around, could that allow use case 1, or have a better story for 2)

pzakatov:
Path response example for if a flow marked as diverse
```
{
  "flowid": "flow",
  "flowpath_forward": [
    {"input_port": 12,"output_port": 1,"switch_id": "00:00:00:00:00:00:01:02"},
    {"input_port": 2,"output_port": 1,"switch_id": "00:00:00:00:00:00:01:01"}
  ],
  "flowpath_reverse": [
    {"input_port": 1,"output_port": 2,"switch_id": "00:00:00:00:00:00:01:01"},
    {"input_port": 1,"output_port": 12,"switch_id": "00:00:00:00:00:00:01:02"
    }
  ],
  "diverse_group": {
    "overlapping_segments": {
      "isl_count": 0,
      "switch_count": 2,
      "isl_percent": 0,
      "switch_percent": 100
    },
    "other_flows": [
      {
        "flowid": "diverse1",
        "flowpath_forward": [
          {"input_port": 13,"output_port": 5,"switch_id": "00:00:00:00:00:00:01:02"},
[...]

PR #1947, opened 2019-01-29
Return paths in flow group for diversity flow

IlyaMoiseev:
#1231 Return paths in flow group

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/1947)


services/wfm/src/main/java/org/openkilda/wfm/topology/flow/service/IntersectionComputer.java
```diff
@@ -0,0 +1,132 @@
...
+
+        int edgesOverlap = Sets.intersection(edges, mainFlowEdges).size();
+        int switchesOverlap = Sets.intersection(switches, mainFlowSwitches).size();
+        return new OverlappingSegmentsStats(edgesOverlap,
+                switchesOverlap,
+                percent(edgesOverlap, mainFlowEdges.size()),
+                percent(switchesOverlap, mainFlowSwitches.size()));
+    }
+
+    private int percent(int n, int from) {
```
timofei-durakov: what if it's zero?

services/wfm/src/main/java/org/openkilda/wfm/topology/flow/service/IntersectionComputer.java
```diff
@@ -0,0 +1,132 @@
...
+        return (int) ((n * 100.0f) / from);
+    }
+
+    /**
+     * Edge representation. Constraint: srcSwitch is always not less than destSwitch
+     */
+    @Value
+    static class Edge {
```
timofei-durakov: do we need nested class here?

services/wfm/src/main/java/org/openkilda/wfm/topology/flow/service/IntersectionComputer.java
IlyaMoiseev: For current logic from couldn't be zero. But if it is - floating point division will give NaN, and external cast to int gives zero. Added a test for this situation

services/wfm/src/main/java/org/openkilda/wfm/topology/flow/service/IntersectionComputer.java
IlyaMoiseev: This is highly specialized class for concrete purpose. And, as we already have several classes with that name, I preferred keep it as much restricted as possible.

IlyaMoiseev:
@nikitamarchenko done

PR #1960, opened 2019-01-30
Bfd api

timofei-durakov:


services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -141,8 +145,10 @@
 
     public static final int VERIFICATION_RULE_PRIORITY = FlowModUtils.PRIORITY_MAX - 1000;
     public static final int DROP_VERIFICATION_LOOP_RULE_PRIORITY = VERIFICATION_RULE_PRIORITY + 1;
+    public static final int CATCH_BFD_RULE_PRIORITY = DROP_VERIFICATION_LOOP_RULE_PRIORITY + 1;
     public static final int FLOW_PRIORITY = FlowModUtils.PRIORITY_HIGH;
     public static final long MAX_CENTEC_SWITCH_BURST_SIZE = 32000L;
+    public static final int BDF_DEFAULT_PORT = 3784;
```
sergii-iakovenko: Why don't use SwitchManagerConfig instead of hard-coding?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/web/SwitchManagerWebRoutable.java
```diff
@@ -33,7 +33,7 @@ public Restlet getRestlet(Context context) {
         router.attach("/flow", FlowResource.class);
         router.attach("/flows/switch_id/{switch_id}", FlowsResource.class);
         router.attach("/meters/switch_id/{switch_id}", MetersResource.class);
-
+        router.attach("/switch/enable_bfd", EnableBfdResource.class);
```
sergii-iakovenko: What if later we need to disable bfd? "/switch/{switch_id}/bfd" ?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
surabujin: It is actually constant switch will not accept any other value.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/web/SwitchManagerWebRoutable.java
surabujin: We have "complete" kafka API. This REST API is some set of back door required for initial system setup. I hope it will be dropped soon.

PR #1965, opened 2019-01-31
mask port while network dump for bfd

timofei-durakov:
this parth do bfd port masking for logical ports


services/src/messaging/src/main/java/org/openkilda/messaging/model/Switch.java
```diff
@@ -53,7 +52,7 @@ public Switch(
         this.datapath = datapath;
         this.ipAddress = ipAddress;
         this.features = ImmutableSet.copyOf(features);
-        this.ports = ImmutableList.copyOf(ports);
+        this.ports = ports;
```
sergii-iakovenko: Switch is a value object and must be immutable. Please use copy builder to clone and modify it.

PR #1979, opened 2019-02-01
Added catch for stats exceptions

niksv:
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/1979)

### DIFF
--- a/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/statistics/StatisticsService.java
+++ b/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/statistics/StatisticsService.java
@@ -108,9 +108,26 @@ public class StatisticsService implements IStatisticsService, IFloodlightModule 
         if (interval > 0) {
             threadPoolService.getScheduledExecutor().scheduleAtFixedRate(
                     () -> switchService.getAllSwitchMap().values().forEach(iofSwitch -> {
-                        gatherPortStats(iofSwitch);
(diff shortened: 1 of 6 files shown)


services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/statistics/StatisticsService.java
```diff
@@ -108,9 +108,23 @@ public void startUp(FloodlightModuleContext context) {
         if (interval > 0) {
             threadPoolService.getScheduledExecutor().scheduleAtFixedRate(
                     () -> switchService.getAllSwitchMap().values().forEach(iofSwitch -> {
-                        gatherPortStats(iofSwitch);
-                        gatherFlowStats(iofSwitch);
-                        gatherMeterStats(iofSwitch);
+                        try {
+                            gatherPortStats(iofSwitch);
+                        } catch (Exception e) {
+                            logger.error("Failed to gather stats for ports on switch {}.", iofSwitch.getId(), e);
```
surabujin: 2 issue - first it will not provide traceback (you can't use "{}" formatting and pass exception object, you must use String.format to format message if you pass exception object and expect traceback in log).

And second - make different log messages for different sections.

niksv: 1. will fix.
2. messages are different

Issue #1157, opened 2018-08-17
Custom logger for logs related to default kibana dashboards


surabujin:
Kibana dashboard use set of log messages to trace notable events in `system`. There is at least 2 problem here:
1. Developer can alter/remove log message used by kibana
1. Log level can be changed and message will be filtered out

In both cases monitoring system become "blind" for some set of events. And even worst - it stays blind until detailed investigation of some failure.

The proposition is to add some "maker" in log message. For example - "{PORT_UP} port event - switch 00:00:00:00:00:01 port 1 become UP". Portion "{PORT_UP}" is a mentioned marker.

This marker make obvious to developer that this message is used in kibana. Also it will simplify and possible speedup search patterns in kibana.

As one more benefit - message body itself can be altered without danger to corrupt search pattern in kibana.

pzakatov:
@surabujin , Some other options:
1. would it be better if we add a comment like "// Used in the default Kibana setup"? To be honest, the marker is not so obvious...
2. we can add a specific log field like "dashboard" with a specific ID (comma separated if multiple dashboards used), so we can clearly see which dashboard it relates to...
3. we can add tests so we make sure that logs are exists and in the expected format

make sense?

sergii-iakovenko:
I'd like to propose another option: implement a custom logger that will be used by Kilda code to log events. E.g. similar to https://www.slf4j.org/apidocs/org/slf4j/ext/XLogger.html
The logger will expose an interface for event logging without an ability to change the log level, format, etc.

We also can introduce an aspect to ensure that the code invokes the logger proper number of times (entry, exit, error/exception).

surabujin:
@sergii-iakovenko I don't like aspects(actualy I don't like amount of magic required to use them in java and amount of problem that bring their usage), but I like rest of your proposition. I.e. we can implement "util" to make log records used by kibana dashboards.

PS Even log messages produced by such unit, should include some obvious match key, used by kibana, to simplify search queries.

surabujin:
The decision was made - use "log wrapper/adapter" proposed by @sergii-iakovenko. With one addition - this wrapper/addapter must inject "dashboard related data" in extra fields using MDC (for ISL up it can add field "event: ISL-UP").

Implementation plan:
* decide the scale of this log wrapper (one per project, one per component, one per topology), implement POC.
* fill created log wrappers with all used now in dashboards messages #692.
* recreate existing kibana dashboards using new/extended log fields for match log records.

[...]

PR #1980, opened 2019-02-01
Improve KibanaLogWrapper

Levitan:
- Using logger from class where used log wrapper
- Add ability for choose log level

Impove issue #1157

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/utils/KibanaLogWrapper.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/utils/KibanaLogWrapper.java
@@ -19,8 +19,10 @@ import org.openkilda.messaging.info.event.IslChangeType;
 import org.openkilda.messaging.info.event.SwitchChangeType;
 import org.openkilda.model.SwitchId;
 
-import lombok.extern.slf4j.Slf4j;
+import lombok.Value;
+import org.slf4j.Logger;
 import org.slf4j.MDC;
+import org.slf4j.event.Level;
 
 import java.util.HashMap;
 import java.util.Map;
@@ -28,7 +30,7 @@ import java.util.Map;
 /**
  * The KIBANA dashboard log wrapper.
  */
-@Slf4j
+@Value
 public class KibanaLogWrapper {
 
     private static final String SWITCH_ID = "switch_id";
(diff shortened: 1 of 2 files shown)


services/wfm/src/main/java/org/openkilda/wfm/topology/utils/KibanaLogWrapper.java
```diff
@@ -43,18 +45,30 @@
...
+    private final Logger logger;
+
+    /**
+     * Constructs a Kibana log wrapper.
+     *
+     * @param logger an instance of logger.
+     */
+    public KibanaLogWrapper(Logger logger) {
+        this.logger = logger;
+    }
```
timofei-durakov: it would be easier to use @Value here

services/wfm/src/main/java/org/openkilda/wfm/topology/utils/KibanaLogWrapper.java
```diff
@@ -115,11 +131,26 @@ public void onIslDiscovery(SwitchId srcSwitch, int srcPort, int dstPort, IslChan
...
+                    logger.warn(message);
+                    break;
+                case ERROR:
+                    logger.error(message);
+                    break;
+                case DEBUG:
+                    logger.debug(message);
+                    break;
+                default:
+                    logger.trace(message);
```
timofei-durakov: why the default is trace?

services/wfm/src/main/java/org/openkilda/wfm/topology/utils/KibanaLogWrapper.java
timofei-durakov: also it looks like @Slf4j could be removed

services/wfm/src/main/java/org/openkilda/wfm/topology/utils/KibanaLogWrapper.java
```diff
@@ -43,83 +45,115 @@
...
-        proceed("State changed", data);
+        proceed(level, message, data);
     }
 
     /**
      * Build and write log message and MDC custom fields.
      *
+     * @param level a log level.
+     * @param message a message text.
      * @param message a message text.
```
niksv: double message param

PR #1996, opened 2019-02-05
Feature discovery for new Noviflow switches

IlyaMoiseev:


services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/feature/AbstractFeature.java
```diff
@@ -17,13 +17,16 @@
...
+import com.google.common.collect.Sets;
 import net.floodlightcontroller.core.IOFSwitch;
 
 import java.util.Optional;
+import java.util.Set;
 
 public abstract class AbstractFeature {
     protected static final String MANUFACTURER_NICIRA = "Nicira, Inc.";
-    protected static final String MANUFACTURER_NOVIFLOW = "NoviFlow Inc";
+    protected static final Set<String> MANUFACTURER_DESCRIPTIONS_NOVIFLOW =
```
surabujin: `_DESCRIPTIONS_` is extra part here.

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/feature/BfdFeature.java
```diff
@@ -31,7 +31,7 @@
         if (description == null) {
             return empty;
         }
-        if (!MANUFACTURER_NOVIFLOW.equals(description.getManufacturerDescription())) {
+        if (!MANUFACTURER_DESCRIPTIONS_NOVIFLOW.contains(description.getManufacturerDescription())) {
```
surabujin: And where is NW440 regexp?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/feature/AbstractFeature.java
IlyaMoiseev: Keep variable name as it was

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/feature/BfdFeature.java
IlyaMoiseev: Changed to regexp above "software description" field

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/feature/AbstractFeature.java
```diff
@@ -20,10 +20,11 @@
 import net.floodlightcontroller.core.IOFSwitch;
 
 import java.util.Optional;
+import java.util.regex.Pattern;
 
 public abstract class AbstractFeature {
     protected static final String MANUFACTURER_NICIRA = "Nicira, Inc.";
-    protected static final String MANUFACTURER_NOVIFLOW = "NoviFlow Inc";
+    protected static final Pattern MANUFACTURER_NOVIFLOW = Pattern.compile("^NW\\d{3}\\.\\d+\\.\\d+$");
```
niksv: I think we need unit tests for this regexp

surabujin: This is definetelly not "MANUFACTURER_NOVIFLOW". Keep this regexp on BfdFeature "level" and drop this constant if it is not used any more.

PR #1997, opened 2019-02-05
drop port event messages for logical ports

timofei-durakov:
if port state changed to something not equal to UP or DOWN
drop port messages and do not process them

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/event/OfeLinkBolt.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/event/OfeLinkBolt.java
@@ -424,9 +424,13 @@ public class OfeLinkBolt
             handleSwitchEvent(tuple, cleanedInfoMessage);
             passToNetworkTopologyBolt(tuple, infoMessage);
         } else if (data instanceof PortInfoData) {
-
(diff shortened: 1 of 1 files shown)


services/wfm/src/main/java/org/openkilda/wfm/topology/event/OfeLinkBolt.java
```diff
@@ -424,9 +424,13 @@ protected void dispatchMain(Tuple tuple, InfoMessage infoMessage) {
             handleSwitchEvent(tuple, cleanedInfoMessage);
             passToNetworkTopologyBolt(tuple, infoMessage);
         } else if (data instanceof PortInfoData) {
-
-            int portNo = ((PortInfoData) data).getPortNo();
+            PortInfoData portInfoData = (PortInfoData) data;
+            int portNo = portInfoData.getPortNo();
             if (portNo > bfdPortOffset) {
+                PortChangeType state = portInfoData.getState();
+                if (state != PortChangeType.UP && state != PortChangeType.DOWN) {
```
surabujin: Usually I like "filter like `if` usage", but in this case I would like to see "positive match"... it should be more natural here?

PR #2007, opened 2019-02-07
Fix functional tests for hardware env

ylobankov:


services/src/test-library/src/main/java/org/openkilda/testing/Constants.java
```diff
@@ -20,7 +20,7 @@
 
 public final class Constants {
     public static final Integer DEFAULT_COST = 700;
-    public static final Integer WAIT_OFFSET = 10;
+    public static final Integer WAIT_OFFSET = 20;
```
rtretyak: Why this should be done?

rtretyak: I think we trying to mask some other problems by doing this. This was supposed to narrow any network delays + internal computation delays and 10 seconds is already a ridiculous timeout. We should locate which test exactly is having problems and pin-point the problem

ylobankov: Sorry, I made this change by a mistake. I was playing locally with some stuff. It should not be touched.

PR #2021, opened 2019-02-08
Add usage minimal port speed for ISL when have different port speed

### LINKED ISSUE #2001: Use minimal port speed value for an ISL capacity (for the case when we have different port speed on ISL endpoints)
Sometimes an ISL have different port speed on its edges. In that case, we need to set ISL capacity and all bandwidth parameters according to minimal speed value. Eg. 10G on one side, and 1G on another side, the ISL should have a 1G capacity.

Levitan:
Closes #2001

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2021)

### DIFF
--- a/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/pathverification/PathVerificationService.java
+++ b/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/pathverification/PathVerificationService.java
@@ -488,7 +488,8 @@ public class PathVerificationService implements IFloodlightModule, IPathVerifica
             PathNode source = new PathNode(new SwitchId(remoteSwitch.getId().getLong()), remotePort.getPortNumber(), 0,
                             latency);
             PathNode destination = new PathNode(new SwitchId(input.getDpId().getLong()), inPort.getPortNumber(), 1);
-            long speed = getSwitchPortSpeed(input.getDpId(), inPort);
+            long speed = Math.min(getSwitchPortSpeed(input.getDpId(), inPort),
+                    getSwitchPortSpeed(remoteSwitch.getId(), remotePort));
             IslInfoData path = IslInfoData.builder()
                     .latency(latency)
                     .source(source)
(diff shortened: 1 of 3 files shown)


services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/pathverification/PathVerificationService.java
```diff
@@ -488,7 +488,8 @@ void handlePacketIn(OfInput input) {
             PathNode source = new PathNode(new SwitchId(remoteSwitch.getId().getLong()), remotePort.getPortNumber(), 0,
                             latency);
             PathNode destination = new PathNode(new SwitchId(input.getDpId().getLong()), inPort.getPortNumber(), 1);
-            long speed = getSwitchPortSpeed(input.getDpId(), inPort);
+            long speed = Math.min(getSwitchPortSpeed(input.getDpId(), inPort),
+                    getSwitchPortSpeed(remoteSwitch.getId(), remotePort));
             IslInfoData path = new IslInfoData(latency, source, destination, speed, IslChangeType.DISCOVERED,
```
niksv: can we create unit tests for that?

services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/pathverification/PathVerificationService.java
```diff
@@ -488,7 +488,8 @@ void handlePacketIn(OfInput input) {
             PathNode source = new PathNode(new SwitchId(remoteSwitch.getId().getLong()), remotePort.getPortNumber(), 0,
                             latency);
             PathNode destination = new PathNode(new SwitchId(input.getDpId().getLong()), inPort.getPortNumber(), 1);
-            long speed = getSwitchPortSpeed(input.getDpId(), inPort);
+            long speed = Math.min(getSwitchPortSpeed(input.getDpId(), inPort),
+                    getSwitchPortSpeed(remoteSwitch.getId(), remotePort));
```
niksv: can we create unit tests for that?

Levitan: I think no.

niksv: I think yes)

PR #2031, opened 2019-02-11
Refactor and fix functional tests for verifying switches deletion

ylobankov:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchDeleteSpec.groovy
```diff
@@ -1,281 +1,194 @@
...
+        // restore switch
+        lockKeeper.knockoutSwitch(sw.dpId)
+        lockKeeper.reviveSwitch(sw.dpId)
+        Wrappers.wait(WAIT_OFFSET) { assert northbound.activeSwitches.any { it.switchId == sw.dpId } }
 
-            actualIsl.size() == 1 && actualIsl.every { it.state == IslChangeType.DISCOVERED }
+        // restore ISLs
+        def allSwIsls = swIsls.collectMany { [it, islUtils.reverseIsl(it)] }
+        allSwIsls.each { northbound.portDown(it.srcSwitch.dpId, it.srcPort) }
+        TimeUnit.SECONDS.sleep(antiflapCooldown + 1)
```
rtretyak: I think this is an excess 1s sleep. bare `antiflapCooldown` will do just fine even if there is 1 isl(the api call itself takes additional ~0.3-0.5s), but especially given there are N isls and you iterate through them twice, so the portUp for isl1 will be called only after portDowns for all isls are called

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchDeleteSpec.groovy
```diff
@@ -1,281 +1,194 @@
...
+        Wrappers.wait(WAIT_OFFSET) { assert northbound.activeSwitches.any { it.switchId == flow.source.datapath } }
+        flowHelper.deleteFlow(flow.id)
+
+        where:
+        [flowType, flow] << [["single-switch", getFlowHelper().singleSwitchFlow(getTopology().getActiveSwitches()[0])],
+                             ["casual", getFlowHelper().randomFlow(*getTopology().getActiveSwitches()[0..1])]]
     }
 
-    def "Can delete inactive switch without any ISLs"() {
+    @Ignore("Unstable. Under investigation.")
```
rtretyak: #2038

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchDeleteSpec.groovy
```diff
@@ -1,281 +1,194 @@
...
-        northbound.deleteSwitch(srcSwitch.getDpId(), false)
+        and: "Try to delete the switch"
+        northbound.deleteSwitch(flow.source.datapath, false)
 
-        then: "Got 400 BadRequest because switch has flow"
+        then: "Got 400 BadRequest error because the switch has the flow assigned"
         def exc = thrown(HttpClientErrorException)
         exc.rawStatusCode == 400
-        exc.responseBodyAsString.matches(".*Switch '$srcSwitch.dpId' has \\d+ assigned flows.*")
+        exc.responseBodyAsString.matches(".*Switch '${flow.source.datapath}' has \\d+ assigned flows.*")
```
rtretyak: can we go the common way of deserializing the message into `MessageError` class? Also, the `\\d+` should be replaced with exact value that we expect here (which is `1`)

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchDeleteSpec.groovy
```diff
@@ -1,281 +1,194 @@
...
-        }
-        northbound.deleteFlow(flow.id)
+        and: "Cleanup: activate the switch back and remove the flow"
+        lockKeeper.reviveSwitch(flow.source.datapath)
+        Wrappers.wait(WAIT_OFFSET) { assert northbound.activeSwitches.any { it.switchId == flow.source.datapath } }
+        flowHelper.deleteFlow(flow.id)
+
+        where:
+        [flowType, flow] << [["single-switch", getFlowHelper().singleSwitchFlow(getTopology().getActiveSwitches()[0])],
+                             ["casual", getFlowHelper().randomFlow(*getTopology().getActiveSwitches()[0..1])]]
```
rtretyak: Looking at this, I feel like it makes sense to try to delete an intermediate switch, which is not explicitly mentioned as source or destination of certain flow, but is used in path. It looks like a separate valid case for me. This may be out of scope for this pr though

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchDeleteSpec.groovy
```diff
@@ -1,281 +1,194 @@
...
+            swIsls.each { northbound.portUp(sw.dpId, it.srcPort) }
+            Wrappers.wait(discoveryTimeout + WAIT_OFFSET) {
+                def links = northbound.getAllLinks()
+                swIsls.each { assert islUtils.getIslInfo(links, it).get().state == IslChangeType.DISCOVERED }
+            }
         }
-        northbound.deleteFlow(flow.id)
+
+        where:
+        [linkType, actionToFix] << [["active", "Unplug and remove them first"], ["inactive", "Remove them first"]]
```
rtretyak: I think table will look prettier here

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchDeleteSpec.groovy
```diff
@@ -1,281 +1,194 @@
...
-        def switchId = topology.getActiveSwitches()[0].dpId
-        // get all active ISLs on switch
-        def activeSwitchLinks = northbound
-                .getAllSwitchLinks(switchId)
-                .findAll { it.state == IslChangeType.DISCOVERED }
-        // deactivate all active ISLs on switch
-        downLinkPorts(activeSwitchLinks)
-        Wrappers.wait(WAIT_OFFSET) {
-            northbound.getAllSwitchLinks(switchId).every { it.state != IslChangeType.DISCOVERED }
+        if (linkType == "inactive") {
```
rtretyak: too many `if`s in this test for no good reason. Instead of branching inside test it is better to create a separate test. I think we should avoid having `if`s in tests at all (which we actually do, if you search our specs for `if`s, you'll get close to 0 results), test case should be a valid set of steps with no branching, every branch is just another test case. Sometimes a use of small `if` is fine and justified, but I don't think that this case falls into such exception, we are trying to mix different steps to save some lines of code

ylobankov: Fixed.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchDeleteSpec.groovy
ylobankov: Fixed.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchDeleteSpec.groovy
ylobankov: Fixed.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchDeleteSpec.groovy
ylobankov: Let's have a separate PR for it. I added this test case to my 'todo' list.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchDeleteSpec.groovy
ylobankov: Fixed.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchDeleteSpec.groovy
ylobankov: Fixed.

PR #2039, opened 2019-02-12
Store all fields from OFPMP_DESC reply as a switch properties

### LINKED ISSUE #2033: Store all fields from OFPMP_DESC reply as a switch properties
When a switch connected to a controller, as a part of the initial exchange there is an OFPMP_DESC. We should store all the fields in a DB, like:
- Manufacturer desc.
- Hardware desc.
- Software desc.
- Serial no.
- Datapath desc.

surabujin:
Close #2033


surabujin:
Deprecated due to merge of new `network` topology.

Issue #2038, opened 2019-02-12
Unable to re-discover an ISL after deleteLink API was used


rtretyak:
Steps:
0. Choose a link that goes through intermediate not connected switch (A-switch)
1. Bring port down on one end of the ISL and wait until ISL is marked as FAILED in both directions
2. Delete the ISL using NB API (or delete directly from DB, it has the same effect)
3. Bring the port Up
**Expected:** Link is discovered again in both directions and have a DISCOVERED state
**Actual:** Link is discovered only in one direction, the other direction is not discovered and not present in DB. As a result, the discovered direction is also marked as FAILED.

Notes:  
We tracked down that the issue is related to ISL still being 'active' in one direction after port was brought down (though it is marked as FAILED in DB, it looks like it is not actually failed in-memory). As a result, when port is brought UP, the phantom 'active' ISL is not being rediscovered again (since it state "didn't change"), leading to a half-failed ISL.  
Another note is that if we wait for more than 'discoveryTimeout' before issuing a "port Up" command, the issue will not reproduce. This is another evidence of the above theory, since after discoveryTimeout the phantom 'active' half-ISL is forced to FAILED state, leading to successful rediscovery afterwards

PR #2041, opened 2019-02-13
#2038 Unable to re-discover an ISL after deleteLink API was used.

dpoltavets:
Added request to deactivate ISL in Event topology when deleting this ISL.

Closes: #2038.


services/wfm/src/main/java/org/openkilda/wfm/topology/event/OfeLinkBolt.java
```diff
@@ -441,6 +442,9 @@ protected void dispatchMain(Tuple tuple, InfoMessage infoMessage) {
             handleIslEvent(tuple, infoMessage);
         } else if (data instanceof DiscoPacketSendingConfirmation) {
             handleSentDiscoPacket((DiscoPacketSendingConfirmation) data);
+        } else if (data instanceof DeactivateIslInfoData) {
+            DeactivateIslInfoData deactivateIslInfoData = (DeactivateIslInfoData) data;
+            discovery.handleFailed(deactivateIslInfoData.getSrcSwitchId(), deactivateIslInfoData.getSrcPort());
```
timofei-durakov: what if we are in the middle of sync process?

dpoltavets: Good point! Now this event is handled before the condition for determining the state of the bolt. But still need to check the behavior of this code during synchronization.

Issue #1900, opened 2019-01-18
[Kilda-Gui] Make application name configurable for 2FA


hiteshgupta1:
1. Make application name configurable on 2FA authnticator
2. Change title on 2FA screen

PR #2044, opened 2019-02-13
 Enhancement to make application name configurable for 2FA

neeraj890kumar:
Resolves: #1900

### DIFF
--- a/confd/templates/gui/application.properties.tmpl
+++ b/confd/templates/gui/application.properties.tmpl
@@ -54,3 +54,5 @@ spring.mvc.view.prefix= /views/
 spring.mvc.view.suffix= .jsp
 
 status.cron.time=43200000
+
+twofa.application.name=Open Kilda
--- a/confd/templates/gui/application.properties.tmpl
+++ b/confd/templates/gui/application.properties.tmpl
@@ -55,4 +55,5 @@ spring.mvc.view.suffix= .jsp
 
 status.cron.time=43200000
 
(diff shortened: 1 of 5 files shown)


confd/templates/gui/application.properties.tmpl
```diff
@@ -54,3 +54,5 @@ spring.mvc.view.prefix= /views/
 spring.mvc.view.suffix= .jsp
 
 status.cron.time=43200000
+
+twofa.application.name=Open Kilda
```
ashishaggarwalCS: Change this to application name

services/src/openkilda-gui/src/main/java/org/openkilda/controller/LoginController.java
```diff
@@ -70,6 +71,9 @@
 
     @Autowired
     private PermissionRepository permissionRepository;
+    
+    @Value("${twofa.application.name}")
+    private String twofaApplicationName;
```
ashishaggarwalCS: Remove twofa the property name should be application name only

confd/templates/gui/application.properties.tmpl
hiteshgupta1: changes done

services/src/openkilda-gui/src/main/java/org/openkilda/controller/LoginController.java
hiteshgupta1: changes done

PR #2049, opened 2019-02-13
Various fixes for functional tests

ylobankov:


IlyaMoiseev:
Good, but need to add switch locking in DatabaseSupportImpl.resetCosts() before isl updating, to prevent deadlocks in test logic:
```
switchRepository.lockSwitches(
                    switchRepository.findAll().toArray(new Switch[0]));
```

PR #2062, opened 2019-02-19
Add functional tests to verify flow diversity

ylobankov:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowDiversitySpec.groovy
```diff
@@ -0,0 +1,326 @@
...
+        [flow1, flow2, flow3].each { flowHelper.addFlow(it) }
+
+        def (flow1Path, flow2Path, flow3Path) = [flow1, flow2, flow3].collect { northbound.getFlowPath(it.id) }
+        def allInvolvedIsls = [flow1Path, flow2Path, flow3Path].collect {
+            pathHelper.getInvolvedIsls(PathHelper.convert(it))
+        }.flatten()
+        assert allInvolvedIsls.unique(false) == allInvolvedIsls
+
+        expect: "Flow path response for all flows has correct overlapping segments stats"
+        def flow2SwitchCount = pathHelper.getInvolvedSwitches(flow2.id).size()
```
rtretyak: since you already have paths for each flow it is better to call `getInvolvedSwitches(path)` instead of passing the flow id

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowDiversitySpec.groovy
```diff
@@ -0,0 +1,326 @@
...
+        }
+        [flow1, flow2, flow3].each { flowHelper.addFlow(it) }
+
+        def (flow1Path, flow2Path, flow3Path) = [flow1, flow2, flow3].collect { northbound.getFlowPath(it.id) }
+        def allInvolvedIsls = [flow1Path, flow2Path, flow3Path].collect {
+            pathHelper.getInvolvedIsls(PathHelper.convert(it))
+        }.flatten()
+        assert allInvolvedIsls.unique(false) == allInvolvedIsls
+
+        expect: "Flow path response for all flows has correct overlapping segments stats"
```
rtretyak: this can be part of the first test. You can also add the same verification to the test where overlapping actually happens

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowDiversitySpec.groovy
```diff
@@ -0,0 +1,326 @@
...
+                        islPercent == expectedValuesMap["diverseGroup"][flowPath.id]["islPercent"]
+                        switchPercent == expectedValuesMap["diverseGroup"][flowPath.id]["switchPercent"]
+                    }
+                }
+                with(diverseGroup.otherFlows) { otherFlows ->
+                    otherFlows.each { flow ->
+                        assert flow.id in ([flow1, flow2, flow3]*.id - flowPath.id)
+                        with(flow.segmentsStats) {
+                            verifyAll {
+                                islCount == expectedValuesMap["otherFlows"][flowPath.id][flow.id]["islCount"]
```
rtretyak: `[flowPath.id][flow.id]`
how flow path id is different from flow id? I don't understand this. I see that these are params from different loops, but I think the names of these vars are strange

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowDiversitySpec.groovy
```diff
@@ -0,0 +1,326 @@
...
+        def flow1Path = PathHelper.convert(northbound.getFlowPath(flow1.id))
+
+        and: "Make all alternative paths less preferable by 1 for diverse flows"
+        allFlowPaths.findAll { it != flow1Path }.unique { it.first() }.findAll {
+            it.first().portNo != flow1Path.first().portNo
+        }.each {
+            def isl = pathHelper.getInvolvedIsls(it)[0]
+            northbound.updateLinkProps([new LinkPropsDto(isl.srcSwitch.dpId.toString(), isl.srcPort,
+                    isl.dstSwitch.dpId.toString(), isl.dstPort, ["cost": (diversityIslWeight + 1).toString()])])
+        }
```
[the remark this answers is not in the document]
ylobankov: Fixed.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowDiversitySpec.groovy
```diff
@@ -0,0 +1,326 @@
...
+        flow2Path == flow1Path
+
+        when: "Create the third flow with diversity enabled"
+        def flow3 = flowHelper.randomFlow(srcSwitch, dstSwitch, false, [flow1, flow2]).tap {
+            it.diverseFlowId = flow2.id
+        }
+        flowHelper.addFlow(flow3)
+        def flow3Path = PathHelper.convert(northbound.getFlowPath(flow3.id))
+
+        then: "The flow is built through one of the alternative paths because they are preferable already"
```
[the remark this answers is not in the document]
ylobankov: Fixed 209-215 lines. Now it should be clear.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowDiversitySpec.groovy
ylobankov: I would prefer to keep it as a separate test to not make the test bulky.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowDiversitySpec.groovy
```diff
@@ -0,0 +1,417 @@
...
+        def flow1 = flowHelper.randomFlow(srcSwitch, dstSwitch)
+        flowHelper.addFlow(flow1)
+        def flow1Path = PathHelper.convert(northbound.getFlowPath(flow1.id))
+
+        and: "Make all min alternative paths less preferable by 1 for diverse flows"
+        def altPaths = database.getPaths(srcSwitch.dpId, dstSwitch.dpId)*.path
+        altPaths.remove(flow1Path)
+        def flow1PathCost = pathHelper.getCost(flow1Path) + diversityIslWeight + diversitySwitchWeight * 2
+        def minAltPathCost = pathHelper.getCost(altPaths.min { it.size() }) + diversitySwitchWeight * 2
+        if (flow1PathCost > minAltPathCost) {
```
rtretyak: Ok, but we should understand that if `flow1PathCost <= minAltPathCost` then we will not test the `+1` boundary cost difference (which is explicitly stated by step description). You could have just removed the `if`, then you'll get the negative difference and it will kind of work, but only while `islCost` is bigger than the negative `difference` (L187 may get a negative cost). Just commenting so that all of us understand it.
I would opt for eliminating the conditional branching and always go for the expected +1 difference between current and alt path. Right now the test has multiple different ways of being executed

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowDiversitySpec.groovy
```diff
@@ -0,0 +1,417 @@
...
+        when: "Create the second flow with diversity enabled"
+        def flow2 = flowHelper.randomFlow(srcSwitch, dstSwitch, false, [flow1]).tap { it.diverseFlowId = flow1.id }
+        flowHelper.addFlow(flow2)
+        def flow2Path = PathHelper.convert(northbound.getFlowPath(flow2.id))
+
+        then: "The flow is built through the most preferable path (path of the first flow)"
+        flow2Path == flow1Path
+
+        when: "Create the third flow with diversity enabled"
+        if (flow1PathCost < minAltPathCost) {
```
rtretyak: One more `if`, which is actually a wrong assumption from my point of view, since `minAltPathCost` has already changed on lines L180-L189(maybe changed or maybe not ¯\\_(ツ)_/¯), so it is operating with potentially outdated values. or is it intentional?
`flow1PathCost` is also outdated at this point, sinse flow1PathCost now has the `flow2` on it, so the 'diverse cost' of this path has changed as soon as you built a second flow through it

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowDiversitySpec.groovy
```diff
@@ -0,0 +1,417 @@
...
+        def involvedIsls = [flow2Path, flow3Path].collectMany { pathHelper.getInvolvedIsls(it) }
+        flow3Path != flow2Path
+        involvedIsls.unique(false) == involvedIsls
+
+        and: "Delete flows and link props"
+        [flow1, flow2, flow3].each { flowHelper.deleteFlow(it.id) }
+        northbound.deleteLinkProps(northbound.getAllLinkProps())
+    }
+
+    def "Able to get flow paths with correct overlapping segments stats (casual flows)"() {
```
rtretyak: This overlaps with the first test, isn't it? Moreover, this test will get more complete if you explicitly show the `then: "All flows have different paths"` verification from the first test (which is the only difference between the tests). Then it looks more obvious _why_ the overlapping segments have given values

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowDiversitySpec.groovy
rtretyak: Ok, got it. I think renaming `flow.id` to `otherFlow.id` will make it more clear, since `flow.id` unintentionally associates with the main `flowid` in that json

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowDiversitySpec.groovy
rtretyak: Ok, just saw your comment. We can ofc leave as is, this is not crucial

Issue #2065, opened 2019-02-20
A reroute got initiated in case of port(non-ISL) down on one of the flow endpoints


gauravchug:
# Description
We have a flow sw1_p1-sw2_p2. 

# Expected
if sw1_p1 or sw2_p2 change its state (UP/DOWN), a flow reroute command should not be initiated

# Actual
A flow reroute command observed for a flow when on one of the endpoints a port DOWN event detected.

PR #2077, opened 2019-02-25
#2065 Disable reroute flow when one of the flow endpoints is not active.

dpoltavets:
Closes: #2065.


dpoltavets:
@pzakatov I will redo it. Thank you!

PR #2091, opened 2019-02-27
2076: Add cleanup extension for hardware env

rtretyak:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/extension/env/EnvCleanupExtension.groovy
```diff
@@ -0,0 +1,155 @@
...
+
+    def resetBandwidth(List<IslInfoData> links) {
+        def topoIsls = topology.isls.collectMany { [it, islUtils.reverseIsl(it)] }
+        links.each { link ->
+            if (link.maxBandwidth != link.availableBandwidth || link.maxBandwidth != link.speed) {
+                def isl = topoIsls.find {
+                    it.srcSwitch.dpId == link.source.switchId && it.srcPort == link.source.portNo &&
+                            it.dstSwitch.dpId == link.destination.switchId && it.dstPort == link.destination.portNo
+                }
+                assert isl
```
ylobankov: I am not sure that we need any asserting in extensions. In my opinion, it is better to raise an exception that topology has missing ISL.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/extension/env/EnvCleanupExtension.groovy
```diff
@@ -0,0 +1,155 @@
...
+    void start() {
+        SpringContextExtension.addListener(this)
+    }
+
+    def deleteAllFlows() {
+        log.info("Deleting all flows")
+        northbound.deleteAllFlows()
+    }
+
+    def unsetIslMaintenance(List<IslInfoData> links) {
```
ylobankov: unsetIslMaintenance -> unsetLinkMaintenance. We use 'link' in other methods.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/extension/env/EnvCleanupExtension.groovy
```diff
@@ -0,0 +1,155 @@
...
+    def resetBandwidth(List<IslInfoData> links) {
+        def topoIsls = topology.isls.collectMany { [it, islUtils.reverseIsl(it)] }
+        links.each { link ->
+            if (link.maxBandwidth != link.availableBandwidth || link.maxBandwidth != link.speed) {
+                def isl = topoIsls.find {
+                    it.srcSwitch.dpId == link.source.switchId && it.srcPort == link.source.portNo &&
+                            it.dstSwitch.dpId == link.destination.switchId && it.dstPort == link.destination.portNo
+                }
+                assert isl
+                log.info("Resetting available bandwidth on ISL $isl")
```
ylobankov: The same question with printing as above.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/extension/env/HardwareEnvCleanupExtension.groovy
```diff
@@ -0,0 +1,49 @@
...
+ * Turned on on-demand by passing a `-Dcleanup` property to the build params.<br>
+ * This has no guarantee to fully repair the broken env and should be used with care,
+ * especially if the target environment is shared between many people.<br>
+ *
+ * WARNING: do not over-use this option. Cleanup may be destructive and mask potential defects related to improper
+ * topology discovery, default rules setup on switch discovery, bandwidth discovery etc.
+ */
+@Slf4j
+class HardwareEnvCleanupExtension extends EnvCleanupExtension {
+    @Value('${cleanup:false}')
```
ylobankov: I think it should name 'env.hardware.cleanup'. Also, it should be added to kilda.properties.example and 'confd'.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/extension/env/EnvCleanupExtension.groovy
rtretyak: This is an absolutely abnormal and rare situation, which can moreover mask a defective ISL discovery by Kilda. From my point of view it even deserves to be of a 'warn' level.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/extension/env/HardwareEnvCleanupExtension.groovy
rtretyak: > I think it should name 'env.hardware.cleanup'.

agree
> it should be added to kilda.properties.example and 'confd'

Maybe add to kilda.properties.example, but have doubts about 'confd'. I see this more as a flag which is being set when calling `mvn test...` command, rather that an option to be stored along with all the static properties. Not sure whether it belongs to the list of all kilda variables. My thought process is like: It is false by default, turn it on during the build if you know what you're doing.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/extension/env/EnvCleanupExtension.groovy
rtretyak: done

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/extension/env/EnvCleanupExtension.groovy
ylobankov: I understand that 'assert is a way to raise an exception. But in my opinion, we should do asserting only in tests. In cases like this, we should raise a certain exception with the comprehensive message. Moreover in your case, if there is no ISL, we will get something like this:
```
Condition not satisfied:

isl
|
null
```
which in fact gives us nothing.
If we have a construction like this:
```
if (!isl) {
    throw new Exception("Missing ISL in topology. ISL: $link")
}
```
we will understand that we have missing ISL and what exactly ISL is missing.
Of course we can do
```
assert isl, "Missing ISL in topology. ISL: $link"
```
but it is not good I think.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/extension/env/EnvCleanupExtension.groovy
ylobankov: Ok. Please add ':' after 'ISL' to be consistent with other messages.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/extension/env/EnvCleanupExtension.groovy
```diff
@@ -0,0 +1,155 @@
...
+                database.revertIslBandwidth(isl)
+            }
+        }
+    }
+
+    def unsetSwitchMaintenance() {
+        def maintenanceSwitches = northbound.getAllSwitches().findAll { it.underMaintenance }
+        if (maintenanceSwitches) {
+            log.info("Unset maintenance mode from all affected switches")
```
ylobankov: Isn't logical to do the same for switches as for links? Or there is another logic and we log only that what is related to links?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/extension/env/EnvCleanupExtension.groovy
```diff
@@ -0,0 +1,155 @@
...
+                    // This situation may require manual intervention
+                }
+            }
+        }
+    }
+
+    def deleteLinkProps() {
+        def linkProps = northbound.getAllLinkProps()
+        if (linkProps) {
+            log.info("Deleting all link props")
```
ylobankov: According to your answer above, let's log link props as well. I like the idea to be more verbose.

services/src/functional-tests/kilda.properties.example
```diff
@@ -22,6 +22,7 @@ lab-api.endpoint=http://localhost:8288
 lockkeeper.endpoint=http://localhost:5001
 
 otsdb.endpoint=http://localhost:4242
+opentsdb.metric.prefix=kilda.
```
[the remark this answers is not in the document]
rtretyak: yes you're right. I pulled this during rebase, will rename here and in corresponding tests

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/extension/env/EnvCleanupExtension.groovy
rtretyak: done

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/extension/env/EnvCleanupExtension.groovy
rtretyak: done

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/extension/env/HardwareEnvCleanupExtension.groovy
rtretyak: done, no confd

services/src/functional-tests/kilda.properties.example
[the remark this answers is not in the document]
rtretyak: done. 
otsdb.endpoint -> opentsdb.endpoint

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/extension/env/EnvCleanupExtension.groovy
rtretyak: done

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/extension/env/EnvCleanupExtension.groovy
rtretyak: done

PR #2092, opened 2019-02-27
Add test to verify flow is not rerouted when flow port goes down

### LINKED ISSUE #2090: Create a functional test to verify flow is not rerouted when one of the flow endpoints goes to the "down" state

ylobankov:
Closes #2090.

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/AutoRerouteSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/AutoRerouteSpec.groovy
@@ -388,6 +388,36 @@ class AutoRerouteSpec extends BaseSpecification {
         flowHelper.deleteFlow(flow.id)
     }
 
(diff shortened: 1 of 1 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/AutoRerouteSpec.groovy
```diff
@@ -388,6 +388,36 @@ class AutoRerouteSpec extends BaseSpecification {
         flowHelper.deleteFlow(flow.id)
     }
 
+    def "Flow is not rerouted when one of the flow endpoints goes down"() {
```
rtretyak: 'flow endpoints' sounds a bit misleading to me. If I didn't see the test, I would have thought that the src/dst switch goes down.

ylobankov: fixed.

PR #2096, opened 2019-02-28
Extend/fix functional tests for meters

ylobankov:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/MetersSpec.groovy
```diff
@@ -55,87 +57,96 @@ class MetersSpec extends BaseSpecification {
...
         exc.rawStatusCode == 400
 
         where:
-        meterId | sw                        | switchType
-        -1      | getNonCentecSwitches()[0] | "non-Centec"
-        0       | getCentecSwitches()[0]    | "Centec"
-        -1      | getCentecSwitches()[0]    | "non-Centec"
-        0       | getNonCentecSwitches()[0] | "Centec"
+        meterId | switches               | switchType
+        -1      | getNonCentecSwitches() | "non-Centec"
```
rtretyak: We can leave previous implementation with `sw` and add the assumption directly to the `getCertainSwitches` method. This is not that important, but will allow to get rid of duplicate assumptions in every test

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/MetersSpec.groovy
```diff
@@ -55,87 +57,96 @@ class MetersSpec extends BaseSpecification {
...
+
+        given: "A #switchType switch with OpenFlow 1.3 support"
+        def sw = switches.first()
 
         when: "Get default meters from the switch"
         def defaultMeters = northbound.getAllMeters(sw.dpId)
         assert defaultMeters
 
-        and: "Create a flow"
+        and: "Create a single-switch flow"
```
rtretyak: I think this detail is important enough to be reflected in the test name

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/MetersSpec.groovy
ylobankov: Unfortunately, we cannot. If the first test is skipped, all next iterations will be skipped as well. Probably, this behavior is introduced by the 'assumeTrue' method. My guess is that '
getCertainSwitches' method is called the very first, the condition is not satisfied and we skip the test not reaching Spock.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/MetersSpec.groovy
ylobankov: Fixed.

Issue #2088, opened 2019-02-27
Priority rerouting


pzakatov:
# Description
It is required to prioritize flows in case a reroute event occur. So we need to introduce a new parameter for a flow, and take this parameter when we issue a reroute requests

# Details
Add a new parameter `priority` to a flow. Integer, Optional, default configurable.

A logic for rerouting (simplified):
1 Find all affected flows
2 Sort them by priority ascending, creation time ascending
3 Send reroute requests based on this ordered list

Example: a flow with priority=555 will be scheduled for rerouting earlier than flow with priority=777

kafka key = correlation-id
implementation area: reroute throttling

PR #2100, opened 2019-03-01
Add "time_create" field to the flow object.

dpoltavets:
Part of: #2088.


services/src/messaging/src/main/java/org/openkilda/messaging/model/FlowDto.java
```diff
@@ -156,6 +159,7 @@ public FlowDto() {
      * @param ignoreBandwidth   ignore bandwidth flag
      * @param cookie            cookie
      * @param description       description
+     * @param timeCreate       flow create timestamp
```
niksv: minor: missed space. `flow create timestamp` should be moved left

PR #2111, opened 2019-03-01
Refactor and fix atdd-staging

### LINKED ISSUE #2004: fix Atdd-staging
Atdd-staging has run a bit out of date and some tests are not valid. This needs to be fixed

ylobankov:
Closes #2004.


services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/steps/HealthCheckSteps.java
```diff
@@ -0,0 +1,45 @@
+/* Copyright 2018 Telstra Open Source
```
andriidovhan: 2018-2019

andriidovhan: the same issue in all new created  files

ylobankov: Fixed.

PR #2119, opened 2019-03-04
AF - only for review process

nikitamarchenko:


nikitamarchenko:
merged to discovery-topology branch

Issue #1630, opened 2018-11-22
Make OpenTSDB metric prefix configurable


pzakatov:
now it's hardcoded. should be configurable

Metric template proposal: `<prefix><metric name>`

pzakatov:
@dpoltavets , you forgot about GUI. Please fix.

PR #2122, opened 2019-03-04
#1630 Make OpenTSDB metric prefix configurable.

dpoltavets:
@gauravchug, @ashishaggarwalCS  I've added configurable OpenTSDB metric prefix to GUI module, but struggle with the next line: https://github.com/telstra/open-kilda/blob/develop/services/src/openkilda-gui/ui/src/app/common/services/dygraph.service.ts#L336
Could you please help me to fix it? You can push in this PR.

### DIFF
--- a/confd/templates/gui/application.properties.tmpl
+++ b/confd/templates/gui/application.properties.tmpl
@@ -25,9 +25,9 @@ spring.mvc.throw-exception-if-no-handler-found=true
 nb.base.url={{ getv "/kilda_northbound_endpoint" }}:{{ getv "/kilda_northbound_rest_port" }}/api/v1
 
 
-#OPEN TSDB Base URL
+#OPEN TSDB Base URL and metric prefix
 opentsdb.base.url=http://{{ getv "/kilda_opentsdb_hosts" }}:{{ getv "/kilda_opentsdb_port" }}
-
+opentsdb.metric.prefix = {{ getv "/kilda_opentsdb_metric_prefix" }}
 
 
 #Kilda username and password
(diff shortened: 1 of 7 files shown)


hiteshgupta1:
> > Reviewed 3 of 6 files at r1, 4 of 4 files at r2.
> > _[Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2122)_ status: all files reviewed, 1 unresolved discussion (waiting on @ashishaggarwalCS, @dpoltavets, and @gauravchug)
> 
> _[services/src/openkilda-gui/ui/src/app/common/services/dygraph.service.ts, line 336 at r2](https://reviewable.io/reviews/telstra/open-kilda/2122#-L_Mn2LPD7Z3TakiZNYo:-L_Mn2LQ7_UpikggJNHv:b-xxejq0) ([raw file](https://github.com/telstra/open-kilda/blob/be02668e0126e2a72abd076a393a396aad59bf0f/services/src/openkilda-gui/ui/src/app/common/services/dygraph.service.ts#L336)):_
> 
> > ```
> >           var dataValues = typeof data[j] !== "undefined" ? data[j].dps : null;
> >           var metric = typeof data[j] !== "undefined" ? data[j].metric : "";
> >           var tempMetric = metric.split('.').slice(1).join('.') /**remove metric prefix */
> > ```
> 
> prefix could be an empty string. In that case, the logic here would be broken. Should we use 'String search()' method here?

Changes Updated : There is no need to place metrics check on front-end as its already fixed from back-end, so removed for now.

Issue #2142, opened 2019-03-12
Discovery failure keeps ISL actual status as active


sergii-iakovenko:
Discovery failure keeps the actual status of one of paired ISLs as active. This may lead to the situation when a subsequent discovery event activates ISL although only one direction of ISL is confirmed.

Steps to reproduce:
- Discovered and active ISL (there must be ability to cut the forward and reverse links separately).
- Cut the forward link, wait until the discovery process marks it as inactive (both status ="inactive" and actual="inactive").
- Cut the reverse link.

Expected result:
- The discovery process marks the reverse link as inactive (actual="inactive", status is already "inactive").

Actual result:
- The discovery process keeps the reverse link actual status as is (actual="active").

PR #2151, opened 2019-03-13
Mark ISL as inactive is discovery failed.

Settis:
The ISL will be marked as inactive while discovery failed processing.
It's done regardless ISL previous state.


Settis:
Close: #2142

PR #2152, opened 2019-03-13
Self-executable floodlight commands

nikitacherevko:
- created new module fl-root with two submodules: floodlight-api with messages and floodlight-modules with main functionality for interacting with switches;
- created self executable commands for installing rules (it should help to get rid of SwitchManager and RecordHandler);
- made commands idempotent, so you are able to execute commands as much as necessary;
- implemented the way how to floodlight can sent error response with details what exactly went wrong during rule installation;
- added "flat"messages for communication between storm topologies and floodlight.

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2152)


services/src/fl-root/floodlight-modules/src/main/java/org/openkilda/floodlight/command/flow/InstallIngressRuleCommand.java
```diff
@@ -0,0 +1,182 @@
...
+
+@Getter
+public class InstallIngressRuleCommand extends InstallTransitRuleCommand {
+
+    private final Long bandwidth;
+    private final Integer inputVlanId;
+    private final OutputVlanType outputVlanType;
+    private final Long meterId;
+
+    @JsonCreator
```
sergii-iakovenko: This class is serializable to JSON, which means that it's supposed to be built and transferred by another modules than FL. Is it correct? If so then those modules get dependence on org.projectfloodlight.

nikitacherevko: No, these "commands" are only for internal floodlight purposes. They are created by deserializing from "floodligh API" classes that are used by another modules

services/wfm/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/FlowCreateFsm.java
```diff
@@ -0,0 +1,135 @@
...
+package org.openkilda.wfm.topology.flowhs.fsm;
+
+import org.openkilda.wfm.topology.flowhs.fsm.FlowCreateFsm.Event;
+import org.openkilda.wfm.topology.flowhs.fsm.FlowCreateFsm.State;
+
+import org.squirrelframework.foundation.fsm.StateMachineBuilder;
+import org.squirrelframework.foundation.fsm.StateMachineBuilderFactory;
+import org.squirrelframework.foundation.fsm.impl.AbstractStateMachine;
+
+public class FlowCreateFsm extends AbstractStateMachine<FlowCreateFsm, State, Event, Object> {
```
sergii-iakovenko: The flowhs package and FlowCreateFsm class shouldn't be a part of this PR.

sergii-iakovenko:
This PR has wrong base, therefore another PR is to be created to merge these changes into develop.

PR #2153, opened 2019-03-13
Rule install commands with new data model support

nikitacherevko:


services/wfm/src/main/java/org/openkilda/wfm/topology/flowhs/service/FlowCommandFactory.java
```diff
@@ -0,0 +1,345 @@
...
+import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.collections4.ListUtils;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+
+@Slf4j
+public class FlowCommandFactory {
```
sergii-iakovenko: Shouldn't we somehow define that this factory is tightly bound to TransitVlan encapsulation? And provide a contract for other implementation of flow command factory.

PR #2155, opened 2019-03-13
Protected paths implementation

### LINKED ISSUE #1232: As a user I'd like to create a flow with a protected path
# Details
- a user should be able to create flows with a protected path (optionally)
- the path should be pre-calculated, reserved, and deployed (except ingress rule), so we can switch traffic fast.

# API
1. `flow` object should be extended with an integer parameter `allocate-protected-path` with values `false`(default) and `true`
2. `/flows/{flow-id}/path` should also return `protected_path'  with the protected path and its diversity factor in comparison to the primary one.

IlyaMoiseev:
Closes #1232

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2155)


services/src/messaging/src/main/java/org/openkilda/messaging/payload/flow/GroupFlowPathPayload.java
```diff
@@ -53,15 +55,46 @@
...
         this.id = id;
         this.forwardPath = forwardPath;
         this.reversePath = reversePath;
         this.segmentsStats = segmentsStats;
+        this.protectedPath = protectedPath;
+    }
+
+    @Data
+    @NoArgsConstructor
+    public static class FlowProtectedPathPayload implements Serializable {
```
sergii-iakovenko: Shouldn't it be named in plural form? E.g. FlowProtectedPathsPayload

services/wfm/src/main/java/org/openkilda/wfm/topology/flow/StreamType.java
```diff
@@ -24,6 +24,11 @@
      */
     CREATE,
 
+    /**
+     * Get flow path swap stream.
```
sergii-iakovenko: "Get" is not needed here.

PR #2161, opened 2019-03-14
Combine nb api and northbound modules, prepare NB for API v2 support

nikitacherevko:


sergii-iakovenko:
In general looks good. Need to discuss the name convention for grouping modules (nb-root).

Issue #1342, opened 2018-09-17
Implement Floodlight network events feature


surabujin:
Replace existing event-topology to FL sync scheme with new one described in design documents in #650.

Design document define "automatic" sync procedure, but we also need on-demand sync procedure. It is required to start work after `event`-topology (re)start.

This feature require to change FL, FL-router and event-topology.

Issue #1341, opened 2018-09-17
Implement Floodlight Router topology


surabujin:
Implement FL-router storm topology using design from #650. Also alter FL kafka-write endpoints to route them via new topology.

Support of `multi-switch commands` is outside scope of this task.

PR #2170, opened 2019-03-18
Mfl part three

timofei-durakov:


services/wfm/src/main/java/org/openkilda/wfm/topology/floodlightrouter/bolts/DiscoveryBolt.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    @Override
+    protected void doWork(Tuple input) {
+        String sourceComponent = input.getSourceComponent();
+        currentTuple = input;
+        Message message = null;
+        try {
+            String json = input.getValueByField(AbstractTopology.MESSAGE_FIELD).toString();
+            message = MAPPER.readValue(json, Message.class);
```
nikitacherevko: Why not use the spout that deserializes message and you don't have to do it by yourself.

services/wfm/src/main/java/org/openkilda/wfm/topology/floodlightrouter/service/RouterService.java
```diff
@@ -0,0 +1,140 @@
...
+            } else if (infoData instanceof NetworkDumpSwitchData) {
+                switchId = ((NetworkDumpSwitchData) infoData).getSwitchRecord().getDatapath();
+            } else if (infoData instanceof SwitchInfoData) {
+                switchId = ((SwitchInfoData) infoData).getSwitchId();
+            } else if (infoData instanceof PortInfoData) {
+                switchId = ((PortInfoData) infoData).getSwitchId();
+            }
+
+            // NOTE(tdurakov): need to notify of a mapping update
+            if (switchId != null && region != null && floodlightTracker.updateSwitchRegion(switchId, region)) {
```
nikitacherevko: 1. Is it a possible scenario that we can get a response from speaker without region?
2. Why we always update mapping between a region and a switch? I believe we don't have to do it so often, only do it if it was changed.

services/wfm/src/main/java/org/openkilda/wfm/topology/floodlightrouter/service/FloodlightTracker.java
```diff
@@ -18,32 +18,34 @@
...
 import java.util.stream.Collectors;
 
 @Slf4j
 public class FloodlightTracker {
     @VisibleForTesting
-    protected ConcurrentMap<SwitchId, String> switchRegionMap = new ConcurrentHashMap<>();
+    protected Map<SwitchId, String> switchRegionMap = new HashMap<>();
     @VisibleForTesting
-    protected ConcurrentMap<String, FloodlightInstance> floodlightStatus = new ConcurrentHashMap<>();
+    protected Map<String, FloodlightInstance> floodlightStatus = new HashMap<>();
```
nikitacherevko: It's not really storage for floodlight statuses, is it?

services/src/configuration/src/main/java/org/openkilda/config/KafkaTopicsConfig.java
```diff
@@ -36,18 +36,28 @@
...
+    @Default("kilda.northbound.storm")
     String getNorthboundTopic();
 
+    @Key("northbound.region")
+    @FallbackKey("kafka.northbound.topic.region")
+    @Default("kilda.northbound")
+    String getNorthboundRegionTopic();
+
     @Key("opentsdb")
     @Default("kilda.otsdb")
```
nikitacherevko: Why you skipped this one for opentsdb storm topology? It would be great to have kafka topics for storm topologies written in a same way. Do you skipped this topic because we are not the only ones who read this topic?

nikitacherevko:
In general looks good, left minor comments

services/wfm/src/main/java/org/openkilda/wfm/topology/floodlightrouter/bolts/DiscoveryBolt.java
timofei-durakov: will change it in follow up patch

services/wfm/src/main/java/org/openkilda/wfm/topology/floodlightrouter/service/RouterService.java
timofei-durakov: 1 - no, but this validation should be here
2 - it's important actually. Regions would be set up dynamically, so this regular update allows not to redeploy topology once switches change region

services/wfm/src/main/java/org/openkilda/wfm/topology/floodlightrouter/service/FloodlightTracker.java
timofei-durakov: kind of, maybe worth renaming indeed

services/src/configuration/src/main/java/org/openkilda/config/KafkaTopicsConfig.java
timofei-durakov: otsdb doesn't communicate with floodlight directly, we could change this also, but let's do it in a separate patch

timofei-durakov:
Closes #650 #1341 #1342

PR #2171, opened 2019-03-18
Allow to run ElasticSearch on remote Docker.

Settis:
Open ElasticSearch ports for any net interfaces not only localhost.
Change host name in elk-dashboard load script.


Settis:
I'll split it.

PR #2181, opened 2019-03-19
MultiFL. Add FL for statistics.

Settis:
Add FL_second container with FL only for statistics.
Add a service to FL for setting default switch role.
Changing lab-service to work with two FL simultaneously.


Settis:
Work proceed in #2279.

PR #2182, opened 2019-03-20
Fix incorrect calculation of "burst_size".

dpoltavets:


rtretyak:
We need more investigation with hw switches. It seems that different Noviflow firmware rounds that burst in different ways..

PR #2186, opened 2019-03-21
Remove execute method override into `CoordinatedBolt`

surabujin:
This method is entry point for tuple processing and AbstractBolt make
different sort of initialisazation and cleenup. If some class override
this method it must be keeped in sync with parent class.

This change allow to remove existing override of execute method. And get
rid from "keep in sync" problem completelly.

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2186)

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/share/hubandspoke/CoordinatedBolt.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/share/hubandspoke/CoordinatedBolt.java
@@ -16,6 +16,7 @@
 package org.openkilda.wfm.share.hubandspoke;
 
 import org.openkilda.wfm.AbstractBolt;
+import org.openkilda.wfm.error.AbstractException;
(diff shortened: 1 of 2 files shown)


services/wfm/src/main/java/org/openkilda/wfm/share/hubandspoke/CoordinatedBolt.java
```diff
@@ -40,22 +41,19 @@
...
+        if (CoordinatorBolt.ID.equals(input.getSourceComponent())) {
+            String key = input.getStringByField(MessageTranslator.KEY_FIELD);
+            onTimeout(key);
+        } else {
+            super.dispatch(input);
+        }
+    }
+
+    @Override
+    public void ack(Tuple input) {
```
timofei-durakov: tbh I don't like suck ack implementation of public method that will do something if and only if I set flag to true. This is error leading way as for me. Maybe it's worth to improve this thing.

PR #2190, opened 2019-03-22
Floodlight router sends periodic network dumps for to fl

timofei-durakov:
added new config param
and periodic action for network dump requests


surabujin:
Can be closed because this commit have been absorbed by new discovery(network) topology.

Issue #1060, opened 2018-07-27
Restore flow vs ISL collision validation


surabujin:
We had to revert "flow vs ISL collision validation" due to existing issue with self-looped ISLs. i.e. on customer port we detect fake ISL due to receiving out discovery packages.

As result this switch/port marked as ISL and customer flow can't be created.

To be able to restore this check we need to "remove" self-looped ISL https://github.com/telstra/open-kilda/issues/1059. Also we should implement https://github.com/telstra/open-kilda/issues/579 to completely get rid from "invalid" ISL.

Original issue https://github.com/telstra/open-kilda/issues/183
Revert https://github.com/telstra/open-kilda/commit/8bd677d9e7f42676feb8077faf2b753d979d9e3b (#1024)

PR #2192, opened 2019-03-25
#1060 Add check for conflicts with ISL ports in FlowValidator.

dpoltavets:
Closes: #1060.

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/flow/validation/FlowValidator.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/flow/validation/FlowValidator.java
@@ -21,6 +21,7 @@ import org.openkilda.messaging.error.ErrorType;
 import org.openkilda.model.Flow;
 import org.openkilda.model.SwitchId;
 import org.openkilda.persistence.repositories.FlowRepository;
+import org.openkilda.persistence.repositories.IslRepository;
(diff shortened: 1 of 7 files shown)


services/wfm/src/main/java/org/openkilda/wfm/topology/flow/validation/FlowValidator.java
```diff
@@ -68,6 +71,34 @@ void checkBandwidth(Flow flow) throws FlowValidationException {
...
 
+    /**
+     * Checks a flow for conflicts with ISL ports.
+     *
+     * @param requestedFlow a flow to be validated.
+     * @throws FlowValidationException is thrown in a case when flow endpoints conflict with existing ISL ports.
+     */
+    void checkFlowForIslConflicts(Flow requestedFlow) throws FlowValidationException {
+        // Check the source
+        if (!islRepository.findBySrcEndpoint(requestedFlow.getSrcSwitch().getSwitchId(),
```
sergii-iakovenko: It would be more optimal to create a combined method in IslRepository to lookup an isl by both source or destination. So we'll have a single call to Neo4j instead of 2.

dpoltavets: You are right. Fixed.

timofei-durakov:
@dpoltavets fix confilicts, please

Issue #2194, opened 2019-03-26
NullPointerException in PathVerificationService.


dpoltavets:
When a switch is disconnected when sending a discovery event, we get NPE:
```
java.lang.NullPointerException: null
    at org.openkilda.floodlight.pathverification.PathVerificationService.getSwitchPortSpeed(PathVerificationService.java:537)
    at org.openkilda.floodlight.pathverification.PathVerificationService.handlePacketIn(PathVerificationService.java:497)
    at org.openkilda.floodlight.pathverification.PathVerificationService$1.call(PathVerificationService.java:190)
    at org.openkilda.floodlight.pathverification.PathVerificationService$1.call(PathVerificationService.java:187)
    at org.openkilda.floodlight.command.CommandWrapper.call(CommandWrapper.java:38)
    at org.openkilda.floodlight.command.CommandWrapper.call(CommandWrapper.java:26)
    at org.openkilda.floodlight.service.CommandProcessorService.lambda$executeOneShot$0(CommandProcessorService.java:130)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
    at java.lang.Thread.run(Thread.java:748)
```

PR #2195, opened 2019-03-26
Fix NPE in PathVerificationService.

dpoltavets:
Closes: #2194.

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2195)

### DIFF
--- a/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/pathverification/PathVerificationService.java
+++ b/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/pathverification/PathVerificationService.java
@@ -411,6 +411,12 @@ public class PathVerificationService implements IFloodlightModule, IPathVerifica
         }
 
         try {
+            IOFSwitch destSwitch = switchService.getSwitch(input.getDpId());
+            if (destSwitch == null) {
+                logger.error("Destination switch {} was not found", input.getDpId());
+                return;
+            }
+
             ByteBuffer portBb = ByteBuffer.wrap(verificationPacket.getPortId().getValue());
             portBb.position(1);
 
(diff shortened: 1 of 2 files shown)


dpoltavets:
___
*[services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/pathverification/PathVerificationService.java, line 415 at r1](https://reviewable.io/reviews/telstra/open-kilda/2195#-Lb3tUpg2WDKT2Y4WPPO:-Lb4BAdg7xtwN3cvWBPK:b-dn3zt4) ([raw file](https://github.com/telstra/open-kilda/blob/7da2fe02cb78985aaf31646585f15a27c509ad53/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/pathverification/PathVerificationService.java#L415)):*
<details><summary><i>Previously, surabujin (Dmitriy Bogun) wrote…</i></summary><blockquote>

Better name it `destSwitch`, it is more correct(and more definitely) in terms of our discovery process.
</blockquote></details>

You are right. Thank you! 
Fixed.

dpoltavets:
___
*[services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/pathverification/PathVerificationService.java, line 417 at r1](https://reviewable.io/reviews/telstra/open-kilda/2195#-Lb3sdXl9eNgOSDrcDhh:-Lb4BKQzDVndNmjxM-oT:b-lx4fjp) ([raw file](https://github.com/telstra/open-kilda/blob/7da2fe02cb78985aaf31646585f15a27c509ad53/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/pathverification/PathVerificationService.java#L417)):*
<details><summary><i>Previously, surabujin (Dmitriy Bogun) wrote…</i></summary><blockquote>

Exception here is a bad idea. There is no one upper in the stack who(our code who know what is going on here) can catch and handle it. Properly log error here and make `return`.
</blockquote></details>

Fixed.

Issue #1434, opened 2018-10-08
OTSDB sometimes throws 'duplicate timestamp' error when requesting data


rtretyak:
From time to time a call to OTSDB throws exception in functional-tests run: [error.txt](https://github.com/telstra/open-kilda/files/2455201/error.txt). 
Didn't find the exact steps to reproduce. Need help addressing the issue.
1. Should we turn on the `tsd.storage.fix_duplicates=true` param in OTSDB?
2. Should we change the way we request data from OTSDB in tests?
3. Should we change the way we post data into OTSDB?

PR #2203, opened 2019-03-27
Improve SkippedTestsLogger to process ignored tests as well

rtretyak:
-Also ignore BfdSpec as BFD is not available in Kilda atm
-Update stats test due to #1434 being fixed recently

depends on #2197

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/resilience/StormHeavyLoadSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/resilience/StormHeavyLoadSpec.groovy
@@ -32,7 +32,7 @@ class StormHeavyLoadSpec extends BaseSpecification {
     def r = new Random()
 
     /**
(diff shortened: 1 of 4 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/resilience/StormHeavyLoadSpec.groovy
```diff
@@ -40,7 +40,7 @@ class StormHeavyLoadSpec extends BaseSpecification {
 
         when: "Produce massive amount of messages into topo.disco topic"
         def messages = 100000 //total sum of messages of all types produced
-        def operations = 3 //discovery, port up, port down
+        def operations = 2 //port up, port down
```
ylobankov: Need to update description at L35.

rtretyak: fxd

PR #2209, opened 2019-03-28
Add test to verify system consistency after wfm restart

rtretyak:
- also fix setupOnce extension to properly work with healthcheck tests

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/BaseSpecification.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/BaseSpecification.groovy
@@ -8,6 +8,7 @@ import org.openkilda.functionaltests.extension.fixture.SetupOnce
 import org.openkilda.functionaltests.extension.healthcheck.HealthCheck
 import org.openkilda.functionaltests.helpers.FlowHelper
 import org.openkilda.functionaltests.helpers.PathHelper
+import org.openkilda.functionaltests.helpers.TopologyHelper
 import org.openkilda.functionaltests.helpers.Wrappers
 import org.openkilda.messaging.info.event.IslChangeType
 import org.openkilda.model.SwitchId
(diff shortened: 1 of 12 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/resilience/StormLcmSpec.groovy
```diff
@@ -47,30 +64,29 @@ class StormLcmSpec extends BaseSpecification {
...
+            northbound.validateFlow(flow.id).each { direction ->
+                assert direction.discrepancies.findAll { it.field != "meterId" }.empty
+            }
         }
 
         and: "Flow can be updated"
-        flowHelper.updateFlow(flow.id, flow.tap { it.source.vlanId++ })
-        northbound.validateFlow(flow.id).each { direction ->
+        def flowToUpdate = flows[0]
+        flowHelper.updateFlow(flowToUpdate.id, flowToUpdate.tap { it.source.vlanId++ })
```
ylobankov: It looks like such a VLAN update can lead to conflicts. Can we update flow more gracefully to avoid it?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/BaseSpecification.groovy
```diff
@@ -50,6 +51,9 @@ class BaseSpecification extends SpringSpecification implements SetupOnce {
     @Autowired
     FlowHelper flowHelper
 
+    @Autowired
+    TopologyHelper topoHelper
```
ylobankov: It is not really important but why not topologyHelper? It's not so long name.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/resilience/StormLcmSpec.groovy
rtretyak: nice catch

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/BaseSpecification.groovy
rtretyak: renamed

PR #2213, opened 2019-03-29
Add performance-tests skeleton and example test

rtretyak:


services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/config/TopologyConfig.java
```diff
@@ -0,0 +1,57 @@
+/* Copyright 2018 Telstra Open Source
```
andriidovhan: 2018 -> 2019

services/src/performance-tests/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+                                    </includes>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
+</project>
```
andriidovhan: Please, add a new line at the end of the file.

services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/extension/healthcheck/HealthCheckExtension.groovy
```diff
@@ -1,16 +1,18 @@
 package org.openkilda.functionaltests.extension.healthcheck
 
+import org.openkilda.functionaltests.extension.FeatureOrderExtension
```
ylobankov: To be consistent with `perftests` it should be named `functests`. Or we can use `performancetests`, `functionaltests`.

rtretyak: changed to performancetests

PR #2234, opened 2019-04-08
Bolts in fl-router are stateless in terms of storm

timofei-durakov:
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2234)


surabujin:
I would prefer to use "external" tick events... Use inheritor of AbstractTick to create MonotonicTick and route its output into DiscoveryBolt.

timofei-durakov:
> I would prefer to use "external" tick events... Use inheritor of AbstractTick to create MonotonicTick and route its output into DiscoveryBolt.

Not a problem, will do it in a separate PR

PR #2236, opened 2019-04-08
Fix for network dump issue in floodlight router

timofei-durakov:
Router passes key field into tuple based on a direction of the stream.
Since request from speaker generates network dump, it shouldn't try to
pass key field to speaker. This patch address this issues

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/floodlightrouter/bolts/DiscoveryBolt.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/floodlightrouter/bolts/DiscoveryBolt.java
@@ -145,11 +145,11 @@ public class DiscoveryBolt extends AbstractTickStatefulBolt<InMemoryKeyValueStat
 
 
     @Override
-    public void send(Message message, String outputStream) {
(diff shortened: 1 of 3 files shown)


services/wfm/src/main/java/org/openkilda/wfm/topology/floodlightrouter/bolts/DiscoveryBolt.java
```diff
@@ -145,11 +145,11 @@ public void prepare(Map map, TopologyContext topologyContext, OutputCollector ou
 
 
     @Override
-    public void send(Message message, String outputStream) {
+    public void send(Message message, String outputStream, boolean lookupKey) {
```
IlyaMoiseev: Looks like lookupKey param now always false

timofei-durakov: since it's public it's OK, need to revisit this while fixing kafka keys, IMHO

surabujin:
I would prefer to use "external" tick events... Use inheritor of AbstractTick to create `MonotonicTick` and route its output into `DiscoveryBolt`.

PR #2241, opened 2019-04-09
Refactoring of VerificationPacket class

niksv:
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2241)

### DIFF
--- a/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/pathverification/VerificationPacket.java
+++ b/services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/pathverification/VerificationPacket.java
@@ -15,107 +15,37 @@
 
 package org.openkilda.floodlight.pathverification;
 
+import lombok.AllArgsConstructor;
+import lombok.Builder;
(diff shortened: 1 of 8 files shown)


services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/pathverification/VerificationPacket.java
```diff
@@ -185,76 +111,4 @@ public IPacket deserialize(byte[] data, int offset, int length) {
...
-                return false;
-            }
-        } else if (!ttl.equals(other.ttl)) {
-            return false;
-        }
-        return true;
-    }
-
-    @Override
-    public String toString() {
```
nikitamarchenko: Are you sure we don't use that in log files and kibana dashboards?

niksv: I didn't found any place where we are using Verification Packet in logs. So I guess it wouldn't effect on Kibana

Issue #2043, opened 2019-02-13
Trying to sync meters for unmetered flow leads to response timeout


rtretyak:
1. Create a flow with bandwidth:0, ignoreBandwidth: true
2. call `PATCH /api/v1/flows/:flowId/meters` for the flow
**Expected:** No solid requirements here, but we should at least not hang. I would expect a `200 OK` or an error that the given flow is 'unmetered'
**Actual:** System fails to respond after 30 seconds of waiting.
```
error processing message 'CommandMessage{timestamp=1550056286148, correlation_id=6b6f1f36-14e1-4a41-8ce2-18de2d4663be, destination=null, payload=MeterModifyCommandRequest(fwdSwitchId=00:00:00:00:00:00:00:03, fwdMeterId=null, rvsSwitchId=00:00:00:00:00:00:00:02, rvsMeterId=null, bandwidth=0)}'
--
```
```
  java.lang.NullPointerException: null
	at org.openkilda.floodlight.kafka.RecordHandler.doModifyMeterRequest(RecordHandler.java:926)
	at org.openkilda.floodlight.kafka.RecordHandler.handleCommand(RecordHandler.java:203)
	at org.openkilda.floodlight.kafka.RecordHandler.doControllerMsg(RecordHandler.java:142)
	at org.openkilda.floodlight.kafka.RecordHandler.parseRecord(RecordHandler.java:1015)
	at org.openkilda.floodlight.kafka.RecordHandler.run(RecordHandler.java:1024)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

andriidovhan:
Auto test has been created https://github.com/telstra/open-kilda/pull/2048
Please remove the Ignore annotation when the issue is fixed.

PR #2245, opened 2019-04-10
Add error response for sync meters on umetered flow

Levitan:
- Add error response
- Complete functional test for this case

Closes: #2043


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/MetersSpec.groovy
```diff
@@ -457,7 +455,10 @@ meters in flow rules at all (#flowType flow)"() {
 
         when: "Resetting meter burst and rate to default"
         northbound.resetMeters(flow.id)
-//        TODO(andriidovhan) add then and check response: body or message
```
andriidovhan: please, add an empty line between the `when` and `then` blocks

Levitan: Resolved

PR #2246, opened 2019-04-11
Added new manual field to flow model

timofei-durakov:
manual field is used to forbid auto-reroutng flows


surabujin:
Can we use "pinned" flow term instead of "manual"?

Issue #2255, opened 2019-04-12
Flow stats issues


rtretyak:
Steps:
1. Create a flow
2. Pick a switch. Dump rules and remember cookie ids for forward and reverse paths
3. Simulate 2 FlowStats entries for that switch (for fw and reverse cookies) with packetCount and byteCount equal to `Long.MAX_VALUE`
4. Dump stats for created flowid for all flow-related metrics:
```
["flow.bytes", "flow.packets", "flow.ingress.bytes", "flow.ingress.packets",
"flow.raw.packets", "flow.raw.bytes", "flow.bits", "flow.ingress.bits", "flow.raw.bits"]
```
**Expected:** to be discussed
**Actual:** Observe negative values saved for metrics 
"flow.raw.packets": -2, 
"flow.raw.bytes": -2, 
"flow.bits": -8, 
"flow.ingress.bits": -8, 
"flow.raw.bits": -8

rtretyak:
Discussed with team.
For noviflow we assume that packet counter will always roll over BEFORE byte count, so not testing  Long.MAX_VALUE for bytes. That's fine, since packets should be 4G+ for bytes to roll over before packets on Novis, which is unlikely. Actually, we are not able to properly handle 2^64 bytes at this point, for reasons:
            1. OTSDB itself is only capable of storing up to 2^63, while switch bytes are up to 2^64
            2. Our default Long implementation in Java backend is also capable of only 2^63
            3. Even for 2^63 we get overflowed 'negative' values when converting bytes to bits (doing bytesx8)
Leaving the issue open, since the problem exists, but lowering the priority, because it won't be affecting us any time soon.

PR #2256, opened 2019-04-12
Add test for stats max values. Bump spock to 1.3

rtretyak:
reproduces issue #2255


rtretyak:
No longer reproduces #2255, issue updated with more detailed comments

PR #2260, opened 2019-04-15
Add fields to SwitchDto from database model.

dpoltavets:
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2260)


services/src/northbound-service/nb-api/src/main/java/org/openkilda/northbound/dto/v1/switches/SwitchDto.java
```diff
@@ -42,4 +42,19 @@
 
     @JsonProperty("under_maintenance")
     private boolean underMaintenance;
+
+    @JsonProperty(value = "of_version")
```
niksv: nit: I think it's better to remove "value" to be consistent with other fields

dpoltavets: Fixed.

PR #2269, opened 2019-04-17
ADD: System is able to reroute(intentional) flow with protected according to the priority field

andriidovhan:
ADD: System is able to reroute(intentional) flow with protected path in the correct order based on the priority field

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowPriorityRerouteSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowPriorityRerouteSpec.groovy
@@ -10,10 +10,17 @@ import org.openkilda.messaging.payload.flow.FlowPayload
 import org.openkilda.messaging.payload.flow.FlowState
 import org.openkilda.testing.model.topology.TopologyDefinition.Switch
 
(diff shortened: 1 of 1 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowPriorityRerouteSpec.groovy
```diff
@@ -56,7 +64,10 @@ class FlowPriorityRerouteSpec extends BaseSpecification {
         }
 
         and: "Reroute procedure was done based on the priority field"
-        flows.sort { it.priority }*.id == northbound.getAllFlows().sort { it.lastUpdated }*.id
+        int delay = protectedPath ? rerouteDelay * 2 : 1
```
rtretyak: I absolutely don't understand what happens here

andriidovhan: I will use `WAIT_OFFSET` instead of the `delay` variable

andriidovhan: done

PR #2273, opened 2019-04-17
Improve fetching of Flow sub-entities (FlowPath and PathSegments) in DAO

sergii-iakovenko:
The changes:
- Improve fetching of FlowPath and PathSegments entities from Neo4J in FlowRepository and FlowPathRepository.

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2273)


sergii-iakovenko:
The changes don't increase DAO performance to the acceptable level: dumping of flows is still much slower than it was with the old data model. Require further improvements.

Issue #2053, opened 2019-02-15
Broadcast MAC address for a discovery packet should be configurable


pzakatov:
`VERIFICATION_BCAST_PACKET_DST` should be configurable. 
The default should be `01:80:C2:00:00:00`
Reference: https://standards.ieee.org/products-services/regauth/grpmac/public.html

PR #2276, opened 2019-04-18
#2053 Make broadcast MAC address for a discovery packet configurable.

dpoltavets:
Closes: #2053.


services/src/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -178,6 +178,8 @@
     private ConnectModeRequest.Mode connectMode;
     private SwitchManagerConfig config;
 
+    private String verificationBcastPacketDst;
```
niksv: nit: You can use MacAddress instead of String because it used as String nowhere

niksv: Plus it's a kind of validation

Issue #2266, opened 2019-04-16
Getting 'speed=0' on an ISL under certain conditions


rtretyak:
Sometimes after blinking ports on a centec switch I end up with a link discovered with a 0 speed. @surabujin tracked down that the Floodlight itself returns the correct speed and we lose it somewhere in the middle. Reproducing only on hardware switch, was not able to catch it on virtual OVS.
Reproducing script: branch `functional/reproduce_`, command `mvn test -Pfunctional -Dtest=LinkSpec#Get all flows (UP/DOWN) going through a particular link`

rtretyak:
We can lower the priority if prove that the issue is due to a broken hardware or at least locate the issue

PR #2286, opened 2019-04-19
Log discovery data received by speaker

surabujin:
Log discovery data received from speaker to be able to trace
invalid/corrupted data.

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2286)


surabujin:
Required to trace issue #2266

PR #2215, opened 2019-03-29
Switch synchronize action

IlyaMoiseev:
Closes #2189 
Depends on #2289

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2215)

PR #2289, opened 2019-04-22
SwitchManager topology refactoring

IlyaMoiseev:
Preparation for #2215 
Rules validate action uses switch validate FSM
Eliminate duplicate logic in switch validate and switch sync FSMs
Query rules and meters in parallel for switch validation
Consolidate both FSM in one bolt with parallelism 2
Add unit tests for FSMs services 

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2289)


timofei-durakov:
overall looks good.  2 points:
- add unit tests for service layer classes, check network topology for samples.
- I'm not sure if it's ValidateRequest(with sync flag) of SyncRequest(with smth like dry-run flag). Worth discussing offline.

Issue #2073, opened 2019-02-21
Fix "CorrelationId was not sent, generated one" warning from "org.openkilda.northbound.utils.RequestCorrelationFilter"


pzakatov:
These logs are flooding

PR #2310, opened 2019-04-26
Fix the spam of warnings "CorrelationId was not sent, generated one".

dpoltavets:
Closes: #2073.


nikitamarchenko:
We need to find who send a request without correlation-id and fix that but not just mask that message!

timofei-durakov:
> We need to find who send a request without correlation-id and fix that but not just mask that message!

do you think it's possible? AFAIR we agreed to make the correlation id mandatory for V2, and reduce the log level for v1

PR #2315, opened 2019-04-30
Add flow create API V2

nikitacherevko:
- add models and controller
- add lombok.config: now mapstruct is able to convert object into immutable object (with @Value annotation)

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2315)


services/src/northbound-service/nb-api/src/main/java/org/openkilda/northbound/dto/v2/flows/FlowEndpointV2.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import lombok.Value;
+
+@Value
+@JsonNaming(value = SnakeCaseStrategy.class)
+public class FlowEndpointV2 {
+    SwitchId switchId;
+    Integer portNumber;
```
sergii-iakovenko: What is the reason to use a wrapper class instead of primitive here?

PR #2328, opened 2019-05-08
Pinned flow

timofei-durakov:
Closes:  #2334

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2328)


services/src/kilda-core/kilda-model/src/main/java/org/openkilda/model/PathSegment.java
```diff
@@ -78,6 +78,8 @@
 
     private Long latency;
 
+    private boolean failed = false;
```
sergii-iakovenko: Shouldn't we add it to constructor, so the copying builder copies the value? Also, shouldn't it be a enum like SegmentStatus?

PR #2330, opened 2019-05-08
extend existing 'Unable to create a flow on an isl port' test

andriidovhan:
extend existing _'Unable to create a flow on an isl port'_ test


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowCrudSpec.groovy
```diff
@@ -507,6 +510,58 @@ class FlowCrudSpec extends BaseSpecification {
         ]
     }
 
+    def "Unable to create a flow on an isl port when ISL status is FAILED"() {
+        given: "An inactive isl with failed state"
+        Isl isl = topology.islsForActiveSwitches.find { it.aswitch && it.dstSwitch }
+        assert isl
```
rtretyak: This is fine, but usually we use assumptions to check whether topology has the required elements for the test.

andriidovhan: noted
and fixed

PR #2347, opened 2019-05-13
Add functional tests for BFD feature

rtretyak:
Also remove 'bfd' flag from TopologyDefinition and topology.yaml. Now by default we assume NO bfd sessions in the system. Any BFD session is created on demand.
New pre-condition: all switches should have proper virtual ports created beforehand.


services/src/test-library/src/main/java/org/openkilda/testing/service/northbound/NorthboundServiceImpl.java
```diff
@@ -430,6 +438,13 @@ public BatchResults deleteLinkProps(List<LinkPropsDto> keys) {
         return Arrays.asList(updatedLink);
     }
 
+    @Override
+    public List<LinkDto> setLinkBfd(LinkEnableBfdDto link) {
```
andriidovhan: Would be great if you add logging for this request
`log.debug("Setting BDF ....");`

Issue #647, opened 2018-04-21
use Provider Backbone Bridges (PBB) instead of VLAN


pzakatov:

pzakatov:
Proposal: make transport mechanism configurable (per flow?, per contoller?, etc.)

PR #2350, opened 2019-05-14
Add an ability to choose encapsulation type in the CRUD flow over REST API.

dpoltavets:
Part of: #647.


services/src/messaging/src/main/java/org/openkilda/messaging/model/FlowDto.java
```diff
@@ -182,6 +186,7 @@ public FlowDto() {
      * @param maxLatency        max latency
      * @param priority          flow priority
      * @param pinned            pinned flag
+     * @param encapsulationType flow encapsulation type
```
niksv: Should we add it to equals and hashCode? Or to comment above hashcode?

dpoltavets: Added.

PR #2379, opened 2019-05-21
fix tests for protected path according to staging env

andriidovhan:
fix tests for protected path according to staging environment


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/ProtectedPathSpec.groovy
```diff
@@ -721,10 +722,17 @@ class ProtectedPathSpec extends BaseSpecification {
             involvedTransitSwitches.each { switchId ->
                 def amountOfRules = (switchId in currentProtectedPath*.switchId && switchId in currentPath*.switchId)
                         ? 4 : 2
-                def switchValidateInfo = northbound.validateSwitch(switchId)
-                assert switchValidateInfo.rules.proper.size() == amountOfRules
-                switchHelper.verifyRuleSectionsAreEmpty(switchValidateInfo, ["missing", "excess"])
-                switchHelper.verifyMeterSectionsAreEmpty(switchValidateInfo)
+                if (northbound.getSwitch(switchId).description.contains("OF_12")) {
```
ylobankov: As an alternative option, we can choose only OF_13 switches and leave it how it was before.

PR #2432, opened 2019-05-31
Tests pinned + protected flow

andriidovhan:
- add tests related to a pinned+protected flow
- and add verification: check that system marks flow as DOWN when flow path is partially broken

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/PinnedFlowSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/PinnedFlowSpec.groovy
@@ -8,21 +8,27 @@ import org.openkilda.functionaltests.BaseSpecification
 import org.openkilda.functionaltests.helpers.Wrappers
 import org.openkilda.messaging.error.MessageError
 import org.openkilda.messaging.info.event.IslChangeType
+import org.openkilda.messaging.info.event.PathNode
 import org.openkilda.messaging.payload.flow.FlowState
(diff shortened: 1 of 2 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/PinnedFlowSpec.groovy
```diff
@@ -71,12 +80,20 @@ class PinnedFlowSpec extends BaseSpecification {
         cookiesMap.sort() == cookiesMapAfterReroute.sort()
         metersMap.sort() == metersMapAfterReroute.sort()
 
-        when: "The broken ISL is restored"
-        northbound.portUp(islToBreak.srcSwitch.dpId, islToBreak.srcPort)
+        when: "The broken ISLs are restored one by one"
+        islsToBreak[0..-2].each { islToBreak ->
+            northbound.portUp(islToBreak.srcSwitch.dpId, islToBreak.srcPort)
+            Wrappers.wait(WAIT_OFFSET + discoveryInterval) {
+                assert islUtils.getIslInfo(islToBreak).get().state == IslChangeType.DISCOVERED
```
rtretyak: There is a reroute delay after ISL is discovered, so these check may not catch the problem if it appears

andriidovhan: fixed, added: `TimeUnit.SECONDS.sleep(rerouteDelay - 1)`

PR #2464, opened 2019-06-06
Fix switch validate and rules sync on meters unsupported switches

### LINKED ISSUE #2453: Sync switch rules operation fails for OF_12 switches
Trying to sync rules on OF_12 switch results in error
```
{
    "correlation_id": "f67fa316-413e-4e8d-8975-227615a5558c",
    "timestamp": 1559731780092,
    "error-type": "Invalid request parameters",
    "error-message": "Meters are not supported on switch $switchId because of OF version OF_12",
    "error-description": "Not supported: $switchId"
}
```

**Expected:** Sync __rules__ operation should not try to do anything with meters (only sync __switch__ should). Sync rules for OF_12 switch should not end with 'meters' error

IlyaMoiseev:
Closes #2453

### DIFF
--- a/services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
+++ b/services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
@@ -89,6 +89,7 @@ import org.openkilda.messaging.info.discovery.DiscoPacketSendingConfirmation;
 import org.openkilda.messaging.info.meter.FlowMeterEntries;
 import org.openkilda.messaging.info.meter.MeterEntry;
 import org.openkilda.messaging.info.meter.SwitchMeterEntries;
+import org.openkilda.messaging.info.meter.SwitchMeterUnsupported;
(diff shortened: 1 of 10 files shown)


services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/kafka/RecordHandler.java
```diff
@@ -969,13 +970,11 @@ private void dumpMeters(SwitchId switchId, String correlationId, String replyToT
             producerService.sendMessageAndTrack(replyToTopic, correlationId, infoMessage);
         } catch (UnsupportedSwitchOperationException e) {
             String messageString = "Not supported: " + switchId;
-            logger.error(messageString, e);
```
niksv: I think we should left logger

dpoltavets: +1

IlyaMoiseev: This case is not error anymore, add info level logging

PR #2471, opened 2019-06-09
RTL Part 3: Handle RTL packet in Floodlight

niksv:
Floodlight gets RTL packet and measure round trip latency,
but does not send it to any topology (will be implemented in next commit)

DB still stores one way latency

Related-To #580

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2471)

### DIFF
--- a/services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/model/OfInput.java
+++ b/services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/model/OfInput.java
@@ -68,12 +68,12 @@ public class OfInput {
     }
 
     /**
-     * Check current packet(PACKET_IN) cookie value is mismatched with expected value.
+     * Check current packet(PACKET_IN) cookie value is mismatched with all expected values.
      */
-    public boolean packetInCookieMismatch(U64 expected, Logger log) {
+    public boolean packetInCookieMismatchAll(Logger log, U64... expected) {
         boolean isMismatched = packetInCookieMismatchCheck(expected);
         if (isMismatched) {
-            log.debug("{} - cookie mismatch (expected:{} != actual:{})", this, expected, packetInCookie());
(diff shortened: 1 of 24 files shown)


services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/model/OfInput.java
```diff
@@ -107,7 +118,7 @@ public String toString() {
         return String.format("%s ===> %s.%s:%d", dpId, message.getType(), message.getVersion(), message.getXid());
     }
 
-    private boolean packetInCookieMismatchCheck(U64 expected) {
+    private boolean packetInCookieMismatchCheck(U64... expectedArray) {
```
timofei-durakov: Ugly naming:-P

services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/pathverification/PathVerificationService.java
```diff
@@ -469,10 +510,28 @@ static long noviflowTimestamp(byte[] timestamp) {
...
+     * @param switchId Switch ID of endpoint which received packet
+     * @param port port of endpoint which received packet
+     * @param switchT0 sent packet timestamp in nanoseconds
+     * @param switchT1 receive packet timestamp in nanoseconds
+     * @return long
+     */
+    @VisibleForTesting
+    static long calculateRoundTripLatency(SwitchId switchId, int port, long switchT0, long switchT1) {
+        if (switchT0 <= 0 || switchT1 <= 0) {
+            logger.warn("Invalid round trip latency timestamps on endpoint {}_{}. t0 = '{}', t1 = '{}'.",
```
timofei-durakov: Do we need to handle this in a special case on the storm side. Need to figure out

niksv: will handle this case on storm side

PR #2472, opened 2019-06-10
Storm side vxlan rules

timofei-durakov:
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2472)


services/wfm/src/main/java/org/openkilda/wfm/topology/network/controller/sw/SwitchFsm.java
```diff
@@ -57,6 +59,7 @@
 
     private final TransactionManager transactionManager;
     private final SwitchRepository switchRepository;
+    private final SwitchFeaturesRepository switchFeaturesRepository;
```
dpoltavets: This repository is never used.

PR #2477, opened 2019-06-10
Add test that slowly discovers switches one by one

rtretyak:
-Change templates and make floodlight controller property to be a list


services/src/atdd-staging/src/main/java/org/openkilda/atdd/staging/config/TopologyConfig.java
```diff
@@ -57,11 +56,8 @@ public TopologyDefinition topologyDefinition() throws IOException {
 
         topologyDefinition.setBfdOffset(bfdOffset);
         topologyDefinition.setControllers(controllerHosts);
-        // TODO(tdurakov): it should be possible to reproduce env by dumping existing topology or changing this random
-        // pick
         for (TopologyDefinition.Switch sw : topologyDefinition.getSwitches()) {
-            int pick = new Random().nextInt(controllerHosts.size());
-            sw.setController(managementController + " " + statController);
+            sw.setController(managementController);
```
andriidovhan: Did you intentionally miss the 'statController'?

rtretyak: No, will add

PR #2481, opened 2019-06-11
set correct value for a statsRouterInterval variable in the mflStatsSpec file

andriidovhan:
set correct value for a statsRouterInterval variable in the mflStatsSpec file


andriidovhan:
should be investigated more deeply

andriidovhan:
need more time for investigation

Issue #2391, opened 2019-05-23
Add db transaction repeat into network topology


surabujin:
Due to data update nature there network topology can make attempt to create already existed records in DB. DB will fail with constraint violation error. So network topology should make one more attempt (this attempt will have other precondition so it can completes with success).

Depends on #2296

rtretyak:
May result in pretty severe consequences and one of them is ISL not properly changing status FAILED or DISCOVERED

PR #2493, opened 2019-06-12
Repeat db transation on db locks

surabujin:
Do DB transaction repeat on transient errors.

Issue #2391


services/src/kilda-core/kilda-persistence-neo4j/src/main/java/org/openkilda/persistence/Neo4jTransactionManager.java
```diff
@@ -79,8 +86,16 @@ public Session getSession() {
         execute(retryPolicy, callableOf(action));
     }
 
+    @Override
+    public RetryPolicy makeRetryPolicyBlack() {
```
nikitacherevko: Did you really want it to be black? :)

surabujin: actually not... making fix

PR #2495, opened 2019-06-12
add possibility to get transit vlan for a flow via database

andriidovhan:
add possibility to get transit vlan for a flow via database

### DIFF
--- a/services/src/test-library/src/main/java/org/openkilda/testing/service/database/Database.java
+++ b/services/src/test-library/src/main/java/org/openkilda/testing/service/database/Database.java
@@ -18,9 +18,12 @@ package org.openkilda.testing.service.database;
 import org.openkilda.messaging.info.event.PathInfoData;
 import org.openkilda.model.Flow;
 import org.openkilda.model.MeterId;
+import org.openkilda.model.PathId;
 import org.openkilda.model.SwitchId;
+import org.openkilda.model.TransitVlan;
 import org.openkilda.testing.model.topology.TopologyDefinition.Isl;
 
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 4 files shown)


services/src/test-library/src/main/java/org/openkilda/testing/service/database/DatabaseSupportImpl.java
```diff
@@ -300,6 +305,18 @@ public Flow getFlow(String flowId) {
         return flowRepository.findById(flowId).get();
     }
 
+    /**
+     * Get flowPair.
```
rtretyak: fix description

services/src/test-library/src/main/java/org/openkilda/testing/service/database/DatabaseSupportImpl.java
```diff
@@ -300,6 +305,18 @@ public Flow getFlow(String flowId) {
...
+     *
+     * @param forwardPathId forward path Id
+     * @param reversePathId reverse path Id
+     * @return Collection  of TransitVlan
```
rtretyak: afaik our checkstyle policy does not force you to write this kind of useless docs. If you have nothing to add - do not add. This kind of doc is just a distraction and waste of space. You'd better describe when it returns 1 vlan vs cases when it returns 2 vlans

services/src/test-library/src/main/java/org/openkilda/testing/service/database/Database.java
```diff
@@ -54,6 +57,8 @@
 
     Flow getFlow(String flowId);
 
+    Collection<TransitVlan> getTransitVlan(PathId forwardPathId, PathId reversePathId);
```
rtretyak: since it returns a _collection_, the method name should be in plural: getTransitVlans

andriidovhan: fixed

services/src/test-library/src/main/java/org/openkilda/testing/service/database/DatabaseSupportImpl.java
andriidovhan: description was updated

Issue #2513, opened 2019-06-18
Represent protected path status in API


IlyaMoiseev:
Design in #2486

PR #2518, opened 2019-06-20
Add protected path status to API.

dpoltavets:
- added the `status-details` field to FlowResponsePayload;
- replaced FlowPayload with FlowResponsePayload in FlowController;
- added `DEGRADED` flow status and changed logic for flow status computing with that status.

Closes: #2513.

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2518)

### DIFF
--- a/services/src/messaging/src/main/java/org/openkilda/messaging/payload/flow/FlowResponsePayload.java
+++ b/services/src/messaging/src/main/java/org/openkilda/messaging/payload/flow/FlowResponsePayload.java
@@ -37,6 +37,9 @@ public class FlowResponsePayload extends FlowPayload {
     @JsonProperty("diverse_with")
     private List<String> diverseWith;
 
+    @JsonProperty("status-details")
+    private FlowStatusDetailsPayload flowStatusDetails;
+
     /**
      * Instance constructor.
      *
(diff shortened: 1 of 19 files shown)


services/src/messaging/src/main/java/org/openkilda/messaging/payload/flow/FlowResponsePayload.java
```diff
@@ -37,6 +37,9 @@
     @JsonProperty("diverse_with")
     private List<String> diverseWith;
 
+    @JsonProperty("status-details")
```
niksv: why '-' instead of '_'?

services/wfm/src/main/java/org/openkilda/wfm/share/mappers/FlowMapper.java
```diff
@@ -63,6 +64,10 @@
     @Mapping(source = "timeModify", target = "lastUpdated")
     @Mapping(source = "timeCreate", target = "createdTime")
     @Mapping(source = "pinned", target = "pinned")
+    @Mapping(target = "flowStatusDetails",
+            expression = "java(flow.getFlow().isAllocateProtectedPath() ? "
+                    + "new FlowStatusDetails(flow.getFlow().getMainFlowPrioritizedPathsStatus(), "
+                    + "flow.getFlow().getProtectedFlowPrioritizedPathsStatus()) : null)")
```
niksv: Do we have test of this map?

services/src/messaging/src/main/java/org/openkilda/messaging/payload/flow/FlowResponsePayload.java
dpoltavets: This field has been made like other fields in this class.

services/wfm/src/main/java/org/openkilda/wfm/share/mappers/FlowMapper.java
dpoltavets: Yes, I'll add test.

dpoltavets: Done.

dpoltavets:
___
*[services/src/messaging/src/main/java/org/openkilda/messaging/payload/flow/FlowStatusDetailsDto.java, line 27 at r1](https://reviewable.io/reviews/telstra/open-kilda/2518#-LhuCuhU3SBLd5dQsDNm:-Li7JFOV6--7qfRPrZnL:bt1b19p) ([raw file](https://github.com/telstra/open-kilda/blob/c4dafb17db259bed844ca4943b571227ed001077/services/src/messaging/src/main/java/org/openkilda/messaging/payload/flow/FlowStatusDetailsDto.java#L27)):*
<details><summary><i>Previously, IlyaMoiseev (Ilya Moiseev) wrote…</i></summary><blockquote>

It uses only in response payload and should be named FlowStatusDetailsPayload
</blockquote></details>

Renamed.

PR #2537, opened 2019-06-25
BFD port management design update

surabujin:
The goal is to add LCM of logical BFD ports into existing BFD management
process.

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2537)


surabujin:
There are no available resources to work on this now. Keeping produced artefacts as a git branch.

PR #2541, opened 2019-06-26
Fix correlation id checks for swagger ui

### LINKED ISSUE #2515: Auto-generated correlation id is not visible
Once Northbound generate correlation id to request it's impossible to track it next to kilda, since no message with that correlation_id been logged

nikitacherevko:
Closes #2515
- Added possibility to pass correlation id using swagger ui
- Fixed generating correlation id while accessing to swagger resources


services/src/northbound-service/northbound/src/main/java/org/openkilda/northbound/utils/RequestCorrelationFilter.java
```diff
@@ -48,7 +48,11 @@
     private static final Logger logger = LoggerFactory.getLogger(RequestCorrelationFilter.class);
 
     private static final List<String> EXCLUDE_PATTERNS = ImmutableList.of(
-            "/v1/health-check"
+            "/v1/health-check",
+            // swagger related patterns
+            "/swagger*/**",
```
timofei-durakov: nit: there are the same exclude patterns for RequestCorrelationFilter as well, Maybe it's worth to move it in constants

PR #2549, opened 2019-06-27
Fix for swagger correlation id header name

timofei-durakov:


services/src/northbound-service/northbound/src/main/java/org/openkilda/northbound/config/SwaggerConfig.java
```diff
@@ -15,6 +15,8 @@
 
 package org.openkilda.northbound.config;
 
+import static org.openkilda.messaging.Utils.CORRELATION_ID;
```
surabujin: Unexpected location.

Issue #2542, opened 2019-06-26
Configurable pings


IlyaMoiseev:
**Ping options:**
- Total packet size from min to 9216 bytes ( where min is the minimal size based on all headers and meta)
- Packet count (1 - 1million, default: 10)
- Packet interval parameter: in seconds (default is 1, range 1-10) 
- On the interval >= 1 seconds
- Test time (0-24 hours)
- Flag: do not fragment

Make ping service (start, status(emit/receive count, latency), stop, etc)
Only one ping per flow
Realtime ping progress (remaining time or count)
Multiple pings on different flows
History (summary only)

sergii-iakovenko:
The following changes were staged in https://github.com/telstra/open-kilda/tree/feat/configurable_pings_fl_extension and require review:
- Rework ping packet serialization to reduce minimum packet size and configurable packet size support. Minimum ping packet size is now 78 bytes (32 bytes payload)
- Add ability to set do not fragment L3 packet flag

PR #2509, opened 2019-06-14
Configurable ping design

IlyaMoiseev:
Design for #2542

This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2509)

PR #2550, opened 2019-06-28
FL ping packet size and don't fragment flag extensions

IlyaMoiseev:
Part of #2542 #2509 

- Rework ping packet serialization to reduce minimum packet size and configurable packet size support. Minimum ping packet size is now 78 bytes (32 bytes payload)
- Add ability to set `do not fragment` L3 packet flag


sergii-iakovenko:
The changes requires review / rework - to be done in #2542

PR #2551, opened 2019-06-28
Remove old switch rules validate code

IlyaMoiseev:


niksv:
PR removes code which doesn't exist now

PR #2552, opened 2019-06-28
[WIP] Save port status into DB.

Settis:


rozdy:
Port history feature is already done by PR #2718

PR #2559, opened 2019-07-01
Document of custormer traffic encapsulation transormation

surabujin:


surabujin:
No need in this PR because some of it's content already merged into mainstream other part has been fully implemented(not POC) and pushed into separate PRs (waits for review).

PR #2568, opened 2019-07-03
fix creating topology for performance test

andriidovhan:


services/src/performance-tests/src/main/groovy/org/openkilda/performancetests/helpers/TopologyHelper.groovy
```diff
@@ -45,20 +45,18 @@ class TopologyHelper extends org.openkilda.functionaltests.helpers.TopologyHelpe
...
             topo.addCasualSwitch("${managementControllers[region]} ${statControllers[region]}") 
         }
-        islsAmount.times {
+        //create links between neighbor switches
+        def allSwitches = topo.getSwitches()
+        for (def i = 0; i < allSwitches.size() - 1; i++) {
+            topo.addIsl(allSwitches[i], allSwitches[i+1])
+        }
+        //create links between random switches
+        islsAmount-topo.getIsls().size().times {
```
rtretyak: I have doubts about this working as you expect. Round braces missing? 
`(islsAmount-topo.getIsls().size()).times {`
Also, this code may throw because you can end up with negative number here `islsAmount-topo.getIsls().size()`

andriidovhan: yep, round braces were missed.
L54 should not fail in case negative number(it should just to return `null`)

services/src/performance-tests/src/main/groovy/org/openkilda/performancetests/helpers/TopologyHelper.groovy
```diff
@@ -40,25 +31,26 @@ class TopologyHelper extends org.openkilda.functionaltests.helpers.TopologyHelpe
     List<String> statControllers
 
     CustomTopology createRandomTopology(int switchesAmount, int islsAmount) {
+        if (islsAmount < (switchesAmount -1)) {
+            throw new Exception("Not enough ISLs were requested. islsAmount should be >= switchesAmount")
```
rtretyak: I thought I can request 3 switches and 2 isls, why not?

andriidovhan: I am with you on it.
fixed

Issue #2574, opened 2019-07-05
[Stats] As a user I can see table stats in OpenTSDB


pzakatov:
In [statsresponse](http://flowgrammable.org/sdn/openflow/message-layer/statsresponse/#ofp_1_3_0) Table message wee need to collect `active-count`, `lookup-count` and `matched-count` per table in a same cadense we do flow stats.

Also, we should record a diff: `lookup-count` minus `matched-count`. Matric name should be `missed`

metric details:
- name: `switch.table.*` without `-count` (e.g. `switch.table.active`, etc.)
  - `switch.table.active`
  - `switch.table.lookup`
  - `switch.table.matched`
  - `switch.table.missed`
- tags: switchid, tableid

PR #2588, opened 2019-07-09
Add ability to load table stats from the switch (#2574)

nikitacherevko:
It is the first stage just to be able to see in logs if have any mismatches of table counters on switches

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2588)

### DIFF
--- a/services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/statistics/StatisticsService.java
+++ b/services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/statistics/StatisticsService.java
@@ -15,6 +15,8 @@
 
 package org.openkilda.floodlight.statistics;
 
+import static java.lang.String.format;
+
 import org.openkilda.floodlight.converter.OfFlowStatsMapper;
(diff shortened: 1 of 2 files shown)


services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/statistics/StatisticsService.java
```diff
@@ -172,6 +180,52 @@ private void gatherFlowStats(IOFSwitch iofSwitch) {
...
+                                            "Lookup {} and matched {} counters are not equal on the switch {} table {}",
+                                            entry.getLookupCount().getValue(), entry.getMatchedCount().getValue(),
+                                            iofSwitch.getId(), entry.getTableId().getValue());
+                                }
+                            });
+                }
+
+                @Override
+                public void onFailure(Throwable t) {
+                    tableStatsLogger.error("Failed to read table stats", t);
```
niksv: I think we should add  SwitchId information

 ` on switch={}", iofSwitch.getId()`

PR #2599, opened 2019-07-16
Filter out noviflow's special OF ports

surabujin:


surabujin:
#2216 already fix this issue.

PR #2606, opened 2019-07-19
add new endpoint for managing iptables rules on floodlight

### LINKED ISSUE #1268: Test framework: Add ability to disconnect a certain hardware switch from controller
As a qa engineer, I want to have an ability to disconnect/connect a certain staging (hardware) switch from the controller programmatically, so that I can test certain Kilda behavior when switch/switches go down.
This can probably be done in the Lock-keeper service.

andriidovhan:
Closes #1268

### DIFF
--- a/services/src/test-library/src/main/java/org/openkilda/testing/service/lockkeeper/LockKeeperVirtualImpl.java
+++ b/services/src/test-library/src/main/java/org/openkilda/testing/service/lockkeeper/LockKeeperVirtualImpl.java
@@ -61,32 +61,20 @@ public class LockKeeperVirtualImpl extends LockKeeperServiceImpl {
     }
 
     @Override
-    public void blockAccessToIp(String ipAddress) {
+    public void blockFloodlightAccessToPort(Integer port) {
         throw new UnsupportedOperationException(
-                "blockAccessToIp method is not available on virtual env");
+                "blockFloodlightAccessToPort method is not available on virtual env");
     }
 
     @Override
-    public void unblockAccessToIp(String ipAddress) {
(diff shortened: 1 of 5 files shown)


services/src/test-library/src/main/java/org/openkilda/testing/service/lockkeeper/LockKeeperVirtualImpl.java
```diff
@@ -59,4 +59,34 @@ public void setController(Switch sw, String controller) {
...
+    public void unblockAccessToPort(Integer port) {
+        throw new UnsupportedOperationException(
+                "unblockAccessToPort method is not available on virtual env");
+    }
+
+    @Override
+    public void allowAccessToEverything() {
+        throw new UnsupportedOperationException(
+                "allowAccessToEverything method is not available on virtual env");
+    }
```
rtretyak: Not the best solution. We should keep virtual and hw implementations as close to each other as possible.
you can
1. to discuss: remove all the block/unblock methods from the interface at all.
2. Add 'hardware' implementation for `reviveSwitch` and `knockoutSwitch`, which will reduce the amount of UnsupportedOperationException-s (instead of increasing it)
3. add `knockoutFloodlight` to the interface. It will have a hardware implementation (through the lock keeper service) and we should think of how this can be implemented for virtual (use local docker, goto fl container and run the same iptables command?)

andriidovhan: it was discussed internally
1. decided to leave blockAccessToPort/unblockAccessToPort
2. done
3. decided not to add this method

PR #2617, opened 2019-07-23
add virtualImpl for managing floodlight access

andriidovhan:


services/src/test-library/src/main/java/org/openkilda/testing/service/lockkeeper/LockKeeperServiceImpl.java
```diff
@@ -53,6 +54,9 @@
     @Autowired
     NorthboundService northbound;
 
+    @Value("#{'${kafka.bootstrap.server}'.split(':')}")
+    private List<String> kafkaBootstrapServer;
```
rtretyak: Why `kafkaBootstrapServer` is a list of strings? this is weird

timofei-durakov: Because it's should be server**s**

PR #2624, opened 2019-07-25
OF cookie management cleanup

surabujin:
Actualize usage and sort out cookie constatns(masks and flags) and
methods.


services/src/kilda-core/kilda-model/src/main/java/org/openkilda/model/Cookie.java
```diff
@@ -28,19 +28,21 @@
...
-    public static final long ROUND_TRIP_LATENCY_RULE_COOKIE = 0x8000000000000006L;
-    public static final long VERIFICATION_UNICAST_VXLAN_RULE_COOKIE = 0x8000000000000007L;
-    public static final long DEFAULT_RULES_MASK = 0x8000000000000000L;
+    public static final long DEFAULT_RULES_FLAG                  = 0x8000_0000_0000_0000L;
+    public static final long FORWARD_FLOW_PATH_FLAG              = 0x4000_0000_0000_0000L;
+    public static final long REVERSE_FLOW_PATH_FLAG              = 0x2000_0000_0000_0000L;
 
-    public static final long FORWARD_FLOW_COOKIE_MASK = 0x4000000000000000L;
-    public static final long REVERSE_FLOW_COOKIE_MASK = 0x2000000000000000L;
+    public static final long DEPRECATED_FLOW_PATH_DIRECTION_FLAG = 0x0080_0000_0000_0000L;
```
sergii-iakovenko: Does it make sense to add "Deprecated" annotation?

services/wfm/src/main/java/org/openkilda/wfm/topology/stats/FlowDirectionHelper.java
```diff
@@ -28,25 +27,30 @@ private FlowDirectionHelper() {}
...
+            direction = Direction.REVERSE;
+        } else {
             throw new FlowCookieException("unknown direction for " + cookie);
         }
-        return direction == FORWARD_FLOW_COOKIE_MASK ? Direction.FORWARD : Direction.REVERSE;
+
+        return direction;
     }
 
     static boolean isKildaCookie(long cookie) {
```
niksv: It is unused now

services/wfm/src/main/java/org/openkilda/wfm/topology/stats/FlowDirectionHelper.java
```diff
@@ -28,25 +27,30 @@ private FlowDirectionHelper() {}
...
      *
-     * @param cookie cookie
+     * @param rawCookie cookie
      * @return flow direction
      */
-    public static Direction findDirection(long cookie) throws FlowCookieException {
-        // Kilda flow first number represents direction with 4 = forward and 2 = reverse
-        // high order nibble represents type of flow with a 2 representing a forward flow
-        // and a 4 representing the reverse flow
-        if (!isKildaCookie(cookie)) {
```
niksv: Is this a legacy check? Do we really can remove it because there is no chance to catch a packet by "not kilda cookie" rule.

services/wfm/src/main/java/org/openkilda/wfm/topology/stats/FlowDirectionHelper.java
surabujin: Didn't get. What "unused now"?

services/wfm/src/main/java/org/openkilda/wfm/topology/stats/FlowDirectionHelper.java
surabujin: No. But the only valid part of this check (mutually exclusive flags combination) I have moved into `Cookie` class. Other "weird" part about "next 20 bits must be zero" I have dropped because we do not have such "requirement".

services/wfm/src/main/java/org/openkilda/wfm/topology/stats/FlowDirectionHelper.java
niksv: You removed using of `isKildaCookie()` method in line 42.

There is no other place where this method is used (except tests). 
So this method can be removed now

surabujin: Now I got it. My fail, removing it.

PR #2639, opened 2019-07-30
Added iptables to kilda-base image

PinGwynn:
Added iptables to kilda-base image and enabled NET_ADMIN capabilities for FL


rtretyak:
Merged this as part of #2631 . Thanks

Issue #2120, opened 2019-03-04
Wrong CookieID showing in Swagger UI


Settis:
Steps to reproduce:
1. Create a flow.
2. Get flow history via Swagger UI `/flows/{flow_id}/history`.

Expected result:
`dumps` contain right `forwardCookie` and `reverseCookie` like `4611686018427387906` and `2305843009213693954`.

Actual result:
The values are `4611686018427388000` and `2305843009213694000`.

**Please note**, that the bug is also appears in `/switches/{switch-id}/rules`.

Root  cause:
The API itself works fine and provides correct data. The bug in UI only when Swagger parse the result and make pretty view.

Possible solutions:
- Make API return cookie as string instead of long.

sergii-iakovenko:
I removed 
> - Update Swagger UI, possible it was fixed.

from the list of possible solutions as the root cause is the way how browsers handle long values in JSONs.

PR #2642, opened 2019-07-30
Add cookie hex representation to get switch rules API

nikitacherevko:
The first step to 'resolve' https://github.com/telstra/open-kilda/issues/2120 .  
In the next PRs we are going to operate newly added cooke-hex field without changing the old one


niksv:
Need to decide do we need it or not

Issue #2635, opened 2019-07-29
Excess meter for unavailable VXlan rule is being installed on Centec switches


rtretyak:
Meter with id `7` is being installed on Centec switch when it connects to Kilda.
This default meter is bind to the `VERIFICATION_UNICAST_VXLAN_RULE_COOKIE = 0x8000000000000007L`, which is not installed and is unavailable on Centec switches.

**Expected behavior:** Do not install `unicast_vxlan` meter on switches, that do not support the corresponding rule.

PR #2645, opened 2019-07-31
Fix for installing meter for vxlan unicast

timofei-durakov:
This patch re-order check supportability and install
actions for vxlan unicast rule.

Closes: #2635


timofei-durakov:
@dpoltavets fixed

PR #2649, opened 2019-07-31
Fix: Ignore default rules when sync rules.

dpoltavets:


services/wfm/src/main/java/org/openkilda/wfm/topology/switchmanager/fsm/SwitchSyncFsm.java
```diff
@@ -75,6 +77,8 @@
     private final CommandBuilder commandBuilder;
     private SwitchId switchId;
     private ValidationResult validationResult;
+    private List<Long> cookiesFromMissingRules;
```
nikitacherevko: I'd prefer to rename it to something like 'missingRulesCookies', but it's not critical at all

PR #2652, opened 2019-07-31
Propagate correlation id into completable future callbacks

surabujin:
Wrap completable future callbacks into correlation-id context so it's
logs can be tracked by correlatiion-id.

Also some attempt to clean-up/normilize naming into FL "commands".

### DIFF
--- a/services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/command/SessionProxy.java
+++ b/services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/command/SessionProxy.java
@@ -16,26 +16,16 @@
 package org.openkilda.floodlight.command;
 
 import org.openkilda.floodlight.error.SwitchWriteException;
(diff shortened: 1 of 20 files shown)


services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/command/SessionProxy.java
```diff
@@ -0,0 +1,40 @@
...
+import org.openkilda.floodlight.service.session.SessionService;
+import org.openkilda.messaging.MessageContext;
+
+import net.floodlightcontroller.core.IOFSwitch;
+import org.projectfloodlight.openflow.protocol.OFMessage;
+
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+public interface SessionProxy {
```
sergii-iakovenko: Could you add javadoc on what the purpose of this interface is?

PR #2657, opened 2019-08-01
fix pinned flow spec

andriidovhan:
sometimes the PinnedFlowSpec fail on jenkins with an error:
```java.lang.IndexOutOfBoundsException: toIndex = 2
org.openkilda.functionaltests.spec.flows.PinnedFlowSpec.System doesn't reroute(automatically) pinned flow when flow path is partially broken(PinnedFlowSpec.groovy:67)```


rtretyak:
I do not understand how this should fix the issue

PR #2662, opened 2019-08-04
Fixed incorrect converting meter rate/burstsize from packets to kilobits

niksv:
We converted packets to kilobytes, not to kilobits.
It causes incorrect meters on Centec and E switches


services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -1853,7 +1853,14 @@ public InetAddress getSwitchIpAddress(IOFSwitch sw) {
 
     @VisibleForTesting
     OFInstructionMeter installMeterForDefaultRule(IOFSwitch sw, long meterId, long ratePkts,
-                                                  ArrayList<OFAction> actionList) {
+                                                  List<OFAction> actionList) {
+        return installOfReinstallMeter(sw, meterId, ratePkts, config.getDiscoPacketSize(),
+                config.getSystemMeterBurstSizeInPackets(), actionList);
+    }
+
+    private OFInstructionMeter installOfReinstallMeter(IOFSwitch sw, long meterId, long rateInPackets,
```
niksv: new method was added because it will be used in connected devices patch

Issue #2319, opened 2019-05-03
Isl/Switch maintenance mode may lead to negative 'cost' values


rtretyak:
Summary: updating link properties for under-maintenance ISL may lead to negative cost being applied after maintenance flag is removed.

**Steps:**
given: property `isl.cost.when.under.maintenance` is set to 10000
1. Set certain ISL 'under maintenance'
   * can be done by either setting the flag directly on ISL or on a related switch
   * observe cost increased by 10000
2. Update link props for that ISL and set its cost to be less than 10000.
3. Unset the maintenance flag
**Expected:** Do not update cost if it is already less than 10000
**Actual:** 10000 is subtracted from a number which is already less than 10k, which leads to negative cost being set for an ISL.

Note that negative cost may lead to unexpected behavior during path computation and is forbidden

rtretyak:
Increasing priority due to too severe consequences in case it fires

PR #2663, opened 2019-08-05
Log message if ISL has negative cost

niksv:
Related to #2319

### DIFF
--- a/services/src/kilda-pce/src/main/java/org/openkilda/pce/AvailableNetworkFactory.java
+++ b/services/src/kilda-pce/src/main/java/org/openkilda/pce/AvailableNetworkFactory.java
@@ -27,6 +27,7 @@ import org.openkilda.persistence.repositories.RepositoryFactory;
 
 import lombok.extern.slf4j.Slf4j;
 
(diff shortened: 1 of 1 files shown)


services/src/kilda-pce/src/main/java/org/openkilda/pce/AvailableNetworkFactory.java
```diff
@@ -105,14 +108,36 @@ public AvailableNetwork getAvailableNetwork(
...
+    }
+
+    private void validateIslsCost(Collection<Isl> isls) {
+        List<Isl> islsWithNegativeCost = isls.stream()
+                .filter(isl -> isl.getCost() < 0)
+                .collect(Collectors.toList());
+
+        if (!islsWithNegativeCost.isEmpty()) {
+            List<String> islMessages = new ArrayList<>();
+            for (Isl isl : islsWithNegativeCost) {
```
sergii-iakovenko: Why don't you build the list with islMessages during the first iteration over isls? The current solution builds excessive intermediate list of isls with negative costs.

PR #2666, opened 2019-08-05
Fix flow create retries

nikitacherevko:
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2666)


services/wfm/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/create/FlowCreateFsm.java
```diff
@@ -92,6 +92,7 @@
     // The amount of flow create operation retries left: that means how many retries may be executed.
     // NB: it differs from command execution retries amount.
     private int remainRetries;
+    private boolean timedOut;
```
niksv: As far as I see it's always True or undefined. Maybe we need `timeout=false` in constructor?

PR #2671, opened 2019-08-06
Produce LLDP packets by traffexam

### LINKED ISSUE #2661: Add ability to send lldp packets via trafgens
As a QA, I want traffgen to allow to send LLDP packets, so that I can simulate a 'connected device'.
AC:
1. Traffgen has endpoint that orders to send an LLDP packet.
2. LLDP packet should be properly received by a connected switch

surabujin:
New REST endpoint aimed to push LLDP packets via some existing
interface(so it can be VLAN tagged if interface have VLAN
assignment).

Endpoint:

PUT /address/<idnr>/lldp

{
  "mac_address": "02:00:00:00:00:01",
  "port_number": "5",
  "time_to_live": 60
}

Norwal response:

{
  "lldp_push": {
    "sent_packets": 1
  }
}

Be aware with used mac addressed - avoid 1 in less significant bit of
first byte. This bit identify "group"/broadcast recipient and filtered
out on any bridge. And because traffexam are attached to target
interface via linux bridge, such packets will never reach recipient.

Close #2661

### DIFF
--- a/services/lab-service/traffexam/kilda/traffexam/action.py
+++ b/services/lab-service/traffexam/kilda/traffexam/action.py
@@ -1,0 +1,61 @@
+# Copyright 2017 Telstra Open Source
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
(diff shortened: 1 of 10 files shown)


services/lab-service/traffexam/kilda/traffexam/action.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def _encode(self, entry):
+        entries = (
+            lldp.LLDPDUChassisID(
+                subtype=lldp.LLDPDUChassisID.SUBTYPE_MAC_ADDRESS,
+                id=entry.mac_address),
+            lldp.LLDPDUPortID(
+                subtype=lldp.LLDPDUPortID.SUBTYPE_LOCALLY_ASSIGNED,
+                id=entry.port_number),
+            lldp.LLDPDUTimeToLive(ttl=entry.time_to_live),
```
niksv: It will be great to have other LLDP field here like managment_adress, system description, etc. But I can add it by myself

PR #2678, opened 2019-08-07
Fix failed flow creation/reroute without paths

nikitacherevko:


services/src/kilda-core/kilda-model/src/main/java/org/openkilda/model/Flow.java
```diff
@@ -409,6 +409,14 @@ public final FlowPath getProtectedReversePath() {
         return getPath(getProtectedReversePathId()).orElse(null);
     }
 
+    public final void removePaths() {
```
sergii-iakovenko: The method name can be treated as removing all paths of the flow, but that's not what it does. Please, reconsider the name.

dpoltavets: +1

nikitacherevko: fixed

PR #2687, opened 2019-08-12
refactor flowHelperV2,flowCrud,swapEndpoint files

andriidovhan:


services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/FlowHelperV2.groovy
```diff
@@ -141,14 +135,7 @@ class FlowHelperV2 {
...
 
-        def flowEntry = null
-        Wrappers.wait(WAIT_OFFSET) {
-            assert northbound.getFlowStatus(flow.flowId).status == FlowState.UP
-
-            flowEntry = db.getFlow(flow.flowId)
-            assert flowEntry
-        }
-        checkRulesOnSwitches(flowEntry, RULES_INSTALLATION_TIME, true)
+        Wrappers.wait(WAIT_OFFSET) { assert northbound.getFlowStatus(flow.flowId).status == FlowState.UP }
```
rtretyak: not enough to just remove `checkRulesOnSwitches`. There should be separate test/verification for v2 api which ensures that 'UP' status is set when and only when all the rules are actually set on all the switches.

andriidovhan: fixed
new test was added `def "System marks flow as UP when and only when all the rules are actually set on all involved switches"() `

services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/FlowHelperV2.groovy
```diff
@@ -172,6 +160,30 @@ class FlowHelperV2 {
...
     }
 
+    /**
+     * Checks flow rules presence (or absence) on all involved switches.
+     */
+    void checkRulesOnSwitches(String flowId, int timeout, boolean rulesPresent) {
+        def flowEntry = db.getFlow(flowId)
+        def cookies = [flowEntry.forwardPath.cookie.value, flowEntry.reversePath.cookie.value]
+        def switches = PathHelper.convert(northbound.getFlowPath(flowEntry.flowId))*.switchId.toSet()
+        switches.each { sw ->
```
rtretyak: This is fine, but as an improvement it would be good to call this in parallel, verify all the switches simultaneously

andriidovhan: will do in next pr

Issue #2582, opened 2019-07-08
Add ability to get a list of devices connected to Flow srt/dst endpoint


niksv:
There are 2 ways to get a list of connected devices: LLDP packets and ARP packets.
Both of them will be used.

Customer's traffic may contain LLDP packets. Kilda must catch such packets to send one copy of a packet to controller and the other copy to Flow endpoint.

API changes:
Following APIs will be extended by `show_connected_devices` field:
`GET /flow`
`PUT /flow`
`GET /flow/{flow_id}`
`PUT /flow/{flow_id}`
`PATCH /flow/{flow_id}`

new Request body will be looks like:

```{
  "flowid": "string",
  "source": {
    "port-id": 0,
    "switch-id": "string",
    "vlan-id": 0
    "show_connected_devices": true
  },
  "destination": {
    "port-id": 0,
    "switch-id": "string",
    "vlan-id": 0,
    "show_connected_devices": true
  },
  "maximum-bandwidth": 0,
  ****
```

With new field `show_connected_devices` user can enable/disable collecting of connection devices for flow source and destination separately 

To get list of connected devices new API will be introduced:
`GET flows/{flow-id}/devices`

Response:
```
{
   "source": [
     { "chassis_id": "string",
        "port_id": "string",
        "ttl": int,
        "system_name": "string",
        "system_description": "string",
        "system_capabilities": "string",
        "management_address": "string"
     },
     {
          ****
     }],
   "destination": [
      { "chassis_id": "string",
        "port_id": "string",
        "ttl": int,
        "system_name": "string",
        "system_description": "string",
        "system_capabilities": "string",
        "management_address": "string"
      },
      {
           ****
      }]
}
```

PR #2688, opened 2019-08-12
Added detectConnectedDevices fields to v1 Flow API

niksv:
Part of #2582

### DIFF
--- a/services/src/kilda-core/kilda-model/src/main/java/org/openkilda/model/Flow.java
+++ b/services/src/kilda-core/kilda-model/src/main/java/org/openkilda/model/Flow.java
@@ -19,6 +19,8 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static java.lang.String.format;
 import static org.neo4j.ogm.annotation.Relationship.OUTGOING;
 
+import org.openkilda.converters.DetectConnectedDevicesConverter;
+
 import lombok.AccessLevel;
 import lombok.Builder;
 import lombok.Data;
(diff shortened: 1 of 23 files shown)


rtretyak:
I don't think it's a good idea to enable this in public API until its working.
Second point: because of enabled v1->v2 mapping for all flow create requests in prod we can't add this feature only to v1.

services/src/kilda-core/kilda-model/src/main/java/org/openkilda/model/Flow.java
```diff
@@ -85,12 +85,24 @@
     @Property(name = "src_vlan")
     private int srcVlan;
 
+    @Property(name = "detect_src_lldp_connected_devices")
+    private boolean detectSrcLldpConnectedDevices;
```
sergii-iakovenko: Please, consider wrapping the new properties into a dedicated class. This class will be mapped to a set of Flow fields in the storage.

niksv: added converter

services/src/kilda-core/kilda-model/src/main/java/org/openkilda/model/Flow.java
```diff
@@ -493,6 +499,17 @@ private FlowPathStatus getFlowPrioritizedPathStatus(FlowPath... flowPaths) {
                 .orElse(null);
     }
 
+    /**
+     * Return detect connected devices flags.
+     */
+    public void setDetectConnectedDevices(DetectConnectedDevices detectConnectedDevices) {
+        if (detectConnectedDevices == null) {
```
nikitamarchenko: maybe use Optional here

PR #2698, opened 2019-08-14
Allocate LLDP resources for flows

niksv:
We need to add the index to the database:
```
CREATE INDEX ON :`connected_device` (`type`);
```
   
Part of #2582


services/src/kilda-core/kilda-model/src/main/java/org/openkilda/model/Cookie.java
```diff
@@ -42,6 +42,9 @@
 
     public static final long FLOW_COOKIE_VALUE_MASK = 0x00000000FFFFFFFFL;
 
+    public static final long LLDP_TYPE_MASK = 0x1FF0000000000000L; // 9 bits for cookie type
```
timofei-durakov: this is not lldp mask afair

niksv: fixed

PR #2702, opened 2019-08-14
add tests for switch features

andriidovhan:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchFeaturesSpec.groovy
```diff
@@ -0,0 +1,118 @@
...
+        switchFeatures.supportedTransitEncapsulation = ["test"]
+        northbound.updateSwitchFeatures(sw.dpId, switchFeatures)
+
+        then: "Human readable error is returned"
+        def exc = thrown(HttpClientErrorException)
+        exc.statusCode == HttpStatus.BAD_REQUEST
+        exc.responseBodyAsString.to(MessageError).errorMessage == "Not supported"
+    }
+
+    def "Unable to create a transit_vlan flow in case switch feature doesn't support this type of encapsulation"() {
```
rtretyak: Do we need more iterations here? Try empty array as `supportedTransitEncapsulation`, try having both etc.

andriidovhan: the `supportMeters/supportBfd` fields are ignored. 
I mean we can change value for these fields, but they are ignored by system

rtretyak: I mean more iterations only for `supportedTransitEncapsulation`

PR #2713, opened 2019-08-19
Adjust all tests to properly handle antiflap cooldown.

rtretyak:
Increase antiflap cooldown timeout to 10s for default builds

### DIFF
--- a/services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/PortAntiflapHelper.groovy
+++ b/services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/PortAntiflapHelper.groovy
@@ -1,0 +1,42 @@
+package org.openkilda.functionaltests.helpers
+
+import org.openkilda.model.SwitchId
+import org.openkilda.testing.service.northbound.NorthboundService
+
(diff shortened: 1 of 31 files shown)


services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/PortAntiflapHelper.groovy
```diff
@@ -0,0 +1,42 @@
...
+
+    Map<Tuple2<SwitchId, Integer>, Long> history = [:]
+
+    def portUp(SwitchId swId, int portNo) {
+        def swPort = new Tuple2(swId, portNo)
+        def lastEvent = history.get(swPort)
+        if (lastEvent) {
+            sleep(lastEvent + antiflapCooldown * 1000 - System.currentTimeMillis())
+        }
+        history.remove(swPort)
```
andriidovhan: I guess it would be better to move `history.remove(swPort)` inside the  `if` condition.
(there is nothing to 'remove' in case `lastEvent = false`)

rtretyak: fixed

PR #2717, opened 2019-08-19
FL: Read from the latest offset from disco topic

### LINKED ISSUE #2084: Don't send discovery confirmation in the case of failed delivery.
Add check result of sendDiscoveryMessage in doDiscoverIslCommand.

nikitacherevko:
Floodlight should skip all outdated discovery packets on startup.

To prevent ISL fails on floodlight restart network topology should not consider disco packet as confirmed if it can not be sent from floodlight to switches.

Close #2084

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2717)

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/network/service/NetworkWatcherService.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/network/service/NetworkWatcherService.java
@@ -93,13 +93,18 @@ public class NetworkWatcherService {
     }
 
     /**
-     * .
+     * Process discovery packet confirmation.
      */
-    public void confirmation(Endpoint endpoint, long packetNo) {
+    public void processConfirmation(Endpoint endpoint, long packetNo, boolean confirmed) {
         log.debug("Watcher service receive SEND-confirmation for {} id:{} task:{}", endpoint, packetNo, taskId);
         Packet packet = Packet.of(endpoint, packetNo);
         if (producedPackets.remove(packet)) {
-            confirmedPackets.add(packet);
+            if (confirmed) {
+                confirmedPackets.add(packet);
+            } else {
+                log.warn("Discovery packet was rejected for {} id:{} task:{}. "
+                        + "Floodlight failed to send packet to the switch.", endpoint, packetNo, taskId);
+            }
         } else if (log.isDebugEnabled()) {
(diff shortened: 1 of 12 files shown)


rtretyak:
This is aimed to fix the problem when floodlight comes back online after being down for a certain amount of time (previously FL became busy getting through all the discovery requests that happen while it's off).
Test case:
0. 40 switches topology
1. Stop FL while the rest of the kilda is up
2. wait for 15 minutes
3. bring FL up.

Previously this would cause FL to hang/crash/not respond/300%+ CPU load. Now FL properly revives, marks all switches as active and does not show CPU overload.

rtretyak:
This causes all ISLs to 'blink' to UP->DOWN->UP state when FL revives. I assume this would cause reroutes.

services/wfm/src/main/java/org/openkilda/wfm/topology/network/service/NetworkWatcherService.java
```diff
@@ -93,13 +93,17 @@ public void tick() {
...
-    public void confirmation(Endpoint endpoint, long packetNo) {
+    public void processConfirmation(Endpoint endpoint, long packetNo, boolean confirmed) {
         log.debug("Watcher service receive SEND-confirmation for {} id:{} task:{}", endpoint, packetNo, taskId);
         Packet packet = Packet.of(endpoint, packetNo);
         if (producedPackets.remove(packet)) {
-            confirmedPackets.add(packet);
+            if (confirmed) {
+                confirmedPackets.add(packet);
+            } else {
+                log.warn("Discovery packet was rejected for {} id:{} task:{}", endpoint, packetNo, taskId);
```
nikitamarchenko: add more detail message.

rozdy: More detailed message is logged by floodlight.

rozdy:
@nikitamarchenko 
> Split that on two pr -> one for Kafka offset and one for network topology stuff. Also, I don't see new tests for network topology. 

I have splitted this pr into two commits and added tests.

rozdy:
> (nip) I believe it is supposed to be 'seek' not 'seed'.

Fixed

timofei-durakov:
let's close this one in favor of #3518

surabujin:
Implemented in another way in #3518

PR #2721, opened 2019-08-20
Fix CommandBuilder in the SwitchManager topology.

dpoltavets:
- fixed outPort parsing;
- fixed buildRemoveFlow method in accordance with VxLAN encapsulation;
- added unit tests.

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/switchmanager/service/impl/CommandBuilderImpl.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/switchmanager/service/impl/CommandBuilderImpl.java
@@ -41,7 +41,9 @@ import org.openkilda.wfm.topology.switchmanager.service.CommandBuilder;
 
 import com.fasterxml.uuid.Generators;
 import com.fasterxml.uuid.NoArgGenerator;
+import com.google.common.annotations.VisibleForTesting;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang.math.NumberUtils;
 
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


services/wfm/src/main/java/org/openkilda/wfm/topology/switchmanager/service/impl/CommandBuilderImpl.java
```diff
@@ -136,21 +138,44 @@ public CommandBuilderImpl(PersistenceManager persistenceManager, FlowResourcesCo
...
+            }
+        }
 
         Optional<FlowApplyActions> actions = Optional.ofNullable(entry.getInstructions())
                 .map(FlowInstructions::getApplyActions);
 
-        Integer outPort = actions.map(FlowApplyActions::getFlowOutput).map(Integer::valueOf).orElse(null);
+        Integer outPort = actions
+                .map(FlowApplyActions::getFlowOutput)
+                .filter(NumberUtils::isNumber)
```
surabujin: Why `OFPort.IN_PORT` is not a number?

dpoltavets:
```
public static final OFPort IN_PORT = new OFPort.NamedPort(-8, "in_port");
```
More research is needed to represent this field by number. We do not know how the system will behave if negative ports return.

surabujin: All OF the ports represents with 32-bit number (unsigned  - because they have unsigned types)... There is a range for special ports (in_port is one of them). So - one more time - why we have port represented as not a number? Do we really need to pass it as "out-port" into flow-mod message?

This whole ... piece of code with "unpredictable" flow-mod(delete) request is going to be removed soon(as I hope)... so... I remove my -1.

services/wfm/src/main/java/org/openkilda/wfm/topology/switchmanager/service/impl/CommandBuilderImpl.java
```diff
@@ -136,21 +138,44 @@ public CommandBuilderImpl(PersistenceManager persistenceManager, FlowResourcesCo
...
 
         Integer inPort = entryMatch.map(FlowMatchField::getInPort).map(Integer::valueOf).orElse(null);
+
+        FlowEncapsulationType encapsulationType = FlowEncapsulationType.TRANSIT_VLAN;
+        Integer encapsulationId = null;
         Integer vlan = entryMatch.map(FlowMatchField::getVlanVid).map(Integer::valueOf).orElse(null);
+        if (vlan != null) {
+            encapsulationId = vlan;
+        } else {
+            Integer tunnelId = entryMatch.map(FlowMatchField::getTunnelId).map(Integer::valueOf).orElse(null);
```
nikitacherevko: Can you leave optional and do not assign null value and then you can use optional check instead of comparing with a null

Issue #2426, opened 2019-05-30
improve the 'makePathMorePreferable' method


andriidovhan:
set cost dynamically for needed links

now we set the same cost(`NOT_PREFERABLE_COST`) for all needed links

for instance: 
```
def currentCost = northbound.getLink(islToAvoid).cost
northbound.updateLinkProps([islUtils.toLinkProps(islToAvoid, ["cost": currentCost + NOT_PREFERABLE_COST])])
```

PR #2725, opened 2019-08-20
improve procedure of making path more preferable

andriidovhan:
Close: #2426

### DIFF
--- a/services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/PathHelper.groovy
+++ b/services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/PathHelper.groovy
@@ -21,7 +21,7 @@ import org.springframework.stereotype.Component
 @Component
 @Slf4j
 class PathHelper {
-    static final String NOT_PREFERABLE_COST = "99999999"
+    static final Integer NOT_PREFERABLE_COST = 10000
 
     @Autowired
     TopologyDefinition topology
@@ -38,9 +38,8 @@ class PathHelper {
     void makePathNotPreferable(List<PathNode> path) {
         def notPreferableIsls = getInvolvedIsls(path)
         log.debug "ISLs to avoid: $notPreferableIsls"
-        northbound.updateLinkProps(notPreferableIsls.collectMany {
-            [islUtils.toLinkProps(it, ["cost": NOT_PREFERABLE_COST]),
-             islUtils.toLinkProps(it.reversed, ["cost": NOT_PREFERABLE_COST])]
+        northbound.updateLinkProps(notPreferableIsls.collectMany { isl ->
+            [islUtils.toLinkProps(isl, ["cost": (NOT_PREFERABLE_COST ** 2).toString()])]
         })
     }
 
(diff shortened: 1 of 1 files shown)


services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/PathHelper.groovy
```diff
@@ -68,8 +67,11 @@ class PathHelper {
             throw new Exception("Unable to make some path more preferable because both paths use same ISLs")
         }
         log.debug "ISL to avoid: $islToAvoid"
-        northbound.updateLinkProps([islUtils.toLinkProps(islToAvoid, ["cost": NOT_PREFERABLE_COST]),
-                                    islUtils.toLinkProps(islToAvoid.reversed, ["cost": NOT_PREFERABLE_COST])])
+        def maxAvailableCost = morePreferableIsls.findAll { it != islToAvoid }.collect {
```
rtretyak: morePreferableIsls cannot contain islToAvoid, findAll is useless

services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/PathHelper.groovy
```diff
@@ -68,8 +67,11 @@ class PathHelper {
...
+            northbound.getLink(it)
+        }.cost.max()
+        northbound.updateLinkProps([islUtils.toLinkProps(islToAvoid,
+                ["cost": (maxAvailableCost + NOT_PREFERABLE_COST).toString()])])
```
rtretyak: I literally see no sense in the `maxAvailableCost` variable and it's impact. The preferred path always depends on the sum of all isl costs on that path, having the single maximum cost gives you nothing, since two paths can have different amount of isls.
How does setting `maxAvailableCost + NOT_PREFERABLE_COST` instead of `NOT_PREFERABLE_COST` improves the algo?

services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/PathHelper.groovy
andriidovhan: yes, my fault

services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/PathHelper.groovy
andriidovhan: In a nutshell I thought it's more safe then just set `NOT_PREFERABLE_COST`.

This algo was rewritten and now I compare total cost of both paths and then make a decision about creating/updating linkProps
```
  def totalCostOfMorePrefPath = morePreferableIsls.collect {
            northbound.getLink(it)
        }.cost.sum()
        def totalCostOfLessPrefPath = getInvolvedIsls(lessPreferablePath).collect {
            northbound.getLink(it)
        }.cost.sum()
        if(totalCostOfLessPrefPath < totalCostOfMorePrefPath) {
            northbound.updateLinkProps([islUtils.toLinkProps(islToAvoid,
                    ["cost": (totalCostOfMorePrefPath).toString()])])
        }
```

services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/PathHelper.groovy
```diff
@@ -68,8 +67,16 @@ class PathHelper {
...
         log.debug "ISL to avoid: $islToAvoid"
-        northbound.updateLinkProps([islUtils.toLinkProps(islToAvoid, ["cost": NOT_PREFERABLE_COST]),
-                                    islUtils.toLinkProps(islToAvoid.reversed, ["cost": NOT_PREFERABLE_COST])])
+        def totalCostOfMorePrefPath = morePreferableIsls.collect {
+            northbound.getLink(it)
+        }.cost.sum()
+        def totalCostOfLessPrefPath = getInvolvedIsls(lessPreferablePath).collect {
+            northbound.getLink(it)
+        }.cost.sum()
+        if(totalCostOfLessPrefPath <= totalCostOfMorePrefPath) {
```
rtretyak: the `islToAvoid`-related code should be moved inside this if

andriidovhan: fixed

rtretyak: I don't see the change. it should be 
```
if(totalCostOfLessPrefPath <= totalCostOfMorePrefPath) {
         def islToAvoid = getInvolvedIsls(lessPreferablePath).find {
            !morePreferableIsls.contains(it) && !morePreferableIsls.contains(it.reversed)
        }
        if (!islToAvoid) {
            throw new Exception("Unable to make some path more preferable because both paths use same ISLs")
        }
    ....
     .....
}
```
Also, javadoc should be updated according to the new behavior. Also it sohuld be mentioned that 'null' can be returned if we didn't change any isls.

PR #2727, opened 2019-08-21
Split consumer groups for different regions

timofei-durakov:


docker-compose.yml
```diff
@@ -270,7 +270,7 @@ services:
     build:
       context: services/src/floodlight-service/floodlight-modules
     image: "kilda/floodlight:${full_build_number:-latest}"
-    command: ["floodlightStats", "-Dorg.openkilda.floodlight.KafkaChannel.floodlight-region=1"]
+    command: ["floodlightStats", "-Dorg.openkilda.floodlight.KafkaChannel.floodlight-region=1", "-Dorg.openkilda.floodlight.KafkaChannel.kafka-groupid=floodlight_1"]
```
surabujin: Is it supposed to match with `floodlight_one`?

timofei-durakov: yes

PR #2734, opened 2019-08-23
Add second floodlight region for 'make up-test-mode'

rtretyak:


rtretyak:
This requires a lot more work than I expected. Need to teach lock-keeper to work with different regions, which can be located on different hosts/containers.

PR #2745, opened 2019-08-28
Add switch connection port info. Fix #2681

### LINKED ISSUE #2681: System should save switch port information
As a user, I want to have ability to know the switch port (not only switch address), so that I can more precisely identify a switch.
AC:
1. Switch port information is saved to DB
2. Switch port information is returned from Northbound when dumping switch(es)

rozdy:
Closes #2681

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2745)


services/src/messaging/src/main/java/org/openkilda/messaging/nbtopology/response/GetSwitchResponse.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Represents a switch northbound response.
+ */
+@Value
+public class GetSwitchResponse extends InfoData {
+    private static final long serialVersionUID = 1L;
+
+    @JsonProperty("payload")
+    protected Switch payload;
```
timofei-durakov: do not iinject switch in messaging here

services/src/kilda-core/kilda-model/src/main/java/org/openkilda/model/Switch.java
```diff
@@ -73,6 +73,8 @@
 
     private String address;
 
+    private int port;
```
timofei-durakov: it's db model class, please don't add a port

rozdy:
> > ```java
> >     @AfterMapping
> >     default void setSwitchState(Switch data, @MappingTarget SwitchDto.SwitchDtoBuilder target) {
> > ```
> 
> Is it really required? Why mapstruct do not find/use exisging `default String map(SwitchStatus status)` ?

@surabujin 
Yep it's required because we need to map SwitchStatus to corresponding SwitchChangeType string representation manually. Previously this conversion was done on the strom side.

rozdy:
@surabujin 
> Where is the check about "amount" of parts after the split? DB can be updated not only by our code, but by some external tool too, or even by "hands". You can't blindly trust data stored into DB.

Reworked.

> Do we really need our own representation of inet-address? `SpeakerSwitchView` use `InetSocketAddress`. What benefits we have from own representation?

I don't think using `InetSocketAddress` in persistence layer is a good idea. We will have unnecessary address and port validation while recreating this object from `String` representation.


> We definitely need other representation somewhere in v2.

Totally agree but this changes are related to v1.

> And one more point - where is the migration for existing data?

We can't migrate already existing in db switches because we don't have port information now. As far as I can see all switches will be updated with port info when they will be discovered.

rozdy:
@surabujin 
> > > Do we really need our own representation of inet-address? SpeakerSwitchView use InetSocketAddress. What benefits we have from own representation?
> > 
> > 
> > I don't think using InetSocketAddress in persistence layer is a good idea. We will have unnecessary address and port validation while recreating this object from String representation.
> 
> Still didn't get why we need "own" representation of socket address. You already reimplement "validation" on data parsing time. And as far as I remember InetSocketAddress do not perform any "validation" on create time... (only stupid reverse DNS lookup if it created via "incorrect" constructor).
> 
> I like "correct" representation of the domain model, but I don't see any reason to reimplement common data structures/

Reworked with InetSocketAddress.

services/src/kilda-core/kilda-model/src/main/java/org/openkilda/converters/InetSocketAddressConverter.java
```diff
@@ -0,0 +1,59 @@
...
+public class InetSocketAddressConverter implements CompositeAttributeConverter<InetSocketAddress> {
+
+    private static final String ADDRESS = "address";
+    private static final String PORT = "port";
+
+    @Override
+    public Map<String, ?> toGraphProperties(InetSocketAddress value) {
+        Map<String, Object> address = new HashMap<>();
+        if (value != null) {
+            address.put(ADDRESS, value.getAddress().getHostAddress());
```
niksv: should we check that `getAddress()` is not null?

rozdy: Added null check

rozdy:
@surabujin 
> > ```java
> >             return new InetSocketAddress(InetAddress.getByName(address), port);
> >         } catch (UnknownHostException e) {
> >             throw new IllegalStateException(format("Switch address '%s' is invalid", address), e);
> > ```
> 
> Just notice - port can be invalid too - (<= 0 or > 65535), I think it will trigger the same exception. In this case, our error message becomes inaccurate.

Invalid port will cause `new IllegalArgumentException("port out of range:" + port);`

> _[services/src/kilda-core/kilda-model/src/main/java/org/openkilda/model/Switch.java, line 78 at r7](https://reviewable.io/reviews/telstra/open-kilda/2745#-LpFklqU6xvCgEnC4wR4:-LpFklqVBFvqb63Pp6w0:b-6oapei) ([raw file](https://github.com/telstra/open-kilda/blob/cfd8849cf48d90e202ae45752d91f7f1e7d2a33c/services/src/kilda-core/kilda-model/src/main/java/org/openkilda/model/Switch.java#L78)):_
> 
> > ```java
> >     @Convert(InetSocketAddressConverter.class)
> >     private InetSocketAddress socketAddress;
> > ```
> 
> Add a comment here to notify that this is socket address on FL side, not switch side.
> 

Added.

> > ```java
> >             return Switch.builder()
> >                     .switchId(new SwitchId(switchId))
> >                     .socketAddress(new InetSocketAddress(InetAddress.getByName("1.1.1.1"), 30070))
> > ```
> 
> `1.1.1.1` is a real inet address.
> As for me, better to use "grey" addresses ... just to be sure we do not interfere with anything.
> 

Changed.

> > ```java
> >             throw new SwitchReferenceLookupException(datapath, "persistent record is missing");
> >         }
> >         return makeSwitchReference(datapath, sw.get().getSocketAddress().getAddress().getHostAddress());
> > ```
> 
> The only role of makeSwitchReferences is to convert `String` representation of address into `InetAddress`. Now, when switch object can provide InetAddress object, we do not need `makeSwirchReference` at all... Definitely we do not need to make conversion `InetAddress` => `String` => `InetAddress`.

Fixed.

PR #2752, opened 2019-08-30
Add protected paths to flow cache in stats topology. Fix #2749

### LINKED ISSUE #2749: Stats topology flows cache init do not consider flow's protected paths
So at the start of stats topology it's cache has no mapping between flow's protected path cookies and flowId. As the resuls there stats related to flow's protected paths are ignores. Also, stats produce "wanr" message on each flow stats batch.

Connected to #2731

rozdy:
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2752)

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/stats/bolts/CacheBolt.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/stats/bolts/CacheBolt.java
@@ -26,6 +26,7 @@ import org.openkilda.messaging.info.stats.FlowStatsData;
 import org.openkilda.messaging.info.stats.FlowStatsEntry;
 import org.openkilda.messaging.info.stats.MeterStatsData;
 import org.openkilda.messaging.info.stats.MeterStatsEntry;
+import org.openkilda.model.Flow;
 import org.openkilda.model.FlowPath;
 import org.openkilda.model.SwitchId;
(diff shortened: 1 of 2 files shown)


rozdy:
> 
> ... `extractAllFlowPath` sounds better as for me.
> 
> It will raise `throw new NoSuchElementException("No value present")` on null value.

@surabujin Renamed and fixed.

services/wfm/src/main/java/org/openkilda/wfm/topology/stats/bolts/CacheBolt.java
```diff
@@ -116,6 +109,33 @@ private void initFlowCache(FlowRepository flowRepository) {
         }
     }
 
+    private Stream<FlowPath> extractAllFlowPaths(Flow flow) {
```
sergii-iakovenko: This could be simpler:
```
return Stream.of(flow.getForwardPath(), flow.getReversePath(), flow.getProtectedForwardPath(), 
flow.getProtectedReversePath()).filter(Objects::nonNull).peek(p -> {
                    p.setSrcSwitch(flow.getSrcSwitch());
                    p.setDestSwitch(flow.getDestSwitch());
                })
```

rozdy: Fixed

Issue #2762, opened 2019-09-04
system is marking protected path as down(when system can't find not-overlapping new protected path) while rerouting flow


andriidovhan:
**Steps to reproduce:**
1. Create a protected flow on switches with three diverse paths only;
2. detect which path is not used by the flow and make it more preferable;
3. reroute the flow; (response: ` "rerouted": true`)

**Actual result**
The flow status is DEGRADED
Main path of the flow is rerouted and marked as UP
Protected path is not rerouted and marked as DOWN
error is kibana:
`Failed reroute of the flow <flowId>, reason: Reroute <flowId> is unsuccessful: can't find non overlapping new protected path.`

**Expected result:** 
The protected path is operable, so the protected path should be marked as UP
and Flow should be in UP state because 

**NOTE**
described behaviour doesn't depend from ` "flows_reroute_via_flowhs":` configuration

rtretyak:
This is a h&s -only issue, i.e. happens only with h&s reroutes 'on'

andriidovhan:
for QA: remove the Ignore annotation when this issue is fixed
```
"System doesn't mark protected path as DOWN when can't find non overlapping protected path during intentional\
 reroute"()
```

pkazlenka:
Issue is fixed, as well as the test.

PR #2764, opened 2019-09-04
add test  system is able to reuse current protected path when can't find new protected path while intentional reroute

andriidovhan:
Covers #2762

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/ProtectedPathSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/ProtectedPathSpec.groovy
@@ -253,9 +253,10 @@ class ProtectedPathSpec extends HealthCheckSpecification {
     @Unroll
     def "System reroutes #flowDescription flow to more preferable path and ignores protected path when reroute\
  is intentional"() {
-        given: "Two active neighboring switches with two diverse paths at least"
+        // 'and ignores protected path' means that the main path won't changed to protected
+        given: "Two active neighboring switches with four diverse paths at least"
         def switchPair = topologyHelper.getAllNeighboringSwitchPairs().find {
-            it.paths.unique(false) { a, b -> a.intersect(b) == [] ? 1 : 0 }.size() >= 2
+            it.paths.unique(false) { a, b -> a.intersect(b) == [] ? 1 : 0 }.size() >= 4
         } ?: assumeTrue("No suiting switches found", false)
 
         and: "A flow with protected path"
@@ -288,6 +289,7 @@ class ProtectedPathSpec extends HealthCheckSpecification {
         def newCurrentPath = pathHelper.convert(flowPathInfoAfterRerouting)
         newCurrentPath != currentPath
         newCurrentPath != currentProtectedPath
+        //protected path is rerouted too, because more preferable path is exist
         def newCurrentProtectedPath = pathHelper.convert(flowPathInfoAfterRerouting.protectedPath)
         newCurrentProtectedPath != currentPath
         newCurrentProtectedPath != currentProtectedPath
@@ -306,7 +308,7 @@ class ProtectedPathSpec extends HealthCheckSpecification {
     @Unroll
     def "System is able to switch #flowDescription flow to protected path and ignores more preferable path when reroute\
  is automatical"() {
-        given: "Two active not neighboring switches with two diverse paths at least"
+        given: "Two active not neighboring switches with three diverse paths at least"
         def switchPair = topologyHelper.getAllNotNeighboringSwitchPairs().find {
             it.paths.unique(false) { a, b -> a.intersect(b) == [] ? 1 : 0 }.size() >= 3
         } ?: assumeTrue("No suiting switches found", false)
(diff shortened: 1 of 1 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/ProtectedPathSpec.groovy
```diff
@@ -253,9 +253,9 @@ class ProtectedPathSpec extends HealthCheckSpecification {
     @Unroll
     def "System reroutes #flowDescription flow to more preferable path and ignores protected path when reroute\
  is intentional"() {
-        given: "Two active neighboring switches with two diverse paths at least"
+        given: "Two active neighboring switches with four diverse paths at least"
         def switchPair = topologyHelper.getAllNeighboringSwitchPairs().find {
-            it.paths.unique(false) { a, b -> a.intersect(b) == [] ? 1 : 0 }.size() >= 2
+            it.paths.unique(false) { a, b -> a.intersect(b) == [] ? 1 : 0 }.size() >= 4 //issue-2762
```
rtretyak: I don't think this is possible neither at our virtual nor at staging topology, thus it is doomed to always be skipped now with "No suiting switches found"

rtretyak: I don't really understand what means `//issue-2762`. Is picking a path reproduces this issue? no sense

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/ProtectedPathSpec.groovy
```diff
@@ -1105,6 +1105,60 @@ class ProtectedPathSpec extends HealthCheckSpecification {
         flowHelper.deleteFlow(flow.id)
     }
 
+    @Ignore("https://github.com/telstra/open-kilda/issues/2762")
+    def "System doesn't mark protected path as DOWN when can't find non overlapping protected path during intentional\
```
rtretyak: Sorry? The system _SHOULD_ mark protected path as DOWN when it can't find non overlapping protected path

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/ProtectedPathSpec.groovy
rtretyak: Nobody sees a 'diff' when looking at this code. After this pr it would have became
`it.paths.unique(false) { a, b -> a.intersect(b) == [] ? 1 : 0 }.size() >= 4 //issue-2762`

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/ProtectedPathSpec.groovy
```diff
@@ -253,9 +253,10 @@ class ProtectedPathSpec extends HealthCheckSpecification {
     @Unroll
     def "System reroutes #flowDescription flow to more preferable path and ignores protected path when reroute\
  is intentional"() {
-        given: "Two active neighboring switches with two diverse paths at least"
+        // 'and ignores protected path' means that the main path won't changed to protected
```
rtretyak: I don't understand the purpose, sorry. If both main and protected paths are changed, then how can you understand whether they swapped or no? Both paths change and you can't understand whether swap happened or not. It doesn't even matter. What if they swapped? I know that this test is old, but still. This should be renamed or removed

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/ProtectedPathSpec.groovy
```diff
@@ -1105,6 +1107,60 @@ class ProtectedPathSpec extends HealthCheckSpecification {
...
+        and: "Main path is not changed to protected path"
+        def flowPathInfoAfterRerouting = northbound.getFlowPath(flow.id)
+        def newCurrentPath = pathHelper.convert(flowPathInfoAfterRerouting)
+        newCurrentPath != currentPath
+        newCurrentPath != currentProtectedPath
+
+        and: "Protected path is not changed"
+        pathHelper.convert(flowPathInfoAfterRerouting.protectedPath) == currentProtectedPath
+
+        and: "Protected path is not marked as DOWN"
```
rtretyak: `and: "Protected path is UP"`
or even `Flow and both its paths are UP`

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/ProtectedPathSpec.groovy
```diff
@@ -1105,6 +1107,60 @@ class ProtectedPathSpec extends HealthCheckSpecification {
...
+        alternativePaths.each { pathHelper.makePathMorePreferable(it, currentPath) }
+        alternativePaths.each { pathHelper.makePathMorePreferable(it, currentProtectedPath) }
+
+        and: "Init intentional reroute"
+        def rerouteResponse = northbound.rerouteFlow(flow.id)
+
+        then: "Flow is rerouted"
+        rerouteResponse.rerouted
+
+        and: "Main path is not changed to protected path"
```
rtretyak:
```
and: "Init intentional reroute"
then: "Flow is rerouted"
and: "Main path is not changed to protected path"
and: "Protected path is not changed"
```
It's not quite clear from the above steps what have happened to the main path after all. Can we please name steps closer to _what has/should happen_ instead of _what has NOT happen_. For sure it's no a strong rule and we are flexible, but in this current case I don't really understand in which state my main path is expected to be

andriidovhan: fixed
```
and: "Init intentional reroute"
then: "Flow should be rerouted"
and: "Flow main path should be rerouted to a new path and ignore protected path"
and: "Flow protected path shouldn't be rerouted due to lack of non overlapping path"
```

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/ProtectedPathSpec.groovy
andriidovhan: I believe we need this test
aim of this test is: "verify that system ignores protected path while rerouting and reroutes main/protected paths to more preferable"

rtretyak: ok

PR #2767, opened 2019-09-05
extend statistic test coverage for  different type of flow

andriidovhan:
- add tests for verifying that statistic is collected for different type of flow:
-- protected(up/degraded) after swapping/rerouting;
-- pinned and unmetered;
- add possibility to set duration for non bidirectional traffic;
- refactor old tests(limit in time buildBidirectionalExam);


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/stats/MflStatSpec.groovy
```diff
@@ -221,4 +225,361 @@ class MflStatSpec extends HealthCheckSpecification {
...
+            assert protectedPathStat.size() > mainPathStat.size()
+            def newProtectedReverseCookieStat = otsdb.query(startTime, metric, tags + [cookie: protectedReverseCookie]).dps
+            assert !newProtectedReverseCookieStat.values().findAll { it != 0 }.empty
+        }
+
+        and: "Stats is not changed for previous main path cookies"
+        def newMainForwardCookieStat = otsdb.query(startTime, metric, tags + [cookie: mainForwardCookie]).dps
+        def newMainReverseCookieStat = otsdb.query(startTime, metric, tags + [cookie: mainReverseCookie]).dps
+        newMainForwardCookieStat.size() == mainForwardCookieStat.size()
+        //sometimes newMainReverseCookieStat.size() = mainReverseCookieStat.size() + 1
```
rtretyak: Check point values instead of points amount? New point value should equal to old point value.

andriidovhan: decided to change verification
now we are going to check the following:
```
  and: "Stats is still being collected for previous egress cookie of main path"
 and: "System doesn't collect stats anymore for previous ingress cookie of main path"
```

andriidovhan: decided to change verification
now we are going to check the following:
```
  and: "Stats is still being collected for previous egress cookie of main path"
 and: "System doesn't collect stats anymore for previous ingress cookie of main path"
```

PR #2775, opened 2019-09-09
add monitoring section in readme file for performance test

andriidovhan:


services/src/performance-tests/README.md
```diff
@@ -0,0 +1,8 @@
+**Monitoring**
+
+[Sampler](https://github.com/sqshq/sampler) is a tool for shell commands execution, visualization and alerting.
+Configured with a simple YAML file.
+
+If there is a way to get a metric using shell command - then it can be visualized with Sampler momentarily. 
+
```
rtretyak: Any quick commands to make this running?

andriidovhan: `sampler -c config.yml`

timofei-durakov:
Do we want sampler in kilda? please give some feedback on using this

andriidovhan:
> Do we want sampler in kilda? please give some feedback on using this
we need some time to provide you with feedback about that. but now it looks so easy in configuring and getting any metrics what we need.
the `sampler` are going to be used on a local environment only.
we are not going to integrate it in kilda.

PR #2780, opened 2019-09-10
add tets for port history antiflap

andriidovhan:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
```diff
@@ -55,21 +59,22 @@ class PortHistorySpec extends HealthCheckSpecification {
...
+        Wrappers.wait(WAIT_OFFSET) {
+            timestampAfterUp = System.currentTimeMillis()
+            with(northboundV2.getPortHistory(isl.srcSwitch.dpId, isl.srcPort, timestampBefore, timestampAfterUp)
+                    .sort { a, b -> b.date <=> a.date }) {
+                checkPortHistory(it[0], isl.srcSwitch.dpId, isl.srcPort, "PORT_UP")
+                checkPortHistory(it[1], isl.srcSwitch.dpId, isl.srcPort, "ANTI_FLAP_DEACTIVATED")
+            }
         }
 
         and: "Port history on the dst switch is not empty when link is direct"
```
rtretyak: I wondering whether we should run the exact same verifications for `dst` when link is direct, instead of just checking size. That shouldn't be difficult to implement

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
```diff
@@ -39,13 +42,14 @@ class PortHistorySpec extends HealthCheckSpecification {
...
-            it.size() == 2
-            def antiFlapActivatedEvent = it.find { it.event == "ANTI_FLAP_ACTIVATED" }
-            checkPortHistory(antiFlapActivatedEvent, isl.srcSwitch.dpId, isl.srcPort, "ANTI_FLAP_ACTIVATED")
-            def portDownEvent = it.find { it.event == "PORT_DOWN" }
-            checkPortHistory(portDownEvent, isl.srcSwitch.dpId, isl.srcPort, "PORT_DOWN")
+        Wrappers.wait(WAIT_OFFSET) {
+            Long timestampAfterDown = System.currentTimeMillis()
+            with(northboundV2.getPortHistory(isl.srcSwitch.dpId, isl.srcPort, timestampBefore, timestampAfterDown)) {
+                it.size() == 2
+                checkPortHistory(it.find { it.event == "ANTI_FLAP_ACTIVATED" }, isl.srcSwitch.dpId, isl.srcPort,
```
rtretyak: can we use `it[0]` instead of `.find { it.event == "ANTI_FLAP_ACTIVATED" }` here? I suppose the order of these two events is always the same

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
```diff
@@ -150,13 +155,69 @@ class PortHistorySpec extends HealthCheckSpecification {
...
+            Long timestampAfterDown = System.currentTimeMillis()
+            with(northboundV2.getPortHistory(isl.srcSwitch.dpId, isl.srcPort, timestampBefore, timestampAfterDown)) {
+                assert it.size() == 2
+            }
+        }
+
+        when: "Generate antiflap statistic"
+        def count = 0
+        //confd/templates/wfm/topology.properties.tmpl =>  port.antiflap.stats.dumping.interval.seconds = 60
+        def antiflapDumpingInterval = 60
```
rtretyak: I don't understand what this interval means. Can you please explain in a comment?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
```diff
@@ -150,13 +155,69 @@ class PortHistorySpec extends HealthCheckSpecification {
...
+        Wrappers.timedLoop(antiflapDumpingInterval * 0.9) {
+            northbound.portUp(isl.srcSwitch.dpId, isl.srcPort)
+            sleep(2000)
+            northbound.portDown(isl.srcSwitch.dpId, isl.srcPort)
+            sleep(2000)
+            count += 1
+        }
+
+        then: "Antiflap statistic is available in port history"
+        Wrappers.wait(antiflapDumpingInterval * 0.1 + WAIT_OFFSET) {
```
rtretyak: Will I see the `ANTI_FLAP_PERIODIC_STATS` entry if antiflap was deactivated in less than 60 seconds? This would be more interesting test case than flapping for 50s

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
```diff
@@ -150,13 +155,69 @@ class PortHistorySpec extends HealthCheckSpecification {
...
+            Long timestampAfterStat = System.currentTimeMillis()
+            with(northboundV2.getPortHistory(isl.srcSwitch.dpId, isl.srcPort, timestampBefore, timestampAfterStat)) {
+                it.size() == 3
+                def antiflapStat = it[-1]
+                checkPortHistory(antiflapStat, isl.srcSwitch.dpId, isl.srcPort, "ANTI_FLAP_PERIODIC_STATS")
+                assert count == antiflapStat.downCount
+                assert count == antiflapStat.upCount
```
rtretyak: Why verify down/up counts both in `checkPortHistory` and in test directly?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
```diff
@@ -55,21 +59,22 @@ class PortHistorySpec extends HealthCheckSpecification {
...
-            checkPortHistory(antiFlapDeactivatedEvent, isl.srcSwitch.dpId, isl.srcPort, "ANTI_FLAP_DEACTIVATED")
-            def portUpEvent = it.find { it.event == "PORT_UP" }
-            checkPortHistory(portUpEvent, isl.srcSwitch.dpId, isl.srcPort, "PORT_UP")
+        Long timestampAfterUp
+        Wrappers.wait(WAIT_OFFSET) {
+            timestampAfterUp = System.currentTimeMillis()
+            with(northboundV2.getPortHistory(isl.srcSwitch.dpId, isl.srcPort, timestampBefore, timestampAfterUp)
+                    .sort { a, b -> b.date <=> a.date }) {
+                checkPortHistory(it[0], isl.srcSwitch.dpId, isl.srcPort, "PORT_UP")
+                checkPortHistory(it[1], isl.srcSwitch.dpId, isl.srcPort, "ANTI_FLAP_DEACTIVATED")
```
rtretyak: Antiflap can also be deactivated based on just port down (after cooldown). This is not tested here. I'm also wondering whether it makes sense just to add history verifications to AntiflapSpec?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
rtretyak: 1. This cannot be true since your test generates antiflap events for `60*0.9` which is less than 60 seconds.
2. Even if its true, then this is wrong behavior. We should regardless log antiflap periodic stats even if it lasted 10 seconds. I don't really understand why we shouldn't

andriidovhan: 1. it's true. ANTI_FLAP_PERIODIC_STATS collects stat every 60 seconds
2. I suppose we shouldn't collect  `ANTI_FLAP_PERIODIC_STATS` after antiflap is deactivated
also this situation is possible only on virtual env, because `antiflap.cooldown=10` and  `antiflapDumpingInterval = 60`
it isn't  possible on staging/production envs where `antiflap.cooldown=60 / antiflap.cooldown=300`

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
andriidovhan: I will add `VIRTUAL` tag and relevant comment to needed tests

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
andriidovhan: I will add VIRTUAL tag and relevant comment to needed tests

andriidovhan: tag and comment were added

andriidovhan: tests were refactored and new test was added(` def "System shows antiflap statistic in the ANTI_FLAP_DEACTIVATED event when antiflap is deactivated\
 before collecting ANTI_FLAP_PERIODIC_STATS"()`)

now tests should work fine on stage and virtual envs

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
```diff
@@ -150,13 +160,130 @@ class PortHistorySpec extends HealthCheckSpecification {
...
+        antiflap.portUp(isl.srcSwitch.dpId, isl.srcPort)
+        Wrappers.wait(WAIT_OFFSET + discoveryInterval + antiflapCooldown) {
+            assert islUtils.getIslInfo(isl).get().state == IslChangeType.DISCOVERED
+        }
+    }
+
+    def "System shows antiflap statistic in the ANTI_FLAP_DEACTIVATED event when antiflap is deactivated\
+ before collecting ANTI_FLAP_PERIODIC_STATS"() {
+        given: "A direct link"
+        assumeTrue("antiflap.cooldown should be three times less than port.antiflap.stats.dumping.interval.seconds",
```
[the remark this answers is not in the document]
rtretyak: Also I don't really understand what is behind this logic

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
```diff
@@ -150,13 +160,130 @@ class PortHistorySpec extends HealthCheckSpecification {
...
+
+    void checkPortHistory(PortHistoryResponse portHistory, SwitchId switchId, Integer port, PortHistoryEvent event) {
         verifyAll(portHistory) {
             id != null
             switchId == switchId
             portNumber == port
-            it.event == event.toUpperCase() // PORT_UP, PORT_DOWN, ANTI_FLAP_ACTIVATED, ANTI_FLAP_DEACTIVATED
+            it.event == event.toString()
             date != null
+            if (event == "ANTI_FLAP_PERIODIC_STATS") {
```
rtretyak: Will it be difficult/bulky to make strict verifications in every test instead?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
```diff
@@ -150,13 +160,130 @@ class PortHistorySpec extends HealthCheckSpecification {
...
+                assert upCount >= 1
+                assert downCount >= 1
+            }
         }
     }
+
+    enum PortHistoryEvent {
```
rtretyak: Did you check whether there is dev class available with these constants?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
```diff
@@ -150,13 +160,130 @@ class PortHistorySpec extends HealthCheckSpecification {
...
+        }
+
+        when: "Generate antiflap statistic"
+        def count = 0
+        //confd/templates/wfm/topology.properties.tmpl =>  port.antiflap.stats.dumping.interval.seconds = 60
+        // this variable means how often the 'ANTI_FLAP_PERIODIC_STATS' will be logged in port history
+        def antiflapDumpingInterval = 60
+        Wrappers.timedLoop(antiflapDumpingInterval * 0.9) {
+            northbound.portUp(isl.srcSwitch.dpId, isl.srcPort)
+            sleep(2000)
```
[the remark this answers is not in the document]
andriidovhan: done
```
@Shared
    Integer intervalBetweenPortStateManipulation = (antiflapCooldown / 10).toInteger()
```

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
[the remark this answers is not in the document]
andriidovhan: The aim of this test is: check that system collects stats in `ANTI_FLAP_DEACTIVATED` event when `ANTI_FLAP_PERIODIC_STATS` was never executed.
it can be done when `antiflapCooldown` is three times less than `port.antiflap.stats.dumping.interval.seconds`

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
```diff
@@ -150,13 +160,130 @@ class PortHistorySpec extends HealthCheckSpecification {
...
+            assert islUtils.getIslInfo(isl).get().state == IslChangeType.FAILED
+            Long timestampAfterDown = System.currentTimeMillis()
+            with(northboundV2.getPortHistory(isl.srcSwitch.dpId, isl.srcPort, timestampBefore, timestampAfterDown)) {
+                assert it.size() == 2
+            }
+        }
+
+        and: "Generate antiflap statistic"
+        def count = 0
+        Wrappers.timedLoop(antiflapCooldown) {
```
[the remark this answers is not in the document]
andriidovhan: `timeLoop` was removed

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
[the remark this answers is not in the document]
rtretyak: You may have problems instantiating a shared variable based on loaded-by-spring value. Afaik the assignment of shared variables happen before context is load

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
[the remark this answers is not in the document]
rtretyak: aaaand how did you calculate the '3 times less' statement? My point of view: you just invented this rule on fly and it is not real. The real limitation is: `cooldown + durationYouFlap < ANTI_FLAP_DEACTIVATED`. I don't see where the '3 times' comes from

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
rtretyak: I don't have strong opinion here. Can be discussed, but not pushing

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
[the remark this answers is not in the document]
andriidovhan: it is not loaded, it is not failed because antiflapCooldown in this line is 0 ( not null )

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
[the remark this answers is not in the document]
andriidovhan: fixed totally

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
```diff
@@ -150,13 +166,118 @@ class PortHistorySpec extends HealthCheckSpecification {
...
+        assumeTrue("Unable to find ISL for this test", isl as boolean)
+        def timestampBefore = System.currentTimeMillis()
+
+        when: "Execute port DOWN on the src switch for activating antiflap"
+        northbound.portDown(isl.srcSwitch.dpId, isl.srcPort)
+        Wrappers.wait(WAIT_OFFSET) {
+            assert islUtils.getIslInfo(isl).get().state == IslChangeType.FAILED
+            Long timestampAfterDown = System.currentTimeMillis()
+            with(northboundV2.getPortHistory(isl.srcSwitch.dpId, isl.srcPort, timestampBefore, timestampAfterDown)) {
+                it.size() == 2
```
rtretyak: What are these two entries? Port down + antiflap activated?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
```diff
@@ -150,13 +166,118 @@ class PortHistorySpec extends HealthCheckSpecification {
...
+            }
+        }
+
+        and: "Generate antiflap statistic"
```
rtretyak: This assumes a pretty big cooldown. You did port down, then waited for isl to get down and history to appear, and only then the port up. Can we add a cooldown assumption of at least couple of seconds at the start of the test?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
```diff
@@ -150,13 +166,118 @@ class PortHistorySpec extends HealthCheckSpecification {
...
+        Wrappers.wait(antiflapCooldown + WAIT_OFFSET) {
+            Long timestampAfterStat = System.currentTimeMillis()
+            with(northboundV2.getPortHistory(isl.srcSwitch.dpId, isl.srcPort, timestampBefore, timestampAfterStat)) {
+                it.findAll { it.event == PortHistoryEvent.ANTI_FLAP_PERIODIC_STATS.toString() }.empty
+                it.size() == 3
+                def antiflapStat = it[-1]
+                checkPortHistory(antiflapStat, isl.srcSwitch.dpId, isl.srcPort,
+                        PortHistoryEvent.ANTI_FLAP_DEACTIVATED)
+                antiflapStat.downCount == 1
+                antiflapStat.upCount == 1
```
rtretyak: Should we check that these values equal to `0` for no-antiflap tests?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
andriidovhan: added
```
 assumeTrue("At least 10 seconds should be available for changing port status", antiflapCooldown >= 10)
        // assume that portDown/portUp can take some time on hardware env(no more than 10 seconds)
```

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
```diff
@@ -150,13 +167,148 @@ class PortHistorySpec extends HealthCheckSpecification {
...
+
+        and: "Cleanup: revert system to original state"
+        northbound.portUp(isl.srcSwitch.dpId, isl.srcPort)
+        Wrappers.wait(WAIT_OFFSET + discoveryInterval + antiflapCooldown) {
+            assert islUtils.getIslInfo(isl).get().state == IslChangeType.DISCOVERED
+        }
+    }
+
+    @Tags(LOW_PRIORITY)
+    def "System doesn't return antiflap statistic in port history if no flapping occurs during cooldown"() {
```
rtretyak: And why do you need a separate test for this if there is already a test without antiflap?

andriidovhan: you know better)
needed check was added to the first test

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
andriidovhan: `"System doesn't return antiflap statistic in port history if no flapping occurs during cooldown"()` - removed.

needed verification's were added to the first test

Issue #2781, opened 2019-09-11
Auto re-route could leave flow in up state in case of failure


timofei-durakov:
auto re-route triggered by isl down could leave flow in up state in case of failure.

PR #2785, opened 2019-09-11
Reroute topology updates flow status for flows

timofei-durakov:
reroute topology is responsible for updating status of all flows
before spinnning up reroute

Closes: #2781

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2785)


rtretyak:
We can also observe flows in wrong status due to #2987. Not tested, but I think that at current revision this pr does not resolve this issue

PR #2791, opened 2019-09-12
add new flowHelper method: getFlowHistoryByAction

andriidovhan:


services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/FlowHelper.groovy
```diff
@@ -261,6 +262,14 @@ class FlowHelper {
         }
     }
 
+    List<FlowHistoryPayload> getFlowHistoryByAction(String flowId, String action) {
+        getFlowHistoryByAction(flowId, null, null, action)
+    }
+
+    List<FlowHistoryPayload> getFlowHistoryByAction(String flowId, Long timeFrom, Long timeTo, String action) {
+        northbound.getFlowHistory(flowId, timeFrom, timeTo).findAll { it.action == action }.last().histories
+    }
```
rtretyak: I find this helper method to be too narrow. I 'd rather go for an explicit ` northbound.getFlowHistory(flowId, timeFrom, timeTo)` + search directly in test instead. Also, it should take `FlowAction` as fourth argument instead of `String`

services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/FlowHistoryAction.groovy
```diff
@@ -0,0 +1,20 @@
+package org.openkilda.functionaltests.helpers
+
+enum FlowHistoryAction {
+    FLOW_CREATING("Flow creating"),
+    FLOW_UPDATING("Flow updating"),
+    FLOW_REROUTING("Flow rerouting"),
+    SET_THE_FLOW_STATUS_TO_DOWN("Set the flow status to down."),
+    FLOW_WAS_VALIDATED_SUCCESSFULLY("Flow was validated successfully")
```
rtretyak: This is fine, but you are not supposed to always duplicate the phrase word-to-word. `FLOW_VALIDATION_SUCCESS` would work as well

services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/FlowHelper.groovy
rtretyak: I like this better. Even more narrow, but fully hides pretty common operation for verifying final flow state. I would have made FlowState an argument to use the same method for up/down/inprogress. Also, I think it makes sense to make 'timeTo' to fall to `System.currentTimeMillis()` by default, since I don't see a popular situation where it will be something else. timeFrom is also a candidate for a default value, because most of the time I will not be interested in specifying it.

services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/FlowHistoryAction.groovy
```diff
@@ -0,0 +1,24 @@
...
+enum FlowHistoryAction {
+    //action
+    FLOW_CREATING("Flow creating"),
+    FLOW_UPDATING("Flow updating"),
+    FLOW_REROUTING("Flow rerouting"),
+
+    //nested action
+    SET_THE_FLOW_STATUS_TO_UP("Set the flow status to UP."),
+    SET_THE_FLOW_STATUS_TO_DOWN("Set the flow status to down."),
+    FLOW_WAS_VALIDATED_SUCCESSFULLY("Flow was validated successfully"),
```
rtretyak: Ok for now. JFYI it is also possible to have `Recalculate flow status to DOWN.`, `Recalculate flow status to UP.`, `Recalculate flow status to In Progress.` during Flow Rerouting.

services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/FlowHistoryEvent.groovy
```diff
@@ -0,0 +1,16 @@
+package org.openkilda.functionaltests.helpers
+
+enum FlowHistoryEvent {
+
+    REROUTE_SUCCESS(FlowHistoryAction.FLOW_REROUTING, FlowHistoryAction.SET_THE_FLOW_STATUS_TO_UP),
+    REROUTE_FAIL(FlowHistoryAction.FLOW_REROUTING, FlowHistoryAction.SET_THE_FLOW_STATUS_TO_DOWN),
+    REROUTE_FAIL_PROTECTED_PATH(FlowHistoryAction.FLOW_REROUTING, FlowHistoryAction.FLOW_WAS_VALIDATED_SUCCESSFULLY),
```
rtretyak: I am a bit confused how FLOW_REROUTING + FLOW_WAS_VALIDATED_SUCCESSFULLY is related to protected paths

andriidovhan: I just copied it from flowHistory output after flowState was changed to DEGRADED

services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/FlowHelper.groovy
```diff
@@ -261,6 +262,28 @@ class FlowHelper {
...
 
+    /**
+     * Check that flow state is changed correctly after corresponding flow history action.
+     *
+     * @param flowId
+     * @param timeFrom history will requested from this timestamp
+     * @param event sequence of flow history actions
+     * @param state expected flow state
+     */
+    void verifyFlowState(String flowId, Long timeFrom, FlowHistoryEvent event, state) {
```
rtretyak: Please add a strict type for state

services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/FlowHelper.groovy
```diff
@@ -261,6 +262,28 @@ class FlowHelper {
         }
     }
 
+    /**
+     * Check that flow state is changed correctly after corresponding flow history action.
+     *
+     * @param flowId
+     * @param timeFrom history will requested from this timestamp
```
rtretyak: will be

services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/FlowHelper.groovy
```diff
@@ -261,6 +262,28 @@ class FlowHelper {
...
+     * @param event sequence of flow history actions
+     * @param state expected flow state
+     */
+    void verifyFlowState(String flowId, Long timeFrom, FlowHistoryEvent event, state) {
+        def timeTo = timeFrom ? System.currentTimeSeconds() : null
+        def flowHistory = northbound.getFlowHistory(flowId, timeFrom, timeTo).findAll {
+            it.action == event.initAction.toString()
+        }
+        assert flowHistory.size() == 1
```
rtretyak: are you sure that you want to limit your method with this condition? What if there were more actions on the flow? I'm not insisting, but for me it kind of limits where it can be used or forces to use the 'from' timestamp. that's fine I guess, but requires a lot of 'timestamp' vars for long tests

services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/FlowHelper.groovy
```diff
@@ -261,6 +262,28 @@ class FlowHelper {
...
+    /**
+     * Check that flow state is changed correctly after corresponding flow history action.
+     *
+     * @param flowId
+     * @param timeFrom history will requested from this timestamp
+     * @param event sequence of flow history actions
+     * @param state expected flow state
+     */
+    void verifyFlowState(String flowId, Long timeFrom, FlowHistoryEvent event, state) {
+        def timeTo = timeFrom ? System.currentTimeSeconds() : null
```
rtretyak: well, this is kind of tricky. timeTo based on timeFrom, why would you do this? Very strange and not intuitive approach

services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/FlowHistoryEvent.groovy
rtretyak: Why do you call it like that? It clearly means that `flow is validated after reroute` and I would agree with that, but the 'reroute fail for protected path' is just a special case of your test. The flow is validated, and you expect a degraded status in your exact special test, that's fine. But you cannot globally assume that 'rerouting + successfull validation' means 'reroute for protected path failed', that breaks laws of logic. Crocodiles are green, crocodiles are dangerous. Grass is green -> grass is dangerous?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/ProtectedPathSpec.groovy
```diff
@@ -893,17 +893,22 @@ class ProtectedPathSpec extends HealthCheckSpecification {
         antiflap.portDown(protectedIsls[0].dstSwitch.dpId, protectedIsls[0].dstPort)
 
         then: "Flow state is changed to DEGRADED"
-        Wrappers.wait(WAIT_OFFSET) { assert northbound.getFlowStatus(flow.id).status == FlowState.DEGRADED }
+        Wrappers.wait(WAIT_OFFSET) {
+            flowHelper.verifyFlowState(flow.id, FlowHistoryEvent.REROUTE_FAIL_PROTECTED_PATH, FlowState.DEGRADED)
```
rtretyak: This is literally false expression. You are not verifying that the reroute failed for protected path, you are just checking that a successful validation happened.

andriidovhan:
I am going to rework it at all
So I will create a new pr

PR #2796, opened 2019-09-13
add flowHistory test for v2

andriidovhan:
- add flowHistory test for v2
- fix tests with the `smoke` tag


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowHistorySpec.groovy
```diff
@@ -114,24 +153,30 @@ class FlowHistorySpec extends HealthCheckSpecification {
...
+        if (action == "createV1"){
+            actionMessage = createAction
+            historyMessage = createHistoryActionV1
+        } else if( action == "createV2") {
+            actionMessage = createAction
+            historyMessage = createHistoryActionV2
+        } else if ( action == "update") {
+            actionMessage = updateAction
+            historyMessage = updateHistoryAction
+        }
```
rtretyak: Bad design. 
1. avoid 'magic values'
2. do not try to combine 3 different methods in one method with 3 ifs. create 3 separate methods instead.

PR #2806, opened 2019-09-17
Sync LLDP rules

### LINKED ISSUE #2779: LLDP Connected Devices: add ability to synchronize lldp rules
As a system administrator, I want to be able to synchronize(install missing, remove excess) lldp rules when calling 'sync switch' endpoint, so that discrepancies in lldp rules could be fixed via NB API.
AC:
- system fixes any discrepancies in lldp rules and meters when calling `/api/v1/switches/:sw/synchronize`
- subsequent switch validate call should show no discrepancies in rules or meters

niksv:
Part of #2582
Closes #2779


niksv:
feature was refactored here https://github.com/telstra/open-kilda/pull/2911

PR #2814, opened 2019-09-19
Add disable port discovery feature

### LINKED ISSUE #2794: Disable discovery on a specific port on a switch
As an admin, I can disable discovery on  a specific port on a switch

rozdy:
Closes #2794

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/services/SwitchOperationsService.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/services/SwitchOperationsService.java
@@ -15,6 +15,8 @@
 
 package org.openkilda.wfm.topology.nbworker.services;
 
+import org.openkilda.messaging.error.ErrorType;
+import org.openkilda.messaging.error.MessageException;
 import org.openkilda.messaging.model.SwitchPropertiesDto;
 import org.openkilda.messaging.nbtopology.response.GetSwitchResponse;
 import org.openkilda.model.Flow;
@@ -26,6 +28,7 @@ import org.openkilda.model.Switch;
 import org.openkilda.model.SwitchId;
 import org.openkilda.model.SwitchProperties;
 import org.openkilda.model.SwitchStatus;
+import org.openkilda.persistence.PersistenceException;
 import org.openkilda.persistence.TransactionManager;
 import org.openkilda.persistence.repositories.FlowPathRepository;
 import org.openkilda.persistence.repositories.FlowRepository;
@@ -154,6 +157,8 @@ public class SwitchOperationsService implements ILinkOperationsServiceCarrier {
         transactionManager.doInTransaction(() -> {
             switchPropertiesRepository.findBySwitchId(sw.getSwitchId())
                     .ifPresent(sp -> switchPropertiesRepository.delete(sp));
+            portPropertiesRepository.getAllBySwitchId(sw.getSwitchId())
+                    .forEach(portPropertiesRepository::delete);
             if (force) {
                 // forceDelete() removes switch along with all relationships.
                 switchRepository.forceDelete(sw.getSwitchId());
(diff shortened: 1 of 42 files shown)


rozdy:
@surabujin 
> > ```java
> > @NoArgsConstructor
> > @JsonNaming(SnakeCaseStrategy.class)
> > public class PortPropertiesPayload extends InfoData {
> > ```
> 
> It will be better to put here `PortProperties` object from `kilda-model` package.
> 

PortProperties has reference to Switch object. Is it ok to serialize/deserialize whole Switch object when we need only switch id?

> > ```java
> >     @Mapping(source = "switchObj.switchId", target = "switchId")
> >     @Mapping(source = "port", target = "portNumber")
> >     public abstract PortPropertiesPayload map(PortProperties portInfoData);
> > ```
> 
> `portInfoData`?

Fixed.

> > ```java
> >         PortProperties portProperties = switchOperationsService.getPortProperties(request.getSwitchId(),
> >                 request.getPort());
> >         if (portProperties == null) {
> > ```
> 
> Why "response" logic live outside "service layer"?
> > ```java
> >             String message = String.format("Port properties not found for switch '%s' and port '%d'",
> >                     request.getSwitchId(), request.getPort());
> >             log.error(message);
> > ```
> 
> ??? O_o
> 
> This is the responsibility of the exception handler.

It was made for similarity reason: a lot of methods around has this logic on this layer. Should we move it somewhere?

> > ```java
> >             Optional<PortProperties> foundPortProperties =
> >                     portPropertiesRepository.getBySwitchIdAndPort(switchId, port);
> >             if (foundPortProperties.isPresent()) {
> > ```
> 
> Why for so ... complex. Revert this condition and create new instance inside... later, outside condition apply required update and save new/existing object into DB.

Update logic was moved to network topology and changed.

services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/services/SwitchOperationsService.java
```diff
@@ -147,6 +151,9 @@ public boolean deleteSwitch(SwitchId switchId, boolean force) throws SwitchNotFo
         Switch sw = switchRepository.findById(switchId)
                 .orElseThrow(() -> new SwitchNotFoundException(switchId));
 
+        if (sw.getPortProperties() != null) {
```
timofei-durakov: why is it not a single transaction for switch port properties and switch properties?

services/wfm/src/main/java/org/openkilda/wfm/topology/network/controller/port/PortFsm.java
```diff
@@ -131,6 +150,10 @@ public void finish(PortFsmState from, PortFsmState to, PortFsmEvent event, PortF
                     .callMethod("proxyDiscovery");
             builder.internalTransition().within(PortFsmState.UP).on(PortFsmEvent.FAIL)
                     .callMethod("proxyFail");
+            builder.internalTransition().within(PortFsmState.UP).on(PortFsmEvent.ENABLE_DISCOVERY)
```
sergii-iakovenko: Does it mean that we can't enable/disable discovery for a port in down state?

rozdy:
> _[services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/SwitchOperationsBolt.java, line 213 at r2](https://reviewable.io/reviews/telstra/open-kilda/2814#-LqBg5A71j_JxwcUw8Vp:-LqBg5A71j_JxwcUw8Vq:bxui6hl) ([raw file](https://github.com/telstra/open-kilda/blob/b36ec016c76a3022179ee70368448f2f336f4950/services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/SwitchOperationsBolt.java#L213)):_
> 
> > ```java
> >         } catch (PersistenceException e) {
> >             String message = String.format("Port properties not found: %s", e.getMessage());
> >             log.error(message);
> > ```
> 
> Remove this logger call. The exception handler can write something in the log if it want to.
> 

This exception is handled in PersistenceOperationsBolt and message is not logged. I guess we should refactor this separately.

@sergii-iakovenko @surabujin @timofei-durakov 
All other comments are fixed.

rozdy:
> _[services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/SwitchOperationsBolt.java, line 213 at r1](https://reviewable.io/reviews/telstra/open-kilda/2814#-LpYUqH_6x2ECjgXwp5V:-LqLUwih-_UziSgrnOUi:b-fl6zpg) ([raw file](https://github.com/telstra/open-kilda/blob/acf47ca566f1548a09dd1ad331e22ac7991bcca0/services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/SwitchOperationsBolt.java#L213)):_
> Previously, surabujin (Dmitriy Bogun) wrote…
> 
> Bolt is a transport layer. Move this logic(handling the exception and producing some response on in) into service.
> 
> _[services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/SwitchOperationsBolt.java, line 213 at r2](https://reviewable.io/reviews/telstra/open-kilda/2814#-LqBg5A71j_JxwcUw8Vp:-LqLMCagD0b_y07MZk5t:bhs754y) ([raw file](https://github.com/telstra/open-kilda/blob/b36ec016c76a3022179ee70368448f2f336f4950/services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/SwitchOperationsBolt.java#L213)):_
> Previously, rozdy (Ievgen Rozdymakha) wrote…
> 
> Ok. Create issue/pr for this.
> 

Issue #2847 created

PR #2823, opened 2019-09-24
Add notice for autogenerated files

surabujin:
Add notice about autogenerated nature into autogenerated files. The main
goal is to inform user that all modification should be done into
corresponding template file not target/generated file itself.


confd/templates/atdd-staging/kilda.properties.tmpl
```diff
@@ -1,3 +1,6 @@
+# Generated by confd.
+# Do not change this file, all changes will be lost. Change corresponding template.
```
rtretyak: This is already a template, isn't it?

surabujin: Yes... my goal to have this notice into generated file, so it will be obvious that it was generated from template.

I am not sure that I understood your question.

rtretyak: Mmm, ok, I got your point. But at the same time when you open the template file you see "Generated by confd..do not change this file.." header which is very misleading

services/src/grpc-speaker/.gitignore
```diff
@@ -1,13 +1 @@
 /target/
-/logs/
```
timofei-durakov: this is not a part of add notice pr, right?:)

services/storm/.gitignore
```diff
@@ -1,4 +1,4 @@
 # autogenerated config files
 /log4j2/cluster.xml
 /log4j2/worker.xml
-/conf/storm.yaml
```
timofei-durakov: same as above

services/storm/Dockerfile
```diff
@@ -31,6 +31,6 @@ RUN    wget -q https://archive.apache.org/dist/storm/apache-storm-1.1.0/$PACKAGE
 
 WORKDIR /opt/storm/
 
-COPY conf/storm.yaml /opt/storm/conf/storm.yaml
```
timofei-durakov: why do you change this?

surabujin: Because in other case I had to create .force-dir-exists file into service/storm/config directory.

services/storm/.gitignore
surabujin: ok, keeping the mess in place.

PR #2830, opened 2019-09-26
Fixed different timeFirstSeen and TimeLastSeen for Connected Devices

niksv:


services/wfm/src/test/java/org/openkilda/wfm/topology/connecteddevices/service/PacketServiceTest.java
```diff
@@ -0,0 +1,85 @@
...
+        assertEquals(devices.iterator().next().getTimeFirstSeen(), devices.iterator().next().getTimeLastSeen());
+    }
+
+    @Test
+    public void testHandleLldpDataDifferentTimeOnUpdate() {
+        flowCookieRepository.createOrUpdate(new FlowCookie(FLOW_ID, COOKIE));
+        LldpInfoData data = createLldpInfoData(COOKIE);
+        // create
+        packetService.handleLldpData(data);
+        // update
```
rtretyak: sleep for a millisecond? Otherwise looks like another race condition

niksv: make sense

PR #2836, opened 2019-09-30
Delete switch properties when switch is deleted

rozdy:
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2836)

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/services/SwitchOperationsService.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/services/SwitchOperationsService.java
@@ -147,13 +147,17 @@ public class SwitchOperationsService implements ILinkOperationsServiceCarrier {
         Switch sw = switchRepository.findById(switchId)
                 .orElseThrow(() -> new SwitchNotFoundException(switchId));
 
-        if (force) {
-            // forceDelete() removes switch along with all relationships.
-            switchRepository.forceDelete(sw.getSwitchId());
-        } else {
-            // delete() is used to be sure that we wouldn't delete switch if it has even one relationship.
-            switchRepository.delete(sw);
-        }
+        transactionManager.doInTransaction(() -> {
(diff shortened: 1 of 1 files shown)


services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/services/SwitchOperationsService.java
```diff
@@ -151,6 +151,8 @@ public boolean deleteSwitch(SwitchId switchId, boolean force) throws SwitchNotFo
             // forceDelete() removes switch along with all relationships.
             switchRepository.forceDelete(sw.getSwitchId());
         } else {
+            switchPropertiesRepository.findBySwitchId(sw.getSwitchId())
+                    .ifPresent(sp -> switchPropertiesRepository.delete(sp));
             // delete() is used to be sure that we wouldn't delete switch if it has even one relationship.
```
niksv: Maybe we need to delete properties in `force` case too?

rozdy: Fixed

services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/services/SwitchOperationsService.java
```diff
@@ -147,6 +147,8 @@ public boolean deleteSwitch(SwitchId switchId, boolean force) throws SwitchNotFo
         Switch sw = switchRepository.findById(switchId)
                 .orElseThrow(() -> new SwitchNotFoundException(switchId));
 
+        switchPropertiesRepository.findBySwitchId(sw.getSwitchId())
```
timofei-durakov: why is it not in a transaction?

rozdy: Added transaction

PR #2838, opened 2019-10-01
Add perf test for verifying switch validation with a lot of flows

rtretyak:


services/src/performance-tests/src/test/groovy/org/openkilda/performancetests/spec/VolumeSpec.groovy
```diff
@@ -0,0 +1,65 @@
...
+        verifyAll(northbound.validateSwitch(sw.dpId)) {
+            it.verifyRuleSectionsAreEmpty(["excess", "missing"])
+            it.verifyMeterSectionsAreEmpty(["excess", "missing", "misconfigured"])
+        }
+
+        cleanup: "Remove all flows, delete topology"
+        flows.each { northbound.deleteFlow(it.flowId) }
+        Wrappers.wait(flows.size()) {
+            topo.switches.each {
+                assert northbound.getSwitchRules(it.dpId).flowEntries.findAll { !Cookie.isDefaultRule(it.cookie) }.empty
```
andriidovhan: I would also verify meters
```
northbound.getAllMeters(it.dpId).meterEntries.findAll { it.meterId > MAX_SYSTEM_RULE_METER_ID }.isEmpty()
```

rtretyak: This doesn't really matter. I'm more leaning to completely deleting this verification. Anyway the topology is removed afterwards. I was just not sure whether Kilda is doing fine after not being able to remove those rules from the switches cz the switches are gone

PR #2839, opened 2019-10-01
add tests for port discovery property

andriidovhan:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortPropertiesSpec.groovy
```diff
@@ -0,0 +1,250 @@
...
+        Boolean testIsCompleted = true
+
+        and: "Flow is not rerouted"
+        northbound.getFlowPath(flow.id) == newFlowPath
+
+        cleanup: "Restore discovery port property on the port"
+        if (!testIsCompleted) {
+            northboundV2.updatePortProperties(islToManipulate.srcSwitch.dpId, islToManipulate.srcPort,
+                    new PortPropertiesDto(discoveryEnabled: true))
+        }
```
rtretyak: Again, this is not how cleanup should look like. Flow is not removed, `testIsCompleted` usage is not valid

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortPropertiesSpec.groovy
```diff
@@ -0,0 +1,250 @@
...
+    }
```
rtretyak: I understand the point of this test, but I don't understand what flow is doing here. Looks kinda unrelated. It's pretty obvious that the flow is rerouted when ISL fails so I'm not sure what did you try to show here. You mean the event of ISL failure is being 'different' in this case? That's veeery unlikely, I don't think we should test it here

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortPropertiesSpec.groovy
```diff
@@ -0,0 +1,250 @@
...
+    def "Able to manipulate with port properties"() {
+        given: "A port with port properties"
+        // can't use `getAllowedPortsForSwitch` for virtual env in this test,
+        // portProperties validate port number(port number should be in list of '/api/v1/switches/:switch-id/ports')
+        def isl = topology.notConnectedIsls.first()
+        def initPortProperties = northboundV2.getPortProperties(isl.srcSwitch.dpId, isl.srcPort)
+        with(initPortProperties) {
+            it.switchId == isl.srcSwitch.dpId
+            it.portNumber == isl.srcPort
+            it.discoveryEnabled != null
```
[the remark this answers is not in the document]
rozdy: You can use `public static final boolean DISCOVERY_ENABLED_DEFAULT = true;` from PortProperties class here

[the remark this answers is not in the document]
andriidovhan: let's assume that port discovery feature is enabled for all available ports
narrative block was updated with relevant comment `This spec assumes that port discovery feature is enabled for all available ports. `

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortPropertiesSpec.groovy
```diff
@@ -0,0 +1,250 @@
...
+        }
+
+        when: "Update discovery port property"
+        PortPropertiesDto newPortProperties = new PortPropertiesDto()
+        def newDisoveryEnabled = !initPortProperties.discoveryEnabled
```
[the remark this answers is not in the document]
andriidovhan: I think so, we can assume  that all ports are enabled for discovery

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortPropertiesSpec.groovy
```diff
@@ -0,0 +1,250 @@
...
+
+        when: "Try to update discovery port property info for non-existing port"
+        northboundV2.updatePortProperties(sw.dpId, nonExistentPort, new PortPropertiesDto(discoveryEnabled: true))
+
+        then: "Human readable error is returned"
+        def exc = thrown(HttpClientErrorException)
+        exc.statusCode == HttpStatus.NOT_FOUND
+        exc.responseBodyAsString.to(MessageError).errorMessage ==
+                "Port not found: 'Port FSM not found (${sw.dpId}_${nonExistentPort}).'"
+    }
```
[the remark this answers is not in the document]
rozdy: Yes, It's a bit strange from business perspective but from tech perspective port validation for get request would require refactoring

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortPropertiesSpec.groovy
andriidovhan: test was refactored, flow was removed

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortPropertiesSpec.groovy
```diff
@@ -0,0 +1,250 @@
...
+        antiflap.portUp(islToManipulate.srcSwitch.dpId, islToManipulate.srcPort)
+
+        then: "System detects link, state of isl is FAILED due to disabled port discovery property on the src switch"
+        Wrappers.wait(discoveryInterval + WAIT_OFFSET / 2) {
+            assert islUtils.getIslInfo(islToManipulate).get().state == IslChangeType.FAILED
+        }
+
+        when: "Bring port down on the dst switch"
+        antiflap.portDown(islToManipulate.dstSwitch.dpId, islToManipulate.dstPort)
+        sleep(20) // minor sleep for adding stability, otherwise test can be failed after deleting link
```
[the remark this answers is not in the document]
andriidovhan: test was totally refactored

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortPropertiesSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+By default all ports on all switches are available for discovery.
+Admin has ability to enable/disable discovery on a specific port on a switch using Northbound REST API.
+
+This spec assumes that port discovery property is enabled for all available ports.
+""")
+class PortPropertiesSpec extends HealthCheckSpecification {
+    @Autowired
+    NorthboundServiceV2 northboundV2
+
+    def "Able to manipulate with port properties"() {
```
rtretyak: manipulate sbody/sth. "Able to manipulate port properties"

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortPropertiesSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+        def isl = topology.notConnectedIsls.first()
+        with(northboundV2.getPortProperties(isl.srcSwitch.dpId, isl.srcPort)) {
+            it.switchId == isl.srcSwitch.dpId
+            it.portNumber == isl.srcPort
+            it.discoveryEnabled == DISCOVERY_ENABLED_DEFAULT
+        }
+
+        when: "Update port discovery property"
+        PortPropertiesDto newPortProperties = new PortPropertiesDto()
+        def newDiscoveryEnabled = false
```
rtretyak: Well now this should be the opposite to `DISCOVERY_ENABLED_DEFAULT`. Otherwise if the default is changed to 'false' the test will do nothing.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortPropertiesSpec.groovy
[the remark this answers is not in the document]
rtretyak: Any plans for future updates of this behavior? Anyway any strange behavior requires a comment in the test. With or without a link to future plans of improving this I'd like to see an explicit explanation that this is exactly how it is intended to work right now (since the behavior is weird, not obvious)

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortPropertiesSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+        Wrappers.wait(discoveryInterval + WAIT_OFFSET) {
+            assert northbound.getSwitch(sw.dpId).state == SwitchChangeType.ACTIVATED
+            def links = northbound.getAllLinks()
+            (relatedIsls - islToManipulate).each {
+                assert islUtils.getIslInfo(links, it).get().state == IslChangeType.DISCOVERED
+            }
+        }
+
+        then: "Link is still not detected"
+        Wrappers.timedLoop(discoveryInterval + WAIT_OFFSET / 2) {
```
rtretyak: I just realized that actually we can reduce the time of this timedLoop. Since all other switch-related isls are already discovered (L161) we can actually assume that at least `discoveryInterval` time has passed since the switch was activated, can we?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortPropertiesSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+        }
+
+        when: "Enable port discovery property on the src switch"
+        northboundV2.updatePortProperties(islToManipulate.srcSwitch.dpId, islToManipulate.srcPort,
+                new PortPropertiesDto(discoveryEnabled: true))
+
+        then: "Link is detected and link status is FAILED"
+        Wrappers.wait(discoveryInterval + WAIT_OFFSET) {
+            def islInfoForward = islUtils.getIslInfo(islToManipulate).get()
+            def islInfoReverse = islUtils.getIslInfo(islToManipulate.reversed).get()
```
rtretyak: you are dumping all isls twice here. Actually not a big deal, but a cleaner approach would be to use
`northbound.getLink(islToManipulate)` to get only 1 link instead of all OR user the alternative version of getIslInfo that don't query northbound for links
```
def allLinks = northbound.getAllLinks()
def islInfoForward = islUtils.getIslInfo(allLinks, islToManipulate).get()
def islInfoReverse = islUtils.getIslInfo(allLinks, islToManipulate.reversed).get()

```

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortPropertiesSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+        then: "Link is still not detected"
+        Wrappers.timedLoop(discoveryInterval + WAIT_OFFSET / 2) {
+            assert !islUtils.getIslInfo(islToManipulate).isPresent()
+        }
+
+        when: "Enable port discovery property on the src switch"
+        northboundV2.updatePortProperties(islToManipulate.srcSwitch.dpId, islToManipulate.srcPort,
+                new PortPropertiesDto(discoveryEnabled: true))
+
+        then: "Link is detected and link status is FAILED"
```
rtretyak: I'd say here sth about one-way ISL

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortPropertiesSpec.groovy
```diff
@@ -0,0 +1,250 @@
...
+        }
+        Boolean testIsCompleted = true
+
+        cleanup: "Restore discovery port property on the port"
+        if (!testIsCompleted) {
+            northboundV2.updatePortProperties(islToManipulate.srcSwitch.dpId, islToManipulate.srcPort,
+                    new PortPropertiesDto(discoveryEnabled: true))
+            northboundV2.updatePortProperties(islToManipulate.dstSwitch.dpId, islToManipulate.dstPort,
+                    new PortPropertiesDto(discoveryEnabled: true))
+        }
```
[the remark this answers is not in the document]
rtretyak: Ok. probably fine for now, but these blocks look ugly and bulky. We'll have to work out a cleaner solution in the future..
Also, if the cleanup itself fails there should be a mechanism that will prevent all other tests from executing further.

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortPropertiesSpec.groovy
andriidovhan: good idea,
done

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortPropertiesSpec.groovy
andriidovhan: done

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortPropertiesSpec.groovy
andriidovhan: thx, fixed

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortPropertiesSpec.groovy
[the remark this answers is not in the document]
andriidovhan: `cleanup` in this test was removed

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortPropertiesSpec.groovy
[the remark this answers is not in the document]
andriidovhan: comment was added at the beginning of the test

PR #2841, opened 2019-10-02
Allow to use traffexam on python-3.5.2

surabujin:
Avoid usage of merthod `Path.mkdir(mode=0o777, parents=False,
exist_ok=False)` of pathlib module due to lack of `exists_ok` argument
in python-3.5.2.

### DIFF
--- a/services/lab-service/traffexam/kilda/traffexam/context.py
+++ b/services/lab-service/traffexam/kilda/traffexam/context.py
@@ -71,7 +71,7 @@ class Context(object):
     def set_root(self, root):
         self.root = pathlib.Path(root)
         if not self._init_done:
-            self.root.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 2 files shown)


services/lab-service/traffexam/kilda/traffexam/context.py
```diff
@@ -71,7 +72,11 @@ def make_veth_base_name(self):
     def set_root(self, root):
         self.root = pathlib.Path(root)
         if not self._init_done:
-            self.root.mkdir(parents=True, exist_ok=True)
+            try:
+                os.makedirs(str(self.root))
```
timofei-durakov: AFAIR `os.makedirs` has `exist_ok` flag as well why not to use it here?

surabujin: hm... I have missed the moment when it was added there.

PR #2844, opened 2019-10-03
[WIP] Add test that tries to discover an ISL on a flow port

rtretyak:


rtretyak:
Not yet implemented, tests are not required at this point

PR #2864, opened 2019-10-16
Update jackson lib version

rozdy:


services/wfm/src/test/java/org/openkilda/wfm/topology/opentsdb/OpenTsdbTopologyTest.java
```diff
@@ -121,27 +113,4 @@ public void shouldSendDatapointRequestsTwice() throws Exception {
         //verify that request is sent to OpenTSDB server once
         mockServer.verify(HttpRequest.request(), VerificationTimes.exactly(2));
     }
-
```
sergii-iakovenko: Is this required to "Update jackson lib version"?

rozdy: No, this test failed due to lack of jersey-hk2 lib. I have removed all outdated code in this test while tried to figure out what's wrong.

PR #2870, opened 2019-10-22
Remove outdated code

rozdy:


rozdy:
@surabujin 
> You must not produce network(REST) calls from unit test into external services (OTSBD in this specific case).

Reworked test: added correct mocking and request verification. I guess it's not a unit test. We test integration of `DB <-> OpenTSDB topology <-> OTSDB server mock` components so we need to produce some network calls here.

PR #2877, opened 2019-10-23
Fix minor sonar issues

rozdy:


services/wfm/src/main/java/org/openkilda/wfm/topology/flow/service/FlowService.java
```diff
@@ -171,7 +171,6 @@ public FlowPair createFlow(Flow flow, String diverseFlowId, FlowCommandSender se
         try {
             result = (FlowPathsWithEncapsulation) getFailsafe().get(
                     () -> transactionManager.doInTransaction(() -> {
-                        ensureEncapsulationType(flow);
```
timofei-durakov: not clear why have you removed that

rozdy: In previous patch I have added default encapsulation type to Flow model so we don't need to ensure it's not null manually now.

PR #2888, opened 2019-10-25
Test/improve delete flow procedure

andriidovhan:


andriidovhan:
it is already merged as part of #2896

PR #2891, opened 2019-10-28
Increase PortHistorySpec stability

rozdy:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/PortHistorySpec.groovy
```diff
@@ -184,7 +185,12 @@ class PortHistorySpec extends HealthCheckSpecification {
         assumeTrue("Unable to find ISL for this test", isl as boolean)
         def timestampBefore = System.currentTimeMillis()
 
-        when: "Execute port DOWN on the src switch"
+        when: "floodlightRoutePeriodicSync is disabled"
```
andriidovhan: I would move it to the `given` block

```
   given: "floodlightRoutePeriodicSync toggle is disabled"
        northbound.toggleFeature(FeatureTogglesDto.builder().floodlightRoutePeriodicSync(false).build())
   and: "A direct link"
   and: "Execute port DOWN on the src switch"
   when: ...
```

rtretyak: agree. This is not part of the test but rather a precondition/setup

rozdy: Fixed

PR #2892, opened 2019-10-30
Implement H&S update

sergii-iakovenko:
The changes:
- Implement H&S update via V2 API.
- Refactor the flow create, reroute and delete H&S implementation to unify base actions, logging, error handling and history methods.

**MERGE NOTE: merge only after #2869 and change the base to develop before merging.**

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2892)

PR #2897, opened 2019-10-31
fix typo in rules validation while creating a flow

andriidovhan:


sergii-iakovenko:
Note: the issue has been addressed in #2892.

andriidovhan:
fix is already merged with #2892.

PR #2914, opened 2019-11-08
Switch LLDP 3: catching on ISL ports

niksv:
Part of #2917

**IMPORTANT Merge #2918 first**

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/connecteddevices/service/PacketService.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/connecteddevices/service/PacketService.java
@@ -18,13 +18,18 @@ package org.openkilda.wfm.topology.connecteddevices.service;
 import static org.openkilda.model.ConnectedDeviceType.LLDP;
 
 import org.openkilda.messaging.info.event.LldpInfoData;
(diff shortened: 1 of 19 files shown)


services/wfm/src/main/java/org/openkilda/wfm/topology/connecteddevices/service/PacketService.java
```diff
@@ -35,18 +40,23 @@
...
         connectedDeviceRepository = persistenceManager.getRepositoryFactory().createConnectedDeviceRepository();
+        switchConnectedDeviceRepository = persistenceManager.getRepositoryFactory()
+                .createSwitchConnectedDeviceRepository();
     }
 
     /**
      * Handle LLDP info data.
      */
-    public void handleLldpData(LldpInfoData data) {
+    public void handleFlowLldpData(LldpInfoData data) {
```
sergii-iakovenko: Why is infodata class not called FlowLldpInfoData? To be consistent with SwitchLldpInfoData.

niksv: LldpInfoData class will be removed in next patch anyway

PR #2915, opened 2019-11-09
Make http async timeout for nb configurable

timofei-durakov:
- New property `web.request.asyncTimeout` was introduced
- Added to confd templates.

### DIFF
--- a/confd/templates/northbound/northbound.properties.tmpl
+++ b/confd/templates/northbound/northbound.properties.tmpl
@@ -15,6 +15,8 @@ kafka.hosts={{ getv "/kilda_kafka_hosts" }}
 
 server.contextPath=/api
 
+web.request.asyncTimeout=30000
+
 topology.engine.rest.endpoint={{ getv "/kilda_topology_engine_rest_endpoint" }}:{{ getv "/kilda_topology_engine_rest_port" }}
 northbound.messages.expiration.minutes=15
 northbound.kafka.listener.threads=10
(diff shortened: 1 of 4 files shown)


confd/templates/northbound/northbound.properties.tmpl
```diff
@@ -15,6 +15,8 @@ kafka.hosts={{ getv "/kilda_kafka_hosts" }}
 
 server.contextPath=/api
 
+web.request.asyncTimeout=30000
```
surabujin: Other NB options are starts with `northbound.` why this one differ?

services/src/northbound-service/northbound/src/test/resources/northbound.properties
```diff
@@ -12,6 +12,8 @@ kafka.hosts=kafka.pendev:9092
 
 server.contextPath=/api
 
+web.request.asyncTimeout=30000
```
surabujin: Can we have "default" value for this option? So its presence in property file will become optional.

confd/templates/northbound/northbound.properties.tmpl
timofei-durakov: even from this file I could see a plenty of options that starts with something other than northbound prefix

services/src/northbound-service/northbound/src/test/resources/northbound.properties
timofei-durakov: why not to follow pattern that is used here?

confd/templates/northbound/northbound.properties.tmpl
surabujin: ok

PR #2931, opened 2019-11-19
Fix stability of SwitchPropertiesSpec

rtretyak:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchPropertiesSpec.groovy
```diff
@@ -47,6 +47,10 @@ class SwitchPropertiesSpec extends HealthCheckSpecification {
         }
 
         cleanup: "Restore init switch properties on the switch"
+        //wait for previous props change to actually take place
+        Wrappers.wait(RULES_INSTALLATION_TIME) {
+            assert northbound.getSwitchRules(sw.dpId).flowEntries*.cookie.sort() == sw.defaultCookies.sort()
+        }
```
niksv: Do we really need to add it here or maybe after "when:" section?

rtretyak: agree, fixed

PR #2936, opened 2019-11-19
Improve switch properties validation

### LINKED ISSUE #2889: System returns 500 error in case of bad request for 'switches/{switch_id}/properties'
1. Send a malformed payload to `PUT /api/v1/switches/{switch_id}/properties` (remove one of the keys from payload, e.g. `supported_transit_encapsulation`)
**Expected:** 400 Bad Request response
**Actual:** 500 Internal Server Error with html body (null pointer)

rozdy:
Closes #2889

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/services/SwitchOperationsService.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/services/SwitchOperationsService.java
@@ -15,6 +15,8 @@
 
 package org.openkilda.wfm.topology.nbworker.services;
 
+import static org.apache.commons.collections4.CollectionUtils.isEmpty;
+
 import org.openkilda.messaging.model.SwitchPropertiesDto;
 import org.openkilda.messaging.nbtopology.response.GetSwitchResponse;
 import org.openkilda.model.Flow;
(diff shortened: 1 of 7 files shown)


services/wfm/src/main/java/org/openkilda/wfm/topology/nbworker/services/SwitchOperationsService.java
```diff
@@ -252,37 +256,42 @@ public void checkSwitchHasNoIsls(SwitchId switchId) throws IllegalSwitchStateExc
...
      *
      * @param switchId target switch id
      * @param switchPropertiesDto switch properties
+     *
+     * @throws IllegalSwitchPropertiesException if switch properties are incorrect
+     * @throws SwitchPropertiesNotFoundException if switch properties is not found by switch id
      */
     public SwitchPropertiesDto updateSwitchProperties(SwitchId switchId, SwitchPropertiesDto switchPropertiesDto) {
+        if (isEmpty(switchPropertiesDto.getSupportedTransitEncapsulation())) {
+            throw new IllegalSwitchPropertiesException("Supported transit encapsulations should not be null or empty");
```
sergii-iakovenko: Taking into account how you catch and handle IllegalSwitchPropertiesException and SwitchPropertiesNotFoundException, they can be checked exceptions.

rozdy: I think unchecked exception is better here. As for me checked exceptions should be used when we deal with some external resources and we need to force some cleanup. This exception may be ignored without any damage to system so I decided to make it runtime exception.

PR #2937, opened 2019-11-19
Handle invalid encapsulation_type in v2 FlowCreate API

### LINKED ISSUE #2650: can't create flow via APIv2 in case the 'encapsulation_type' field is set in lowerCase
**Step to reproduce**
try to create a flow via APIv2
(NOTE: value of the `encapsulation_type` field should be in lower case   )

```
{
  "flow_id" : "test-encapsulation-type-filed",
  "source" : {
    "switch_id" : "<switch-id>",
    "port_number" : 10,
    "vlan_id" : 2479
  },
  "destination" : {
    "switch_id" : "<switch-id>",
    "port_number" : 10,
    "vlan_id" : 2129
  },
  "status" : "In progress",
  "status_details" : null,
  "maximum_bandwidth" : 500,
  "ignore_bandwidth" : false,
  "periodic_pings" : false,
  "description" : "autotest flow with depressed",
  "max_latency" : null,
  "priority" : null,
  "diverse_with" : null,
  "pinned" : true,
  "allocate_protected_path" : false,
  "encapsulation_type" : "vxlan",
  "created" : "2019-07-31T12:22:31.173Z",
  "last_updated" : null
}
```

**Actual result:**
An error is returned
`Operation has timed out`

**Expected result:**
flow should be created
system should be able to proceed lower and upper case

niksv:
Closes #2650

**MERGE NOTES: Merge after #2952! Change the base to develop before.**

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowCrudV2Spec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowCrudV2Spec.groovy
@@ -558,6 +558,21 @@ class FlowCrudV2Spec extends HealthCheckSpecification {
         flowHelperV2.deleteFlow(flow.flowId)
     }
 
+    def "Unable to create a flow with invalid encapsulation type"() {
+        given: "A flow with invalid encapsulation type"
+        def (Switch srcSwitch, Switch dstSwitch) = topology.activeSwitches
+        def flow = flowHelperV2.randomFlow(srcSwitch, dstSwitch)
+        flow.setEncapsulationType("fake")
+
+        when: "Try to create a flow"
+        flowHelperV2.addFlow(flow)
+
+        then: "Flow is not created"
+        def exc = thrown(HttpClientErrorException)
+        exc.rawStatusCode == 400
+        exc.responseBodyAsString.to(MessageError).errorDescription == "Can not parse arguments of the create flow request"
+    }
+
 	def "Unable to update a flow with invalid encapsulation type"() {
 		given: "A flow with invalid encapsulation type"
 		def (Switch srcSwitch, Switch dstSwitch) = topology.activeSwitches
--- a/services/src/northbound-service/northbound/src/main/java/org/openkilda/northbound/service/impl/FlowServiceImpl.java
+++ b/services/src/northbound-service/northbound/src/main/java/org/openkilda/northbound/service/impl/FlowServiceImpl.java
@@ -170,7 +170,7 @@ public class FlowServiceImpl implements FlowService {
         try {
             payload = new FlowCreateRequest(new FlowDto(input), input.getDiverseFlowId());
         } catch (IllegalArgumentException e) {
-            logger.error("Can not parse arguments: {}", e.getMessage());
+            logger.error("Can not parse arguments: {}", e.getMessage(), e);
             throw new MessageException(correlationId, System.currentTimeMillis(), ErrorType.DATA_INVALID,
                     e.getMessage(), "Can not parse arguments of the create flow request");
         }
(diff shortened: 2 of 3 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowCrudV2Spec.groovy
```diff
@@ -494,6 +495,37 @@ class FlowCrudV2Spec extends HealthCheckSpecification {
         ]
     }
 
+    def "Able to create a flow with vlan encapsulation type in lower case"() {
```
rtretyak: Not the test case we want to see in 'functional-tests'. We are making specifications, i.e. describing/documenting correct behavior of our features or common use-case scenarios. This hardly can be called a documented feature or a real scenario. 
We are _not_ covering every small defect with explicit single-target tests, not here. Very narrow problem, this test is doomed to never catch any new defects, just increases the run time.

Option1: Make a real functional scenario that will reveal this exact problem. the test could have been named: `User can use 'get flow' response of existing flow to create a new flow`, which is an actual real-life scenario that will cover any other potential problems, not only flow encapsulation type lower case.
Problem with option1 is that it will only reveal issue when the v1->v2 mapping is on. Which is how the defect happens, which is the real problem. But mapping is hard to test, since it is a temporary thing, available only on staging. But the test still can be made and will catch the issue when run against the staging env

Option2 - just remove it, not worth automating (difficult due to need of mapping). Defect was successfully caught and raised without the autotest, so manual testing was enough.

niksv: I choose option 2 :) 
What about second test with Bad request? Remove it too?

rtretyak: Not sure. It's fine I guess. At least until we cover it at unit test level

services/src/northbound-service/northbound/src/main/java/org/openkilda/northbound/service/impl/FlowServiceImpl.java
```diff
@@ -186,8 +187,19 @@
...
-                System.currentTimeMillis(), RequestCorrelationId.getId(), Destination.WFM);
+        final String correlationId = RequestCorrelationId.getId();
+        FlowRequest flowRequest;
+
+        try {
+            flowRequest = flowMapper.toFlowCreateRequest(request);
+        } catch (IllegalArgumentException e) {
+            logger.error(format("Can not parse arguments: %s", e.getMessage()), e);
+            throw new MessageException(correlationId, System.currentTimeMillis(), ErrorType.DATA_INVALID,
+                    e.getMessage(), "Can not parse arguments when create flow request");
```
rtretyak: `when creating`

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowCrudV2Spec.groovy
niksv: fixed

services/src/northbound-service/northbound/src/main/java/org/openkilda/northbound/service/impl/FlowServiceImpl.java
niksv: fixed

services/src/northbound-service/northbound/src/main/java/org/openkilda/northbound/service/impl/FlowServiceImpl.java
```diff
@@ -168,9 +169,9 @@
         try {
             payload = new FlowCreateRequest(new FlowDto(input), input.getDiverseFlowId());
         } catch (IllegalArgumentException e) {
-            logger.error("Can not parse arguments: {}", e.getMessage());
+            logger.error(format("Can not parse arguments: %s", e.getMessage()), e);
```
dpoltavets: Why do you use this way of composing a message?

niksv: fixed

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowCrudV2Spec.groovy
```diff
@@ -494,6 +494,21 @@ class FlowCrudV2Spec extends HealthCheckSpecification {
         ]
     }
 
+    def "Unable to create a flow with invalid encapsulation type"() {
```
sergii-iakovenko: What about a value with lower/upper/mixed casa for encapsulation type?

niksv: Func tests with such cases were rejected by QA team

Issue #2941, opened 2019-11-20
Updating encap type for switch should respect switch features


rtretyak:
As a system, I want to forbid turning on `vxlan` encap type in switch properties if the switch does not actually support vxlan, so that system doesn't get into inconsistent state when switch vxlan is enabled in Kilda, but not acutally supported by switch.

AC:
- When encap type is being updated in switch properties a corresponding validation is run that check switch features and verify that the switch actually supports requested encap type.
- Descriptive error is returned if above validation fails, property does not get updated.

Issue #2932, opened 2019-11-19
system is allowing to enable the 'multiTable' mode for unsupported switches


andriidovhan:
**Step to reproduce:**
* try to enable the 'multiTable' mode for an unsupported switch(e.g. Accton or Centec)

**Actual result:**
MultiTable mode is enabled

**Expected result:**
MultiTable mode shouldn't be enabled
System should return an error(e.g Bad Request)

PR #2945, opened 2019-11-20
Valdate Switch Props against supported features

timofei-durakov:
perform validation of required features provided by switch
while setting switch properties. Throw bad request in case of
failures over northbound.

Closes: #2941
Closes: #2932

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchPropertiesSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchPropertiesSpec.groovy
@@ -1,12 +1,16 @@
 package org.openkilda.functionaltests.spec.switches
 
+import static org.junit.Assume.assumeTrue
+import static org.openkilda.functionaltests.extension.tags.Tag.TOPOLOGY_DEPENDENT
 import static org.openkilda.testing.Constants.NON_EXISTENT_SWITCH_ID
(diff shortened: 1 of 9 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchPropertiesSpec.groovy
```diff
@@ -28,10 +28,7 @@ class SwitchPropertiesSpec extends HealthCheckSpecification {
         when: "Update switch properties"
         SwitchPropertiesDto switchProperties = new SwitchPropertiesDto()
         def newMultiTable = !initSwitchProperties.multiTable
-        def newTransitEncapsulation = (initSwitchProperties.supportedTransitEncapsulation.size() == 1) ?
-                [FlowEncapsulationType.TRANSIT_VLAN.toString().toLowerCase(),
-                 FlowEncapsulationType.VXLAN.toString().toLowerCase()].sort() :
-                [FlowEncapsulationType.VXLAN.toString().toLowerCase()]
+        def newTransitEncapsulation = [FlowEncapsulationType.TRANSIT_VLAN.toString().toLowerCase()]
```
andriidovhan: As for would be better to change the procedure of selecting switch for manipulation
`def sw = topology.activeSwitches.find { it.features.contains(SwitchFeature.NOVIFLOW_COPY_FIELD) }`

otherwise we can't be sure that we are really able to update switch properties.

PR #2953, opened 2019-11-22
improve swapEndpointSpec

andriidovhan:
sometimes for no good reason swapEndpointSpec fails on jenkins
```
java.lang.NullPointerException: Cannot get property 'src' on null object
01:09:38.540  	at org.openkilda.functionaltests.spec.flows.SwapEndpointSpec.getFirstFlow(SwapEndpointSpec.groovy:1195)
```
that PR should prevent that


andriidovhan:
> Not a correct fix. The switch-search algorithm is broken and should be fixed. Instead you propose just to ignore tests.

fixed

Issue #2955, opened 2019-11-22
system is trying to update encapsulation type for a flow to not supported on a switch while syncing/rerouting a flow


andriidovhan:
**Steps to reproduce** 
* enable `flows_reroute_using_default_encap_type` in feature toggle;
* select  a switch which support `transit_vlan` only;
*  create a flow with `transit_vlan`;
* update the kilda configuration (`"flow_encapsulation_type": "vxlan"`);
* update the flow(`encapsulation-type": null`);

**Actual result**
System is tried to update encapsulation type for the flow according to global configuration, but switch doesn't support this kind of encapsulation. 
Flow status is DOWN,
flow is pingable

**Expected result**
Flow state should not be changed to DOWN

**Workaround**
disable `flows_reroute_using_default_encap_type` in feature toggle and sync the flow

sergii-iakovenko:
The issue is actual and reproducible.

niksv:
Root cause:
If `flows_reroute_using_default_encap_type` is set System tries to reroute flow **only** with this encap type. 

There are 2 ways to fix this issue:
1. Add global reroute retry if system can't find a path with `flows_reroute_using_default_encap_type` and try to find path with origin encap type.
2. Add local retry for path computation (with different encap types) or even call path computation engine with set of possible options like: ignore bandwidth or not, which encap types can be used, which backup strategies can be used. In other words allow path computation engine find not optimal way.

Option 1 is a bad idea because it's a workaround and it adds a lot of retries (spending time).
So we decided to implement option 2. But to do it we need to move encapsulation type field from Flow to FlowPath to allow Flow to have primary and protected paths with different encapsulation types

PR #2958, opened 2019-11-25
add test for flows_reroute_using_default_encap_type feature toogle

andriidovhan:
covers: #2955

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/toggles/FeatureTogglesV2Spec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/toggles/FeatureTogglesV2Spec.groovy
@@ -1,15 +1,23 @@
 package org.openkilda.functionaltests.spec.toggles
 
+import static org.junit.Assume.assumeTrue
 import static org.openkilda.functionaltests.extension.tags.Tag.SMOKE
 import static org.openkilda.functionaltests.extension.tags.Tag.VIRTUAL
+import static org.openkilda.testing.Constants.WAIT_OFFSET
 
 import org.openkilda.functionaltests.HealthCheckSpecification
 import org.openkilda.functionaltests.extension.tags.Tags
+import org.openkilda.functionaltests.helpers.Wrappers
 import org.openkilda.messaging.error.MessageError
+import org.openkilda.messaging.info.event.IslChangeType
 import org.openkilda.messaging.model.system.FeatureTogglesDto
+import org.openkilda.messaging.model.system.KildaConfigurationDto
+import org.openkilda.messaging.payload.flow.FlowState
+import org.openkilda.model.FlowEncapsulationType
 
 import org.springframework.http.HttpStatus
 import org.springframework.web.client.HttpClientErrorException
+import spock.lang.Ignore
 import spock.lang.Narrative
 
 @Narrative("""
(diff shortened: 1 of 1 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/toggles/FeatureTogglesV2Spec.groovy
```diff
@@ -112,4 +122,159 @@ class FeatureTogglesV2Spec extends HealthCheckSpecification {
         then: "Able to delete flows"
         [flow, newFlow].each { flowHelperV2.deleteFlow(it.flowId) }
     }
+
+    @Ignore("https://github.com/telstra/open-kilda/issues/2955")
+    @Tags(HARDWARE)
```
rtretyak: why?

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/toggles/FeatureTogglesV2Spec.groovy
```diff
@@ -112,4 +122,159 @@ class FeatureTogglesV2Spec extends HealthCheckSpecification {
...
+    def "Flow encapsulation type is changed while auto rerouting according to 'flows_reroute_using_default_encap_type' \
+feature toogle"() {
+        given: "A switch pair which supports transit_vlan and vxlan encapsulation types"
+        def swPair = topologyHelper.getAllNeighboringSwitchPairs().find {
+            [it.src, it.dst].every { it.noviflow && !it.wb5164 } && it.paths.size() >= 2
```
rtretyak: I understand what you mean, but this is an 'indirect' search. 'noviflow' switch just means that the manufacturer is 'Noviflow', it is not directly say anything about vxlan. Instead, you should look for 'COPY_FIELD' feature OR look at the switch properties allowed encapsulation type (not sure what is better atm. I guess the second option is better given #2941 is implemented)

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/toggles/FeatureTogglesV2Spec.groovy
```diff
@@ -112,4 +122,159 @@ class FeatureTogglesV2Spec extends HealthCheckSpecification {
...
+            [it.src, it.dst].every { !it.features.contains(SwitchFeature.NOVIFLOW_COPY_FIELD) } && it.paths.size() >= 2
+        }
+        assumeTrue("Unable to find required switches in topology", swPair as boolean)
+
+        and: "The 'flows_reroute_using_default_encap_type' feature is enabled"
+        def initFeatureToggle = northbound.getFeatureToggles()
+        !initFeatureToggle.flowsRerouteUsingDefaultEncapType && northbound.toggleFeature(FeatureTogglesDto.builder()
+                .flowsRerouteUsingDefaultEncapType(true).build())
+
+        and: "Default encapsulation type is transit_vlan"
```
rtretyak: This is not required step. You can directly create the flow as 'transit vlan' and then change default config to 'vxlan'

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/toggles/FeatureTogglesV2Spec.groovy
```diff
@@ -112,4 +122,159 @@ class FeatureTogglesV2Spec extends HealthCheckSpecification {
...
+        def initGlobalConfig = northbound.getKildaConfiguration()
+        (initGlobalConfig.flowEncapsulationType == FlowEncapsulationType.TRANSIT_VLAN.toString().toLowerCase()) ?:
+                northbound.updateKildaConfiguration(
+                        new KildaConfigurationDto(flowEncapsulationType: FlowEncapsulationType.TRANSIT_VLAN)
+                )
+
+        and: "A flow with default encapsulation"
+        def initKildaConfig = northbound.getKildaConfiguration()
+        def flow = flowHelperV2.randomFlow(swPair).tap { encapsulationType == null }
```
rtretyak: `encapsulationType == null` is not an assignment

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/toggles/FeatureTogglesV2Spec.groovy
```diff
@@ -112,4 +122,159 @@ class FeatureTogglesV2Spec extends HealthCheckSpecification {
...
+        assumeTrue("Unable to find required switches in topology", swPair as boolean)
+
+        and: "The 'flows_reroute_using_default_encap_type' feature is enabled"
+        def initFeatureToggle = northbound.getFeatureToggles()
+        !initFeatureToggle.flowsRerouteUsingDefaultEncapType && northbound.toggleFeature(FeatureTogglesDto.builder()
+                .flowsRerouteUsingDefaultEncapType(true).build())
+
+        and: "A flow with default encapsulation"
+        def initKildaConfig = northbound.getKildaConfiguration()
+        def flow = flowHelperV2.randomFlow(swPair).tap { encapsulationType == null }
```
rtretyak: `encapsulationType == null` is not an assignment

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/toggles/FeatureTogglesV2Spec.groovy
andriidovhan: before I tried to find a noviflow switch
for now it is fixed, `HARDWARE` tag is removed

services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/toggles/FeatureTogglesV2Spec.groovy
rtretyak: On the other hand, _feature_ is more of a background property available only in database which can be changed, i.e. has no 'contract', while switch props are part of API. Also, I'm wondering if we can actually build a vxlan flow on a switch, which has COPY_FIELD but doesn't have VXLAN encap type enabled. Can you check that?

andriidovhan: I will check it
I believe that system will not create a vxlan flow

rtretyak: System looks on switch properties, not on switch features. you should do the same

rtretyak: Switch can have COPY_FIELD feature, but have a turned off VXLAN encap type property. In this case this switch is considered as a non-vxlan. Your test does not account for this situation

andriidovhan: thx, fixed

PR #2966, opened 2019-11-25
Fix northbound logging after update log4j to 2.11.0

nikitamarchenko:
The update containing breaking change. They removed field timeMillis
from JsonLayout.
https://issues.apache.org/jira/browse/LOG4J2-1883#comment-16342097

### DIFF
--- a/confd/templates/northbound/log4j2.xml.tmpl
+++ b/confd/templates/northbound/log4j2.xml.tmpl
@@ -14,6 +14,7 @@ Do not change this file, all changes will be lost. Change corresponding template
                  filePattern="{{ getv "/kilda_logging_logfile_path" }}/northbound.log.json.%i.gz">
            <JsonLayout compact="true" eventEol="true" properties="true">
                 <KeyValuePair key="correlation_id" value="${ctx:correlation_id}"/>
+                <KeyValuePair key="timeMillis" value="$${timeMillis:}"/>
            </JsonLayout>
            <Policies>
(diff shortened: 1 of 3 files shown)


confd/templates/northbound/log4j2.xml.tmpl
```diff
@@ -14,6 +14,7 @@ Do not change this file, all changes will be lost. Change corresponding template
                  filePattern="{{ getv "/kilda_logging_logfile_path" }}/northbound.log.json.%i.gz">
            <JsonLayout compact="true" eventEol="true" properties="true">
                 <KeyValuePair key="correlation_id" value="${ctx:correlation_id}"/>
+                <KeyValuePair key="timeMillis" value="$${timeMillis:}"/>
```
sergii-iakovenko: Have you already tried the following solution (similar to https://issues.apache.org/jira/browse/LOG4J2-1752)?
```<KeyValuePair key="timeMillis" value="$${date:UNIX_MILLIS}" />```

Comment (nikitamarchenko) [reply confd/templates/northbound/log4j2.xml.tmpl]: > Have you already tried the following solution (similar to https://issues.apache.org/jira/browse/LOG4J2-1752)?
> `<KeyValuePair key="timeMillis" value="$${date:UNIX_MILLIS}" />`
I spend 5h to find sollution. About unix_millis we have 
```

nikitamarchenko: idk how to get unix timestamp from that https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html

PR #2968, opened 2019-11-26
Fix NPE in H&S reroute and update thrown for a flow with no paths

### LINKED ISSUE #2954: V2 Flow Creation creates damaged flow if one of the endpoints disconnects in the process
1. Init flow creation via V2
2. Immediately disconnect src or dst switch before the rules are actually installed
3. At this point flow should be created and have DOWN status.
4. return the switch back
**Expected:** Flow can be manually rerouted and become UP or flow is automatically rerouted when switch returns back (depends on whether any ISL got failed due to timeout when switch was disconnected)
**Actual:** Flow remains DOWN and cannot be rerouted. Intentional reroute returns Internal Server Error. [unable_to_reroute_null_pointer.txt](https://github.com/telstra/open-kilda/files/3879344/unable_to_reroute_null_pointer.txt)

note: test script that reproduces this problem is available in the repo, linked to this ticket

sergii-iakovenko:
The changes:
- Proper handling of a flow with no paths or with incomplete pairs. 
- Improve encapsulation in FlowRerouteFsm and FlowUpdateFsm to make checks for set / not set paths clear.

Closes #2954, #3028

**MERGE NOTES: Don't forget to change the base!**

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2968)


sergii-iakovenko:
The PR is not actual and replaced with #3459

PR #2976, opened 2019-11-28
Check flow response actual type before casting

timofei-durakov:
this patch will mitigate errors caused by class cast errors in
RuleValidationAction during the H&S re-route


surabujin:
Such code is present not only in reroute, but in all other H&S operations.

sergii-iakovenko:
The issue doesn't affect the current develop. These changes are not required.

timofei-durakov:
No longer valid

PR #2992, opened 2019-12-02
Refactor multi-reroute spec to work with bigger amount of flows

rtretyak:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/MultiRerouteSpec.groovy
```diff
@@ -7,66 +7,79 @@ import static org.openkilda.testing.Constants.WAIT_OFFSET
...
             flows << flow
         }
-        def currentPath = pathHelper.convert(northbound.getFlowPath(flows[0].id))
-        //ensure both flows are on the same path
-        assert pathHelper.convert(northbound.getFlowPath(flows[1].id)) == currentPath
+        def currentPath = pathHelper.convert(northbound.getFlowPath(flows[0].flowId))
+        //ensure all flows are on the same path
+        flows[1..-1].each {
+            assert pathHelper.convert(northbound.getFlowPath(it.flowId)) == currentPath
+        }
```
andriidovhan: should we check that 20 flows were created?

rtretyak: Why wouldn't they? FlowHelper kind of guarantees that all flows are created, why would we recheck this here? Didn't get your idea

andriidovhan: >  FlowHelper kind of guarantees that all flows are created

agree

PR #2993, opened 2019-12-03
Fix false-negative meter's validation

surabujin:
Existing implementation does not take into account meter burst deviation (+-1) for accurate meters. As a result it leads to false-negative flow segment validation on some subset of switches.
    
This fix adds such burst behaviour peculiarity of accurate meters.


services/src/kilda-core/kilda-model/src/main/java/org/openkilda/model/of/MeterSchemaBand.java
```diff
@@ -68,6 +68,16 @@ public int hashCode() {
...
 
+    private boolean inaccurateEquals(long left, long right, long deviation) {
+        long diff;
+        if (left < right) {
+            diff = right - left;
+        } else {
+            diff = left - right;
+        }
+        return diff <= deviation;
+    }
```
niksv: `return Math.abs(left - right) <= deviation`

surabujin: too complex :)

PR #3005, opened 2019-12-06
Fix flow segment validation relaxing set field action match

surabujin:
Some switch return not accurate set field action for vlan-vid field. Returned action will have no "presence" VLAN bit set. It leads to flow missmatch by our flow-segment-verify code.
    
Add new feature to mark such switches and relax set field match if this feature is present.

### DIFF
--- a/services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/utils/IOfFlowDumpProducer.java
+++ b/services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/utils/IOfFlowDumpProducer.java
@@ -13,19 +13,18 @@
  *   limitations under the License.
  */
 
-package org.openkilda.model;
+package org.openkilda.floodlight.utils;
 
-public enum SwitchFeature {
-    METERS, INACCURATE_METER,
-    BFD,
-    BFD_REVIEW,
-    GROUP_PACKET_OUT_CONTROLLER,
-    RESET_COUNTS_FLAG,
-    LIMITED_BURST_SIZE,
-    NOVIFLOW_COPY_FIELD,
-    PKTPS_FLAG,
-    MATCH_UDP_PORT,
-    MAX_BURST_COEFFICIENT_LIMITATION,
-    MULTI_TABLE,
-    INACCURATE_SET_VLAN_VID_ACTION
+import org.projectfloodlight.openflow.protocol.OFFlowStatsEntry;
+import org.projectfloodlight.openflow.types.DatapathId;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+public interface IOfFlowDumpProducer {
+    List<CompletableFuture<List<OFFlowStatsEntry>>> getTableRequests();
+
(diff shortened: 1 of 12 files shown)


services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/utils/IOfFlowDumpProducer.java
```diff
@@ -0,0 +1,30 @@
...
+
+package org.openkilda.floodlight.utils;
+
+import org.projectfloodlight.openflow.protocol.OFFlowStatsEntry;
+import org.projectfloodlight.openflow.types.DatapathId;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+public interface IOfFlowDumpProducer {
```
sergii-iakovenko: As per OpenKilda code convention (which is based on Google Java Style), special prefixes or suffixes are NOT used. Please, reconsider the usage of  "I" prefix here.

services/src/kilda-core/kilda-model/src/main/java/org/openkilda/model/SwitchFeature.java
```diff
@@ -26,5 +26,6 @@
     PKTPS_FLAG,
     MATCH_UDP_PORT,
     MAX_BURST_COEFFICIENT_LIMITATION,
-    MULTI_TABLE
+    MULTI_TABLE,
+    INACCURATE_SET_VLAN_VID_ACTION
```
sergii-iakovenko: Are we sure that a switch performs "inaccurate set" operation? Or it's rather "inproper data" returned on the dump operation?

Also, what is exactly "inaccurate" in the vlan returned? You're still able to obtain the exact vlan value from OFOxmVlanVid object.

surabujin: It is inaccurate from OF spec perspective (maybe inaccurate is not best term here, propose another if you have some in mind). Command sent to the switch set "presence" bit in this field, but read operation do not return this bit. I suspect that switch completely ignores this "VLAN presence bit". We can perfome more test later when this issue will not affect our functionality.

services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/utils/IOfFlowDumpProducer.java
surabujin: It means, we should tune style check to deny such names. :) I will rename it, but probably not in this fix.

services/src/kilda-core/kilda-model/src/main/java/org/openkilda/model/SwitchFeature.java
sergii-iakovenko: Ok. Please, create an issue for this.

surabujin:
Later research should be tracked into https://github.com/telstra/open-kilda/issues/3011

Issue #2894, opened 2019-10-30
Add PCE strategy based on latency


pzakatov:
# Scope
1. Add a PCE strategy that takes ISL latency as the main weight parameter instead of ISL cost. 
1. Add the ability to specify the PCE strategy on flow and system level.

PR #3036, opened 2019-12-12
Support path computation strategy param in partial flow update

rozdy:
Related to #2894

### DIFF
--- a/src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/update/actions/UpdateFlowAction.java
+++ b/src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/update/actions/UpdateFlowAction.java
@@ -83,6 +83,7 @@ public class UpdateFlowAction extends NbTrackableAction<FlowUpdateFsm, State, Ev
 
     private RequestedFlow updateFlow(Flow flow, RequestedFlow targetFlow, FlowUpdateFsm stateMachine) {
         RequestedFlow originalFlow = RequestedFlowMapper.INSTANCE.toRequestedFlow(flow);
+        stateMachine.setOldTargetPathComputationStrategy(flow.getTargetPathComputationStrategy());
         stateMachine.setOriginalFlow(originalFlow);
 
         stateMachine.setOriginalFlowGroup(flow.getGroupId());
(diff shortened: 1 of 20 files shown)


rozdy:
> 
> > ```java
> >     /**
> >      * Updates max latency, priority or path computation strategy of existing flow.
> > ```
> 
> AS for me - list all modifiable fields here is bad idea.

Fixed

> 
> > ```java
> >      * Convert {@link String} to {@link PathComputationStrategy}.
> >      */
> >     default PathComputationStrategy mapPathComputationStrategy(String pathComputationStrategy) {
> > ```
> 
> I believe mapstruct is able to produce this mapper by it's own.

Yep but auto generated one is case-sensitive.

sergii-iakovenko:
Please, address the review comments.

rozdy:
> What's the point in this pr? If operator changes pce strategy for the flow, it's expected to be applied right now, which means -> new path computation, and the heavy update. Does it make sense to overwrite actual state this way?

Reworked patch. Introduced target path computation param for partial update.

rozdy:
@sergii-iakovenko 
> This PR contains multiple changes of just renaming pathComputationStrategy -> targetPathComputationStrategy. What is the purpose of this? Why do we need to have targetPathComputationStrategy in the model's Flow?

The idea behind this PR is to provide ability to change flow path computation strategy without immediate reroute. New strategy (called "target") should be used with next reroute or full flow update. Old strategy is saved to provide consistency between `pathComputationStrategy` flow parameter and strategy used for current path computation. So a new field `targetPathComputationStrategy` is introduced. From API perspective old one should be considered as read-only flow param and new one should be used in payloads for flow update/patch.

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/update/actions/UpdateFlowAction.java
```diff
@@ -82,8 +82,7 @@ public UpdateFlowAction(PersistenceManager persistenceManager, int transactionRe
     }
 
     private RequestedFlow updateFlow(Flow flow, RequestedFlow targetFlow, FlowUpdateFsm stateMachine) {
-        RequestedFlow originalFlow = RequestedFlowMapper.INSTANCE.toRequestedFlow(flow);
-        stateMachine.setOriginalFlow(originalFlow);
+        stateMachine.setOriginalFlow(RequestedFlowMapper.INSTANCE.copy(flow));
```
dpoltavets: Need to check that this change will not return the already resolved bugs associated with multi table.

rozdy: Fixed

PR #3037, opened 2019-12-12
Updated Dockerfile for building neo4j from off.site binaries

PinGwynn:
Updated Dockerfile for building neo4j from off.site binaries instead of debian repo

### DIFF
--- a/services/neo4j/Dockerfile
+++ b/services/neo4j/Dockerfile
@@ -23,25 +23,25 @@ ENV NEO4j_VERSION="${neo4j_version}"
 ENV APOC_VERSION=${apoc_version}
 ENV GRAPH_ALGO_VERSION=${graph_algo_version}
 
-RUN wget -O - https://debian.neo4j.org/neotechnology.gpg.key | apt-key add - \
-    && echo 'deb https://debian.neo4j.org/repo stable/' | tee /etc/apt/sources.list.d/neo4j.list \
-    && apt-get update -q \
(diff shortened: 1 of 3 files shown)


services/neo4j/Dockerfile
```diff
@@ -23,25 +23,24 @@ ENV NEO4j_VERSION="${neo4j_version}"
 ENV APOC_VERSION=${apoc_version}
 ENV GRAPH_ALGO_VERSION=${graph_algo_version}
 
-RUN wget -O - https://debian.neo4j.org/neotechnology.gpg.key | apt-key add - \
-    && echo 'deb https://debian.neo4j.org/repo stable/' | tee /etc/apt/sources.list.d/neo4j.list \
-    && apt-get update -q \
-    && apt-get install -yq --no-install-recommends \
-        neo4j=1:${NEO4j_VERSION} \
+RUN wget https://neo4j.com/artifact.php?name=neo4j-community-${neo4j_version}-unix.tar.gz -O /tmp/neo4j-community-${neo4j_version}-unix.tar.gz \
```
andriidovhan: how about adding `-q` option?

services/neo4j/init/init.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+PATH=/opt/neo4j-community-3.3.9/bin:$PATH
```
surabujin: Why version is hardcoded here? Can we use env-variable defined in Docker file here?

PR #3046, opened 2019-12-16
Add minor test that changes switch table mode rapidly

rtretyak:


rtretyak:
Don't like the test, does not catch the issue we have

PR #3058, opened 2019-12-18
add test "Flow ping can detect a broken path for a vxlan flow on an intermediate switch"

andriidovhan:
Covers: #3069

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowPingSpec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowPingSpec.groovy
@@ -2,6 +2,7 @@ package org.openkilda.functionaltests.spec.flows
 
 import static com.shazam.shazamcrest.matcher.Matchers.sameBeanAs
 import static org.junit.Assume.assumeTrue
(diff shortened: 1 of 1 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowPingSpec.groovy
```diff
@@ -239,6 +242,45 @@ class FlowPingSpec extends HealthCheckSpecification {
         }
     }
 
+    @Tidy
+    @Tags(HARDWARE)
+    def "Flow ping can detect a broken path for a vxlan flow on an intermediate switch"() {
+        given: "A vxlan flow with intermediate switch(es)"
+        def switchPair = topologyHelper.getAllNotNeighboringSwitchPairs().find { swP ->
+            swP.paths.findAll { path ->
+                pathHelper.getInvolvedSwitches(path).every { it.noviflow  && !it.wb5164 }
```
rtretyak: You must select by looking at switch properties. This search does not guarantee that switches are vxlan-enabled

andriidovhan: I forgot about that, thx.
 fixed

PR #2962, opened 2019-11-25
Introducing flow with 2 levels of VLAN tagging on endpoints

surabujin:
This change adds the ability to use 2 levels of VLAN tags on flow endpoints.  I.e. you can define 2 (outer and inner) VLAN tags for each flow endpoint.

Flow's endpoints (source and dest) encoding are independent, so it can be defined as "default port" (no VLAN tags), as VLAN tagged and as QinQ or double VLAN tagged for each end of the flow.

OF spec allow to the switch to reject flows that manipulate with possible missing ethernet packet's field. If the OF flow's action list contains pop VLAN tag(for example) action it must have VLAN tag match in it's match field (see section 7.2.6 Action Structures OF specification v1.5.1 for more details). As a result we can't do the flow's endpoint encapsulation on egress switch as before, we need to unwrap ingress endpoint encapsulation on egress switch and apply/wrap egress encapsulation on egress switch.

So ingress switch will drop all(except last in case of transit-VLAN encapsulation) VLAN tags from ethernet packet, wrap it into transit encapsulation and on egress switch transit encapsulation will be unwrapped and corresponding egress endpoint VLAN tags are added.

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/2962)

Fixes #2949

PR #3078, opened 2019-12-23
Proper vlan reencoding on ingress flow segment

surabujin:
Use `org.openkilda.floodlight.utils.OfAdapter.makeVlanReplaceActions` for
produce VLAN reencoding on ingress flow segment. This tool correctly
process case when current and desired VLAN tags match to each other.


surabujin:
This fix already included into #2962

Issue #3077, opened 2019-12-23
Flow validation fails when transit vlan equals to customer vlan on egress rule


rtretyak:
when a flow vlan equals to the assigned transit vlan, the egress rule will not overwrite it in the output command. This leads to flow validation seeing this rule as inaccurate:
```
"discrepancies": [
            {
                "rule": "{sw:00:00:e9:6d:e8:d2:48:a9, ck:2305843009213694136, in:1-185, out:15-185}",
                "field": "outVlan",
                "expected_value": "185",
                "actual_value": "0"
            }
        ]
```

**Expected behavior:**
1. Flow validation should not complain about this format of egress/ingress rules. It is okay that the transit vlan is not reassigned if it already equals to customer vlan
2. Check ingress rule behavior: it should not rewrite customer vlan value if it already equals to the chosen transit vlan (behave the same as egress rule does now)

**Note:**
The most common scenario for this: Create a single-switch flow on different ports but use same vlan on src and dst

PR #3079, opened 2019-12-24
Add test for flow with endpoint vlan matching transit vlan

rtretyak:
This reveals the #3077 issue

### DIFF
--- a/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowCrudV2Spec.groovy
+++ b/services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowCrudV2Spec.groovy
@@ -1252,6 +1252,54 @@ class FlowCrudV2Spec extends HealthCheckSpecification {
         }
     }
 
+    @Ignore("https://github.com/telstra/open-kilda/issues/3077")
+    @Tidy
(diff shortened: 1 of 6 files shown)


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowCrudV2Spec.groovy
```diff
@@ -1193,6 +1193,48 @@ class FlowCrudV2Spec extends HealthCheckSpecification {
...
+        northbound.validateFlow(flow.flowId).each { direction -> assert direction.asExpected }
+
+        and: "Involved switches pass switch validation"
+        pathHelper.getInvolvedSwitches(flow.flowId).each {
+            with(northbound.validateSwitch(it.dpId)) { validation ->
+                validation.verifyRuleSectionsAreEmpty(["missing", "misconfigured", "excess"])
+                validation.verifyMeterSectionsAreEmpty(["missing", "misconfigured", "excess"])
+            }
+        }
+
```
andriidovhan: Do you mind if we check that system allows to pass traffic through the flows?

rtretyak: yeah, had the same thoughts. I'll remove ping then and add traffic check. Because pings don't really leverage the ingress rule itself

PR #3089, opened 2019-12-31
Do not emit reroute commands for path, where edge switches are inactive

timofei-durakov:
patch adds extra check to validate flow terminating endpoints are
in active state before emiting re-route to flow h&s


services/wfm/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/reroute/actions/ValidateFlowAction.java
```diff
@@ -72,6 +72,14 @@ public ValidateFlowAction(PersistenceManager persistenceManager, FlowOperationsD
                 throw new FlowProcessingException(ErrorType.REQUEST_INVALID,
                         format("Flow %s is in progress now", flowId));
             }
+            if (!foundFlow.getSrcSwitch().isActive()) {
```
surabujin, line 75: This is not a reliable check. We do not have reliable data about switch state. I would like to fail on the "switch missing" response from FL instead of this check.

What problem we are solving here?

surabujin:
Commit/pr message are invalid - you are adding checks inside flowhs - i.e. it part of reroute request handling i.e. reroute request have been sent already.

PR #3091, opened 2020-01-02
Now verify cleanup per feature rather than per spec

rtretyak:


services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/extension/fixture/CleanupVerifierExtension.groovy
```diff
@@ -32,23 +33,25 @@ class CleanupVerifierExtension extends ContextAwareGlobalExtension {
...
+            it.addInterceptor(new IMethodInterceptor() {
+                @Override
+                void intercept(IMethodInvocation invocation) throws Throwable {
+                    invocation.proceed()
+                    log.info("Running cleanup verifier for '$invocation.feature.name'")
+                    assert northbound.getAllFlows().empty
+                    northbound.getAllSwitches().each {
+                        def validation = northbound.validateSwitch(it.switchId)
+                        validation.verifyRuleSectionsAreEmpty()
+                        validation.verifyMeterSectionsAreEmpty()
```
andriidovhan: I guess we should check that meter is supported by a switch (`.description.contains("OF_12")` )

rtretyak: `verifyMeterSectionsAreEmpty` inside verifies that the 'meters' section is present in the validation response. So it's already safe

PR #3092, opened 2020-01-03
Minor tweaks in tests according to default meters validation feature

rtretyak:


services/src/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/SwitchHelper.groovy
```diff
@@ -131,6 +132,20 @@ class SwitchHelper {
         }
     }
 
+
+    static List<Long> getDefaultMeters(Switch sw) {
+        def swProps = northbound.getSwitchProperties(sw.dpId)
+        if (swProps.multiTable) {
+            def result = [2, 3, 16, 18]
```
andriidovhan: how about extracting these numbers into global variables?

rtretyak:
Reworked the way we get expected meters. Now bound to exact rule cookie to have more meaning. Now take into account OF_12 switches as well as lldp toggle

PR #3093, opened 2020-01-03
Fix the mess with flow endpoints during the flow update

surabujin:
The mess happened because of the usage of inter-object reference stored inside objects. The caller code directly defines which objects must be used, so service code must not rely on relation stored inside objects. Or else... incorrect data will be used for flow segment speaker requests.
    
So always use provided "parts" of domain objects and avoid usage of inter-objects links/relations.

### DIFF
--- a/src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/service/SpeakerFlowSegmentRequestBuilder.java
+++ b/src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/service/SpeakerFlowSegmentRequestBuilder.java
@@ -121,7 +121,7 @@ public class SpeakerFlowSegmentRequestBuilder implements FlowCommandBuilder {
                     oppositePath != null ? oppositePath.getPathId() : null);
         }
 
-        List<FlowSegmentRequestFactory> requests = new ArrayList<>(makePathRequests(path, context, encapsulation,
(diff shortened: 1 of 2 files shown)


services/wfm/src/main/java/org/openkilda/wfm/share/service/SpeakerFlowSegmentRequestBuilder.java
```diff
@@ -103,22 +103,21 @@ public SpeakerFlowSegmentRequestBuilder(FlowResourcesManager resourcesManager) {
...
             requests.addAll(makePathRequests(
-                    oppositePath, context, encapsulation, doIngress, doTransit, doEgress));
+                    flow, oppositePath, context, encapsulation, doIngress, doTransit, doEgress));
         }
         return requests;
     }
 
     private List<FlowSegmentRequestFactory> makePathRequests(
-            @NonNull FlowPath path, CommandContext context, FlowTransitEncapsulation encapsulation,
+            Flow flow, FlowPath path, CommandContext context, FlowTransitEncapsulation encapsulation,
```
sergii-iakovenko: Why do we allow to build flow segment requests for "null" path?

surabujin: We don't. This is a private method, I have doubts regarding the usefulness of `@NonNull` annotations for private methods.

Issue #3087, opened 2019-12-30
h&s auto-reroute may build path on a failed ISL and mark flow as UP


rtretyak:
Under certain circumstances it is possible for h&s reroute to build flow on a path (and leave it there), which is already known to be broken. Flow will have UP status and pass all validations, only the ping/traffic won't be going through. As far as investigation goes: a first reroute should start, then a subsequent reroute (initiated by one more port down) happens while the first one is in progress and for some reason the last reroute gets suppressed/fail to actually happen.
Adding below some artifacts. Issue can be sometimes reproduced by running EnduranceV2Spec.
history: [path_on_failed_isl.txt](https://github.com/telstra/open-kilda/files/4009815/path_on_failed_isl.txt)
Some logs:
![image](https://user-images.githubusercontent.com/7931639/71582467-0830a480-2b13-11ea-9c21-0db71f1bad2d.png)

What happened to the last legal reroute that should have fixed the flow, but it failed to happen:
![image](https://user-images.githubusercontent.com/7931639/71582515-3d3cf700-2b13-11ea-82b9-3446c0a0e5e7.png)

Note, the very first reroute actually finished only after all the other reroutes failed:
![image](https://user-images.githubusercontent.com/7931639/71582549-652c5a80-2b13-11ea-8055-6937b2ea6864.png)

So the sequence was:
1. Start a reroute
2. After ~7 seconds more reroute requests come, they get throttled into a single request
3. We see that request does not get satisfied for some reason (?)
4. The very first reroute from the step 1 finishes its execution and marks the flow as UP, ignoring the step 2 requests.

surabujin:
There is a "lock" for flow on reroute operation, so when second reroute event come into rerouting handler, it checks this lock and does not perform "second" (and so on) reroute if it busy. It helps to not "break" flow's data persistence, but it can't handle double errors (like in this case).

surabujin:
Fix has been released https://github.com/telstra/open-kilda/releases/tag/v1.45.2

PR #3097, opened 2020-01-08
Fix double network failure handling

surabujin:
Add queue in front of reroute "handler" so if reroute for the flow is
going and new reroute request for this flow received (double network
failure case) second request will be postponed.

At the end of reroute request processing, "handler" check the presence of
postponed request and handle them.

Related to #3087

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/3097)

### DIFF
--- a/services/wfm/src/main/java/org/openkilda/wfm/topology/flowhs/service/FlowRerouteService.java
+++ b/services/wfm/src/main/java/org/openkilda/wfm/topology/flowhs/service/FlowRerouteService.java
@@ -28,6 +28,8 @@ import org.openkilda.wfm.topology.flowhs.fsm.reroute.FlowRerouteContext;
 import org.openkilda.wfm.topology.flowhs.fsm.reroute.FlowRerouteFsm;
 import org.openkilda.wfm.topology.flowhs.fsm.reroute.FlowRerouteFsm.Event;
 import org.openkilda.wfm.topology.flowhs.fsm.reroute.FlowRerouteFsm.State;
+import org.openkilda.wfm.topology.flowhs.model.FlowRerouteFact;
+import org.openkilda.wfm.topology.flowhs.utils.RerouteRetryManager;
 
 import com.google.common.annotations.VisibleForTesting;
 import lombok.extern.slf4j.Slf4j;
@@ -40,6 +42,8 @@ import java.util.Set;
 public class FlowRerouteService {
     @VisibleForTesting
     final Map<String, FlowRerouteFsm> fsms = new HashMap<>();
+
+    private final RerouteRetryManager retryManager = new RerouteRetryManager();
 
     private final FlowRerouteFsm.Factory fsmFactory;
     private final FsmExecutor<FlowRerouteFsm, State, Event, FlowRerouteContext> fsmExecutor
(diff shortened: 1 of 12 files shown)


services/wfm/src/main/java/org/openkilda/wfm/topology/flowhs/utils/RerouteRetryManager.java
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * Save entry into retry queue. Return true if it is only entry for this flowId (i.e. operation should not be
+     * postponed).
+     */
+    public boolean push(FlowRerouteFact entity) {
+        RerouteRetryQueue queue = queueByFlowId.computeIfAbsent(
+                entity.getFlowId(), ignore -> new RerouteRetryQueue());
+        queue.add(entity);
+        log.info("Size of flow reroute queue for {} is {}", entity.getFlowId(), queue.size());
+        return queue.size() == 1;
```
timofei-durakov: this is a very strange return result for a method call, tbh I'd implement size here istead

services/wfm/src/main/java/org/openkilda/wfm/topology/flowhs/service/FlowRerouteService.java
```diff
@@ -87,24 +101,15 @@ public void handleRequest(String key, CommandContext commandContext, String flow
             return;
         }
 
-        String eventKey = commandContext.getCorrelationId();
-        if (flowEventRepository.existsByTaskId(eventKey)) {
-            log.error("Attempt to reuse key {}, but there's a history record(s) for it.", eventKey);
-            return;
+        FlowRerouteFact reroute = new FlowRerouteFact(
+                key, commandContext, flowId, pathsToReroute, forceReroute, rerouteReason);
+        if (retryQueue.push(reroute)) {
```
timofei-durakov: check queue size rather than push return value that is tight to 1 size

services/wfm/src/main/java/org/openkilda/wfm/topology/flowhs/service/FlowRerouteService.java
```diff
@@ -151,12 +156,54 @@ public void handleTimeout(String key) {
...
+    }
+
+    private void performHousekeeping(FlowRerouteFsm fsm, String key) {
+        log.debug("FSM with key {} is finished with state {}", key, fsm.getCurrentState());
+        fsms.remove(key);
+
+        carrier.cancelTimeoutCallback(key);
+
+        FlowRerouteFact reroute = retryQueue.remove(fsm.getFlowId())
+                .orElseThrow(() -> new IllegalStateException(String.format(
```
timofei-durakov: will this exception be raised each time there are no pending reroutes in a queue?

services/wfm/src/main/java/org/openkilda/wfm/topology/flowhs/utils/RerouteRetryQueue.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /**
+     * Return first/active queue entry.
+     */
+    public Optional<FlowRerouteFact> get() {
+        FlowRerouteFact result = active;
+        if (result == null) {
+            result = pending;
+        }
```
rozdy: As far as I can see state with `active == null and pending != null` is inconsistent for this structure. Maybe we should return only active Fact here for consistency with remove method?

services/wfm/src/main/java/org/openkilda/wfm/topology/flowhs/utils/RerouteRetryManager.java
niksv: +1

services/wfm/src/test/java/org/openkilda/wfm/topology/flowhs/utils/RerouteRetryQueueTest.java
```diff
@@ -0,0 +1,167 @@
...
+        Assert.assertEquals(reroutePathB.getFlowId(), reroute.getFlowId());
+
+        Set<PathId> expectedPaths = Stream.of(rerouteEmpty, reroutePathA, rerouteForced, reroutePathB)
+                .map(FlowRerouteFact::getPathsToReroute)
+                .filter(Objects::nonNull)
+                .flatMap(Collection::stream)
+                .collect(Collectors.toSet());
+        Assert.assertEquals(expectedPaths, reroute.getPathsToReroute());
+        Assert.assertTrue(reroute.isForceReroute());
+        Assert.assertEquals(reroutePathB.getRerouteReason(), reroute.getRerouteReason());
```
niksv: all 4 rerouteFact objects has same reason "reason 1". So check if we set right reason you need to have different reasons

services/wfm/src/main/java/org/openkilda/wfm/topology/flowhs/utils/RerouteRetryQueue.java
niksv: or throw exception if `active == null and pending != null`

PR #3116, opened 2020-01-15
minor improvements in featureToogle spec

andriidovhan:


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/toggles/FeatureTogglesSpec.groovy
```diff
@@ -20,11 +20,9 @@ creation of new flows via Northbound API. This spec verifies that Feature Toggle
 /*Note that the 'flowReroute' toggle is tested under AutoRerouteSpec#"Flow goes to 'Down' status when an intermediate
 switch is disconnected and there is no ability to reroute".
 BFD toggle is tested in BfdSpec*/
-@Tags([SMOKE, LOW_PRIORITY])
+//TODO (andriidovhan) remove VIRTUAL tag when mapping is disabled
```
rtretyak: why?

andriidovhan: in can't be tested on the stage env while mapping is enabled
An error is different for APIv1 and APIv2

PR #3133, opened 2020-01-21
Reroute affected flows on switch up event

### LINKED ISSUE #3131: System may be missing a successful reroute possibility on switch UP event
Steps:
1. Flow breaks and tries to reroute
2. System finds a single possible failover path and tries to install rules on that path
3. One of the switches on the failover path goes offline, not letting reroute to finish. Note that the switch-related ISLs still remain UP at this stage.
4. Flow goes to DOWN state, since it wasn't able to install rules on the path and there's no more other paths left. ISLs are still UP
5. Switch goes back online
**Expected result:** We must try to reroute the switch-related DOWN flow one more time, since now the failover path is available, even though there was no ISL UP event (which is a reroute trigger)
**Actual result:** We are not trying to reroute the flow, since currently the reroute trigger is ISL UP, not Switch UP

rozdy:
Close #3131


services/src/messaging/src/main/java/org/openkilda/messaging/command/reroute/RerouteAffectedFlowsRequest.java
```diff
@@ -0,0 +1,42 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.NonNull;
+
+@Data
+@EqualsAndHashCode(callSuper = false)
+@JsonNaming(SnakeCaseStrategy.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class RerouteAffectedFlowsRequest extends CommandData {
```
timofei-durakov: There is RerouteAffectedFlows class in the same package, why not to extend it instead?

services/wfm/src/main/java/org/openkilda/wfm/topology/reroute/service/RerouteService.java
```diff
@@ -271,6 +296,24 @@ private boolean filterForwardPrimaryPath(FlowPath path) {
                 );
     }
 
+    /**
+     * Returns map with affected inactive flow and flow pathId set for rerouting.
+     */
+    public Map<Flow, Set<PathId>> getAffectedInactiveFlowsForRerouting(SwitchId switchId) {
+        log.info("Get affected inactive flows for switch {}", switchId);
+        return flowRepository.findDownFlows().stream()
+                .filter(f -> f.getPaths().stream()
```
timofei-durakov: I would prefer to extend repository with required method and not fetching all flows here.
P.S. it's fine to filter out flows by segments since we do not interested in single switch flows here(kilda do not attempt to re-route them).

services/src/messaging/src/main/java/org/openkilda/messaging/command/reroute/RerouteAffectedFlowsRequest.java
```diff
@@ -0,0 +1,42 @@
+/* Copyright 2018 Telstra Open Source
```
timofei-durakov: 2020:-P

services/wfm/src/main/java/org/openkilda/wfm/topology/reroute/service/RerouteService.java
timofei-durakov: Also there are no unit-tests

services/src/messaging/src/main/java/org/openkilda/messaging/command/reroute/RerouteAffectedFlowsRequest.java
rozdy: Renamed class. This two requests should trigger different logic so I guess better to keep them separately.

services/wfm/src/main/java/org/openkilda/wfm/topology/reroute/service/RerouteService.java
rozdy: Reworked and added tests.

services/src/messaging/src/main/java/org/openkilda/messaging/command/reroute/RerouteAffectedFlowsRequest.java
rozdy: Fixed :)

services/src/messaging/src/main/java/org/openkilda/messaging/command/reroute/RerouteInactiveAffectedFlows.java
```diff
@@ -0,0 +1,42 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.NonNull;
+
+@Data
+@EqualsAndHashCode(callSuper = false)
+@JsonNaming(SnakeCaseStrategy.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class RerouteInactiveAffectedFlows extends CommandData {
```
niksv: I guess it should be RerouteAffectedInactiveFlows

services/wfm/src/main/java/org/openkilda/wfm/topology/reroute/service/RerouteService.java
```diff
@@ -263,14 +289,25 @@ private boolean filterForwardPrimaryPath(FlowPath path) {
...
     }
 
+    /**
+     * Returns affected inactive flow set for rerouting.
+     */
+    public Set<Flow> getAffectedInactiveFlowsForRerouting(SwitchId switchId) {
+        log.info("Get affected inactive flows for switch {}", switchId);
+        return flowPathRepository.findBySegmentSwitch(switchId).stream()
+                .map(FlowPath::getFlow)
+                .filter(f -> f.getStatus() == FlowStatus.DOWN || f.getStatus() == FlowStatus.DEGRADED)
```
niksv: it can be better. You can optimize it by getting only Flow Paths of Down and Degraded flows by Neo4j.

To do it you need to add new method into flowPathRepository. Something like

```

    @Override
    public Collection<FlowPath> findInactiveBySegmentSwitch(SwitchId switchId) {
        Map<String, Object> parameters = ImmutableMap.of(
                "switch_id", switchIdConverter.toGraphProperty(switchId));

        Set<String> pathIds = new HashSet<>();
        queryForStrings("MATCH (ps_src:switch)-[:source]-(ps:path_segment)-[:destination]-(ps_dst:switch)" +
                "WHERE ps_src.name = $switch_id OR ps_dst.name = $switch_id" +
                "MATCH (f:flow)-[:owns]-(fp:flow_path)-[:owns]-(ps)" +
                "WHERE f.status = 'down' or f.status = 'degraded'" +
                "RETURN fp.path_id as path_id", parameters, "path_id").forEach(pathIds::add);

        if (pathIds.isEmpty()) {
            return emptyList();
        }

        Filter pathIdsFilter = new Filter(PATH_ID_PROPERTY_NAME, new InOperatorWithNoConverterComparison(pathIds));

        return loadAll(pathIdsFilter);
    }
```

services/src/messaging/src/main/java/org/openkilda/messaging/command/reroute/RerouteInactiveAffectedFlows.java
rozdy: Fixed

services/wfm/src/main/java/org/openkilda/wfm/topology/reroute/service/RerouteService.java
rozdy: Reworked

services/src/kilda-core/kilda-persistence-neo4j/src/main/java/org/openkilda/persistence/repositories/impl/Neo4jFlowPathRepository.java
```diff
@@ -206,6 +206,27 @@ public Neo4jFlowPathRepository(Neo4jSessionFactory sessionFactory, TransactionMa
         return loadAll(pathIdsFilter);
     }
 
+    @Override
+    public Collection<FlowPath> findInactiveBySegmentSwitch(SwitchId switchId) {
```
dpoltavets: This method is not used anywhere.

rozdy: Fixed

docs/design/network-discovery/switch-FSM.puml
```diff
@@ -60,6 +60,7 @@ ONLINE --> OFFLINE : offline
 
 OFFLINE : enter / update status in DB
 OFFLINE : enter / emit offline for portAll()
+OFFLINE : online / send reroute affected flows
```
nikitamarchenko: Why you did that on sync, I prefer on Online state.

docs/design/network-discovery/switch-FSM.puml
```diff
@@ -28,6 +28,8 @@ endlegend
 
 [*] --> INIT
 
+INIT : set wasOffline = false
```
nikitamarchenko: You already have that flag in DB

rozdy:
> _[docs/design/network-discovery/switch-FSM.puml, line 31 at r2](https://reviewable.io/reviews/telstra/open-kilda/3133#-LzbSFAj3PeySE7e6vQq:-LzbSFAj3PeySE7e6vQr:b-6gkfay) ([raw file](https://github.com/telstra/open-kilda/blob/908567380b0ac3d41a31c4f78b90d6f962fe3df2/docs/design/network-discovery/switch-FSM.puml#L31)):_
> 
> > ```
> > [*] --> INIT
> > 
> > INIT : set wasOffline = false
> > ```
> 
> Why we need this flag?
> 

This flag will help us to detect flapping switches and prevent massive flow reroutes on topology startup.


> > ```java
> >                 + "WHERE ps_src.name = $switch_id OR ps_dst.name = $switch_id "
> >                 + "MATCH (f:flow)-[:owns]-(fp:flow_path)-[:owns]-(ps) "
> >                 + "WHERE f.status = 'down' OR f.status = 'degraded' "
> > ```
> 
> Inject these constants via parameters. In current shape, it will be invisible for refactoring tools provided by IDE.
> 
Fixed.

> > ```java
> >         topology.setBolt(RerouteEncoder.BOLT_ID, bolt, scaleFactor)
> >                 .shuffleGrouping(IslHandler.BOLT_ID, IslHandler.STREAM_REROUTE_ID)
> >                 .shuffleGrouping(SwitchHandler.BOLT_ID, IslHandler.STREAM_REROUTE_ID);
> > ```
> 
> Why do you subscribe on `SwitchHandler` steam via `IslHandler` stream name?
Fixed.

> 

> > ```java
> >     private SpeakerSwitchView speakerData;
> >     private boolean wasOffline;
> > ```
> 
> The explicit initial value is better than implicit default.
Fixed.

> 
> > ```java
> > package org.openkilda.wfm.topology.network.storm.bolt.sw;
> > 
> > import static org.openkilda.wfm.topology.network.storm.bolt.isl.IslHandler.STREAM_REROUTE_FIELDS;
> > ```
> 
> ????????????????????????
Fixed.


> > ```java
> >     }
> > 
> >     private Values makeAffectedFlowRerouteTuple(SwitchId switchId) {
> > ```
> 
> These "make*" methods purpose is to make correct tuple for the specified stream. I.e this method must be named `makeRerouteTuple` and must produce correct tuple for reroute stream.
Fixed.

> > ```java
> >         for (Flow flow : flowsForRerouting) {
> >             if (flow.isPinned()) {
> > ```
> 
> Why flow status is "filtered" inside DB query while it's "pinned" status is filtered here?
For consistency with rerouteInactiveFlows method. I think log message for skipped pinned flows is required.


> > ```
> > Collections.emptySet()
> > ```
> 
> Add a comment describing why we must not fill the affected endpoints field for this reroute request.
> 
> Also need to change the logic that merges affected reroute requests, it must wipe existing affected endpoints set on requests with empty affected endpoint set. I.e. if existing or just received reroute request have empty affected endpoint set - resulting affected endpoint set must stay empty to force flow to reroute despite it's current paths.
Fixed.

rozdy:
@surabujin 
> > ```groovy
> >         //and: "Flow is rerouted due to switchUp event"
> >         //TODO(andriidovhan) check flow history (it is not implemented yet)
> > ```
> 
> Is it going to be implemented for this release or it will be dome sometime in future?

I will fix this later in separate PR.

> Still not enough details. The main piece of info - during failed reroute(that fails due to some switch inaccessibility in new path) flow will be rolled back to it's current path, so we can't map switch to any "down" flow or path, because this switch is part of potential path, not path stored in flow at this moment. So we must force reroute attempt for all down flow/paths.

Correct. This additional reroute will be triggered for terminating switches fails.

PR #3138, opened 2020-01-22
Introduce PoP for the switch and take it into account in pce

timofei-durakov:
protected paths should take into account PoP and try not to
place protected path into the same transit PoPs

### DIFF
--- a/services/src/kilda-pce/src/main/java/org/openkilda/pce/impl/AvailableNetwork.java
+++ b/services/src/kilda-pce/src/main/java/org/openkilda/pce/impl/AvailableNetwork.java
@@ -28,6 +28,7 @@ import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -42,6 +43,9 @@ import java.util.Set;
 public class AvailableNetwork {
     @VisibleForTesting
     final Map<SwitchId, Node> switches = new HashMap<>();
+
+    @VisibleForTesting
+    final Set<Edge> edges = new HashSet<>();
 
     public Node getSwitch(SwitchId dpid) {
         return switches.get(dpid);
@@ -68,6 +72,7 @@ public class AvailableNetwork {
                 .srcSwitch(srcSwitch)
                 .destSwitch(dstSwitch)
                 .build();
+        edges.add(edge);
         boolean srcAdded = srcSwitch.getOutgoingLinks().add(edge);
         boolean dstAdded = dstSwitch.getIncomingLinks().add(edge);
         if (errorOnDuplicates && !(srcAdded && dstAdded)) {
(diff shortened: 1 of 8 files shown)


services/src/kilda-pce/src/main/java/org/openkilda/pce/impl/AvailableNetwork.java
```diff
@@ -113,6 +121,43 @@ public void processDiversitySegments(List<PathSegment> segments) {
...
+            }
+            String dstPop = edge.getDestSwitch().getPop();
+            if (dstPop != null && allocatedPopSet.contains(dstPop)) {
+                edge.increaseDiversityGroupUseCounter();
+                continue;
+            }
+        }
+
+
+    }
```
rozdy: I think we need some combination of this method and old processDiversitySegments without pops method. Should we try to create diverse path through already used pops if no "unused" pops are available? Maybe add separate PopDiversityGroupUseCounter and related diversity.pop.cost param?

timofei-durakov: I think it makes sense, and overall we should use both strategies at the same time. I introduced new param for config and add new counter for edge, it will also mitigate an issues when there are no PoPs in db yet. good catch

services/src/kilda-pce/src/main/java/org/openkilda/pce/impl/AvailableNetwork.java
```diff
@@ -113,6 +118,43 @@ public void processDiversitySegments(List<PathSegment> segments) {
...
 
+    /**
+     * Adds diversity weights into {@link AvailableNetwork} based on passed path segments and configuration.
+     */
+    public void processDiversitySegmentsWithPop(List<PathSegment> segments) {
+        if (segments.size() <= 1) {
+            return;
+        }
+
+        Set<String> allocatedPopSet = new HashSet<>();
```
rozdy: nit: allocatedPopSet -> allocatedPops or transitPops

services/src/kilda-pce/src/main/java/org/openkilda/pce/impl/AvailableNetwork.java
```diff
@@ -113,6 +118,43 @@ public void processDiversitySegments(List<PathSegment> segments) {
...
+        for (Edge edge : edges) {
+            String srcPop = edge.getSrcSwitch().getPop();
+            if (srcPop != null && allocatedPopSet.contains(srcPop)) {
+                edge.increaseDiversityGroupPerPopUseCounter();
+                continue;
+            }
+            String dstPop = edge.getDestSwitch().getPop();
+            if (dstPop != null && allocatedPopSet.contains(dstPop)) {
+                edge.increaseDiversityGroupPerPopUseCounter();
+                continue;
```
rozdy: nit: this continue is redundant.

services/src/kilda-pce/src/main/java/org/openkilda/pce/impl/AvailableNetwork.java
timofei-durakov: fixed as we discussed offline

services/src/kilda-pce/src/main/java/org/openkilda/pce/impl/AvailableNetwork.java
timofei-durakov: it was added intentionally so in case of upcoming changes it won't be forgotten

PR #3139, opened 2020-01-22
Add test for breaking protected path during swap

rtretyak:
- Update topology.yaml to enable 4 diverse paths situation

Reveals #3140


services/src/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/ProtectedPathV2Spec.groovy
```diff
@@ -1122,6 +1125,60 @@ class ProtectedPathV2Spec extends HealthCheckSpecification {
         flowHelperV2.deleteFlow(flow.flowId)
     }
 
+    def "System properly reroutes both paths if protected path breaks during swap"() {
```
andriidovhan: Tidy ?

PR #3147, opened 2020-01-27
Periodic pings fix

timofei-durakov:


timofei-durakov:
@surabujin FlowOpertations bolt don't use that. Moved the method def to FlowGenericCarrier. P.S. consider reviewing #3148   instead of this one

PR #3156, opened 2020-01-31
Dump all switches to return visible switches

timofei-durakov:
switches should be available for dump only after
a handshake

### DIFF
--- a/services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
+++ b/services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
@@ -978,8 +978,11 @@ public class SwitchManager implements IFloodlightModule, IFloodlightService, ISw
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


services/src/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java
```diff
@@ -978,8 +978,11 @@ public void modifyMeterForFlow(DatapathId dpid, long meterId, long bandwidth)
     }
 
     @Override
-    public Map<DatapathId, IOFSwitch> getAllSwitchMap() {
-        return ofSwitchService.getAllSwitchMap();
+    public Map<DatapathId, IOFSwitch> getAllSwitchMap(boolean visible) {
```
surabujin: Do we need anywhere "not ready to be used switches"? I mean when we need to set/use "false" for visible argument?

timofei-durakov: underlying fl logic uses it however doesn't expose, I decided to provide the flag as well.
Use case: don't know, might be something in future

PR #3174, opened 2020-02-05
Correct handle empty affected isl set in reroute throttling

rozdy:


services/wfm/src/main/java/org/openkilda/wfm/topology/reroute/service/ReroutesThrottling.java
```diff
@@ -57,12 +57,16 @@ public ReroutesThrottling(long minDelay, long maxDelay, int defaultFlowPriority)
      * @param throttlingData the correlation ID and flow priority.
      */
     public void putRequest(String flowId, FlowThrottlingData throttlingData) {
-        log.info("Puts flow {} with correlationId {}", flowId, throttlingData.getCorrelationId());
+        log.info("Puts reroute request for flow {} with correlationId {}", flowId, throttlingData.getCorrelationId());
         FlowThrottlingData prevThrottlingData = reroutes.put(flowId, throttlingData);
         if (prevThrottlingData != null) {
-            throttlingData.getAffectedIsl().addAll(prevThrottlingData.getAffectedIsl());
+            if (prevThrottlingData.getAffectedIsl().isEmpty() || throttlingData.getAffectedIsl().isEmpty()) {
```
sergii-iakovenko: I think there's no sense to reset "throttlingData.affectedIsl" if it's empty already. So "throttlingData.getAffectedIsl().isEmpty()" must go to the "else if" block.

rozdy: Fixed

PR #3179, opened 2020-02-07
Packet out prob fl rest endpoint

timofei-durakov:


nikitamarchenko (approved):

src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/prob/ProbService.java
```diff
@@ -0,0 +1,128 @@
...
+
+
+
+        UDP l4 = new UDP();
+        l4.setSourcePort(TransportPort.of(11));
+        l4.setDestinationPort(TransportPort.of(udpDst));
+
+
+
+        Ethernet l2 = new Ethernet().setSourceMACAddress(MacAddress.of("55:44:33:22:11:00"))
```
sergii-iakovenko, line 96: Why not move to the FL configuration?

PR #3191, opened 2020-02-11
Expose OF transaction id into speaker logs

surabujin:
Extend speaker log message by addind OF transaction ID (xid) into all
services that produce/parse OF messages. It should help to correlate
open-kilda logs with switch logs.

### DIFF
--- a/src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/service/session/Session.java
+++ b/src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/service/session/Session.java
@@ -189,7 +189,7 @@ public class Session implements AutoCloseable {
 
     private void actualWrite(OFMessage message)
             throws SwitchWriteException {
-        log.debug("push OF message to {}: {}", sw.getId(), message);
+        log.info("push OF message to {}: {}", sw.getId(), message);
         if (!sw.write(message)) {
             error = true;
             throw new SwitchWriteException(sw.getId(), message);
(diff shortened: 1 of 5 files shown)


src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/service/session/Session.java
```diff
@@ -189,7 +189,7 @@ boolean handleResponse(OFMessage message) {
 
     private void actualWrite(OFMessage message)
             throws SwitchWriteException {
-        log.debug("push OF message to {}: {}", sw.getId(), message);
+        log.info("push OF message to {}: {}", sw.getId(), message);
```
pzakatov: Doubt about this change. Will it repeat a log record for specific messages that already logged?

surabujin: It is not used to send packet-out for discovery requests (but used to send flow-ping). And we have the same log message of "old" way to send OF messages (https://github.com/telstra/open-kilda/blob/develop/src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchManager.java#L2680).

I propose to give it a try and collect some usage experience. After that, we can decide how to improve it.

pzakatov: ok

src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/statistics/StatisticsService.java
```diff
@@ -168,11 +168,10 @@ private void gatherPortStats(IOFSwitch iofSwitch) {
                 .setPortNo(OFPort.ANY)
                 .build();
 
-        logger.trace("Getting port stats for switch={}", iofSwitch.getId());
+        logger.info("Getting port stats for switch={} OF-xid:{}", iofSwitch.getId(), portStatsRequest.getXid());
```
timofei-durakov: could you please confirm that this xid is  already set, the same question is valid for all requests from this StatisticsService

PR #3194, opened 2020-02-11
Add auto reroute reason to flow history

rozdy:


src-java/reroute-topology/reroute-storm-topology/src/main/java/org/openkilda/wfm/topology/reroute/model/FlowThrottlingData.java
```diff
@@ -32,20 +32,22 @@
     private Integer priority;
     private Instant timeCreate;
     private Set<IslEndpoint> affectedIsl;
+    private String reason;
 
     @VisibleForTesting
     public FlowThrottlingData(String correlationId, Integer priority) {
-        this(correlationId, priority, Instant.now(), Collections.emptySet());
+        this(correlationId, priority, Instant.now(), Collections.emptySet(), "");
```
sergii-iakovenko: Why is the empty string used here, not null? Asking as the "reason" field is nullable, so it's possible to set null.

rozdy: This constructor used for tests only. Changed to null.

src-java/reroute-topology/reroute-storm-topology/src/main/java/org/openkilda/wfm/topology/reroute/model/FlowThrottlingData.java
```diff
@@ -32,20 +32,22 @@
     private Integer priority;
     private Instant timeCreate;
     private Set<IslEndpoint> affectedIsl;
+    private String reason;
 
     @VisibleForTesting
     public FlowThrottlingData(String correlationId, Integer priority) {
-        this(correlationId, priority, Instant.now(), Collections.emptySet());
+        this(correlationId, priority, Instant.now(), Collections.emptySet(), null);
```
timofei-durakov: does it make sense to remove this constructor at all?

rozdy: Reworked code to remove this constructor.

rozdy:
Auto reroute reason is fixed by #3295

PR #3215, opened 2020-02-17
Allow flowhs speaker request retry for all error codes

surabujin:
Flow H&S create operation use separate set of tools to perform speaker
request retries. It allow to retry only for SWITCH_UNAVAILABE error.
Other CRUD operation do retry on any error code. Rework create operation
to mimic speaker retry logic of all other CRUD operations.

### DIFF
--- a/src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/SpeakerCommandFsm.java
+++ b/src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/SpeakerCommandFsm.java
@@ -15,11 +15,10 @@
 
 package org.openkilda.wfm.topology.flowhs.fsm.common;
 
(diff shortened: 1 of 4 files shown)


src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/SpeakerCommandFsm.java
```diff
@@ -30,20 +29,28 @@
...
 
-    private SpeakerCommandFsm(FlowSegmentRequest request, FlowCreateHubCarrier carrier, Integer retriesLimit) {
+    private SpeakerCommandFsm(
+            FlowSegmentRequest request, FlowCreateHubCarrier carrier, Set<ErrorCode> giveUpErrors,
+            Integer retriesLimit) {
         super(new CommandContext(request.getMessageContext().getCorrelationId()));
         this.request = request;
         this.carrier = carrier;
-        this.remainingRetries = retriesLimit;
+        this.giveUpErrors.addAll(giveUpErrors);
```
sergii-iakovenko, line 52: nit: It's the only constructor, why not just assign the field?

PR #3217, opened 2020-02-18
Added ARP support to traff gens

niksv:


src-python/lab-service/traffexam/kilda/traffexam/action.py
```diff
@@ -20,6 +20,7 @@
 import nsenter
 from scapy.all import Ether, sendp
 from scapy.contrib import lldp
+from scapy.layers.l2 import ARP
```
dpoltavets: I think it would be better to either take `ARP` from the `.all` package as `Ether`, or explicitly specify `.layers.l2` for `Ether` and `sendp`.

niksv: fixed

Issue #2701, opened 2019-08-14
Create flow v2 api is not properly setting `diverse_with` in response


rtretyak:
1. Create a flow which is diversed with some other flow(s). Check response
**Expected:** response has `diverse_with` field with the list of flow ids that are diversed with the created flow
**Actual:** `diverse_with` is set to `null`

PR #3224, opened 2020-02-20
Fixed display of `diverse_with` field in response via APIv2.

dpoltavets:
Closes: #2701

### DIFF
--- a/src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/model/FlowDto.java
+++ b/src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/model/FlowDto.java
@@ -31,6 +31,7 @@ import lombok.Data;
 import lombok.EqualsAndHashCode;
 
 import java.io.Serializable;
+import java.util.Set;
 
 @Data
 @EqualsAndHashCode(exclude = {"ignoreBandwidth", "periodicPings", "cookie", "createdTime", "lastUpdated", "meterId",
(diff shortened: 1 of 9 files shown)


src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/model/FlowDto.java
```diff
@@ -164,6 +164,9 @@
     @JsonProperty("path_computation_strategy")
     private PathComputationStrategy pathComputationStrategy;
 
+    @JsonProperty("diverse_with")
+    private String[] diverseWith;
```
timofei-durakov: why is it array, not list?

dpoltavets: Changed to Set.

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/FlowProcessingAction.java
```diff
@@ -101,4 +101,11 @@ protected boolean isRemoveCustomerPortSharedCatchRule(String flowId,
 
         return flowIds.size() == 1 && flowIds.iterator().next().equals(flowId);
     }
+
+    protected Set<String> getDiverseWithFlowIds(Flow flow) {
+        return flowRepository.findByGroupId(flow.getGroupId()).stream()
```
sergii-iakovenko: Does it make sense to fetch a collection of Flow entities and then return just flow ids? Please, consider a new method in the repository.

dpoltavets: Fixed.

PR #3226, opened 2020-02-20
Add func tests for ARP connected devices

rtretyak:


src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/SwitchHelper.groovy
```diff
@@ -84,49 +105,53 @@ class SwitchHelper {
         if (swProps.multiTable) {
             multiTableRules = [MULTITABLE_PRE_INGRESS_PASS_THROUGH_COOKIE, MULTITABLE_INGRESS_DROP_COOKIE,
                     MULTITABLE_POST_INGRESS_DROP_COOKIE, MULTITABLE_EGRESS_PASS_THROUGH_COOKIE,
-                    MULTITABLE_TRANSIT_DROP_COOKIE, LLDP_POST_INGRESS_COOKIE, LLDP_POST_INGRESS_ONE_SWITCH_COOKIE]
+                    MULTITABLE_TRANSIT_DROP_COOKIE, LLDP_POST_INGRESS_COOKIE, LLDP_POST_INGRESS_ONE_SWITCH_COOKIE,
+                    ARP_POST_INGRESS_COOKIE, ARP_POST_INGRESS_ONE_SWITCH_COOKIE]
             if (sw.features.contains(SwitchFeature.NOVIFLOW_COPY_FIELD)) {
```
niksv: According to code here must be `NOVIFLOW_PUSH_POP_VXLAN` feature

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/SwitchHelper.groovy
```diff
@@ -164,13 +189,13 @@ class SwitchHelper {
...
-                actualHexCookie.add(Cookie.decode(cookie).toString())
+                actualHexCookie.add(decode(cookie).toString())
             }
             def expectedHexCookie = []
             for (long cookie : sw.defaultCookies) {
-                expectedHexCookie.add(Cookie.decode(cookie).toString())
+                expectedHexCookie.add(decode(cookie).toString())
             }
-            assert actualHexCookie.sort() == expectedHexCookie.sort()
+            assert (actualHexCookie - expectedHexCookie).empty
```
niksv: what will happen if some rules are missed on the switch? 
For example 
`actualHexCookie` = [1,2]
`expectedHexCookie` = [1,2,3]

will condition `(actualHexCookie - expectedHexCookie).empty` be false positive in this case

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/SwitchHelper.groovy
rtretyak: Yeah, I know. Fixed it in https://github.com/telstra/open-kilda/pull/3092/files but not here. Will fix it here too

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/SwitchHelper.groovy
rtretyak: very true, thx. Will fix this

rtretyak: fixed

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/SwitchHelper.groovy
```diff
@@ -84,49 +106,53 @@ class SwitchHelper {
...
+                        switchLldpRules.add(encodeLldpInputCustomer(it.destination.portId))
                     }
                 }
             }
         }
         if (swProps.switchLldp) {
             switchLldpRules.addAll([LLDP_INPUT_PRE_DROP_COOKIE, LLDP_TRANSIT_COOKIE, LLDP_INGRESS_COOKIE])
         }
+        if (swProps.switchArp) {
+            switchLldpRules.addAll([ARP_INPUT_PRE_DROP_COOKIE, ARP_TRANSIT_COOKIE, ARP_INGRESS_COOKIE])
```
niksv: nit: not quite accurate name `switchLldpRules` because we add ARP rules to this list

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/SwitchHelper.groovy
```diff
@@ -164,13 +190,13 @@ class SwitchHelper {
...
-                actualHexCookie.add(Cookie.decode(cookie).toString())
+                actualHexCookie.add(decode(cookie).toString())
             }
             def expectedHexCookie = []
             for (long cookie : sw.defaultCookies) {
-                expectedHexCookie.add(Cookie.decode(cookie).toString())
+                expectedHexCookie.add(decode(cookie).toString())
             }
-            assert actualHexCookie.sort() == expectedHexCookie.sort()
+            assertThat actualHexCookie, containsInAnyOrder(expectedHexCookie.toArray())
```
niksv: Will be false positive in next case:
1. switch is in multiTable mode
2. change switch mode to single table
3. For some reasons all rules were **NOT** reinstalled.

**Expected result:**
Assertion failed

**Actual result**
Assert is OK because:

Single table rules are a subset of multi table rules (see line 143)

so in our case actualHexCookie = [singleTableRules + excess multi table rules]
expectedHexCookies = [singleTableRules]

that is why `actualHexCookie. containsInAnyOrder(expectedHexCookie.toArray())` is OK

rtretyak: Even though the matcher is called 'contains..', it does a full match and will complain in both cases if actual is bigger than expected or expected is bigger than actual.

niksv: okay

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/DefaultRulesValidationSpec.groovy
```diff
@@ -41,7 +55,50 @@ class DefaultRulesValidationSpec extends HealthCheckSpecification {
...
+                    ]
+                ], getTopology().getActiveSwitches().unique { activeSw -> activeSw.description }
+                ].combinations().findAll { Map swProp, Switch _sw ->
+                    //filter out combinations where we pick non-multitable switch and do multitable switch props
+                    !(swProp.multiTable && !_sw.features.contains(SwitchFeature.MULTI_TABLE))
+                }
+        propsDescr = getDescr(swProps.multiTable, swProps.switchLldp, swProps.switchArp)
+    }
+
+    String getDescr(boolean multitable, boolean lldp, boolean arp) {
```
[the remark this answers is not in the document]
rtretyak: Any arguments?

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/SwitchHelper.groovy
rtretyak: fixed

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/DefaultRulesValidationSpec.groovy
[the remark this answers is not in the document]
andriidovhan: All passed parameters can be taken the same object. 
So, I strongly believe it is easier to pass an object. 
In addition you don't need to remember the correct order of passed arguments.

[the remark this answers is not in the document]
rtretyak: My thought process is: Your option is 'not safe'. You have to 'remember' the key names and do not get any suggestions from the IDE. It is not type-safe as well. I prefer not to spread dynamic code across the place and get rid of it as soon as possible, since any typo will only be caught during runtime.
I understand that we pass the same map to the test itself, but I would happily use custom classes(like `ConnectedDeviceTestData` in ConnectedDevicesSpec) if it didn't require additional coding. And I would be happy to pass a strongly-typed object here as a param. I would also may have considered using map if we had too many params to pass(but its only 3 here).
So you gave only 1 argument: ' you don't need remember the order of passed arguments'. Nobody have to 'remember' it, it is the method signature, order will always be suggested by the IDE..unlike remembering keys of your map.
Don't get me wrong, I see your suggestion as an option and don't think its bad, but unless you give more arguments I don't see real reasons to change current code. It's just different

niksv:
@rtretyak Do we need to add tests for v2 Flow CRUD with connected devices to check that LLDP/ARP shared rules will be installed/deleted correctly?

PR #3234, opened 2020-02-25
Make parallelism level configurable

surabujin:
Fill parallelism related properties from template variables. It allows
to change parallelism setittings without rebuiulding deployed images.


nikitamarchenko:
We have issues related to tuple marshaling between more than one worker per process. So don't use that on production.

PR #3239, opened 2020-02-25
Avoid reroute fsm removal on request collision

surabujin:
Fix housekeeping method to not remove FSM on reroute request collision
(same correlation ID).


surabujin:
Became obsolete due to #3295

PR #3247, opened 2020-02-28
Fix and prevent dependency classes conflict in src-java

sergii-iakovenko:


sergii-iakovenko:
> _[src-java/floodlight-service/floodlight-api/src/main/java/org/openkilda/messaging/info/event/IslChangeType.java, line 30 at r1](https://reviewable.io/reviews/telstra/open-kilda/3247#-M4-7R3q2HWB8TcGdT9B:-M4-7R3q2HWB8TcGdT9C:b-leeag2) ([raw file](https://github.com/telstra/open-kilda/blob/62168a91240e02ff882021520662bfb956c354c2/src-java/floodlight-service/floodlight-api/src/main/java/org/openkilda/messaging/info/event/IslChangeType.java#L30)):_
> 
> > ```java
> > @JsonSerialize
> > @JsonInclude(JsonInclude.Include.NON_NULL)
> > public enum IslChangeType {
> > ```
> 
> Is it trigger class conflict? Perhaps it should be removed in separate PR and this one should be dependent on it? Why did you decide to remove this one (because we have a goal to remove/split messaging module, we should remove it's chunks when possible)?

It happened that IslChangeType has a duplicate in floodlight-api and base-messaging modules. The script detects it and fails a build, so the duplicate should be removed.

PR #3263, opened 2020-03-04
Ignore discovery poll events in BFD mode

surabujin:
Altering the logic of ISL liveness detection - do not react on
poll-discovery events in BFD-down state i.e. when the ISL is reported as
down by notification from BFD session. Now ISL can flap (DOWN -> UP(up
to discovery timeout) -> DOWN) right after BFD-down event due to
received stale(it was sent before actual LINK fail and received after
BFD have reported about LINK fail) discovery-poll event.


surabujin:
Superseded by #3407

PR #3274, opened 2020-03-10
Added default value for detect connected devices in V2 API

niksv:


src-java/northbound-service/northbound-api/src/main/java/org/openkilda/northbound/dto/v2/flows/FlowEndpointV2.java
```diff
@@ -33,5 +34,6 @@
     private Integer portNumber;
     private int vlanId;
     @NonNull
-    private DetectConnectedDevicesV2 detectConnectedDevices;
+    @Default
+    private DetectConnectedDevicesV2 detectConnectedDevices = new DetectConnectedDevicesV2(false, false);
```
surabujin: Can we have no-args constructor in `DetectConnectedDevicesV2`?

niksv: I think yes. It will be good to have it

PR #3276, opened 2020-03-10
Clean up compile time warnings for kilda

timofei-durakov:
- removed useless NonNull on primitives
- updated EqualsAndHashCode to explicitly mark not calling super
- supressed reporting from mapstruct

---
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/3276)


timofei-durakov:
@surabujin ok, I fixed implicit warnings for all mappers with explicit ignore field annotations. From the call super perspective: I  quite agree that we should call super, but current way it works - without super. To mitigate possible risks I've created a separate pr to make this change separately https://github.com/telstra/open-kilda/pull/3283

PR #3288, opened 2020-03-12
Add db indexes for flow history objects

surabujin:
Note: Adds DB migration.


surabujin:
PR was reverted because it was merged in master by accident.

PR #3289, opened 2020-03-12
check traffic during rerouteV2 in multiTable mode

andriidovhan:


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
```diff
@@ -585,14 +611,12 @@ mode with existing flows and hold flows of different table-mode types"() {
                 changeSwitchPropsMultiTableValue(initSwProps[involvedSwitches[2].dpId], false))
```
niksv: I think i't better to use `switchHelper.updateSwitchProperties()` here and in line 503 because switchHelper waits for rules installation

andriidovhan: We can't use mentioned helper method everywhere.
Now and then default rules are not reinstalled according to switchProps. It depends on flow which are on a switch.
So it leaves me no choice but to do it on my own. I created separate method for updating swProps(`changeSwitchPropsMultiTableValue`) and checking default rules(`checkDefaultRulesOnSwitches`).
The  `checkDefaultRulesOnSwitches`  method waits for rules installation (L1307)

PR #3303, opened 2020-03-17
Minor tests adjustments

rtretyak:


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/resilience/ChaosSpec.groovy
```diff
@@ -70,9 +69,6 @@ class ChaosSpec extends HealthCheckSpecification {
             }
         }
 
-        and: "All flows are writing stats"
-        statsHelper.verifyFlowsWriteStats(flows*.flowId)
-
```
andriidovhan: Why did you decide not to check stats?

rtretyak: very time consuming ( Decided that not worth it. But I'm not sure that it's the best solution now when you ask

andriidovhan: it is fine to me. We have separate tests for that purpose.

PR #3304, opened 2020-03-17
Multiple updates to EnduranceV2Spec

rtretyak:
Including new test for pre-existing env

### DIFF
--- a/src-java/testing/performance-tests/kilda.properties.example
+++ b/src-java/testing/performance-tests/kilda.properties.example
@@ -14,42 +14,62 @@ floodlight.regions=1,2
 floodlight.alive.timeout=10
 floodlight.alive.interval=2
 
-grpc.endpoint=http://localhost:8091
-grpc.username=kilda
-grpc.password=kilda
-grpc.remote.log.server.ip=1.1.1.1
-grpc.remote.log.server.port=321
-
-elasticsearch.endpoint = http://localhost:9200
-elasticsearch.username = kilda
-elasticsearch.password = kilda
-
-kafka.bootstrap.server = kafka.pendev:9092
-kafka.topic.topo.disco=kilda.topo.disco
-
-lockkeeper.endpoint=http://localhost:5001
 neo.uri=bolt://localhost:7687
 neo.user=
 neo.password=
+
+elasticsearch.endpoint=http://localhost:9200
+elasticsearch.username=kilda
+elasticsearch.password=kilda
+elasticsearch.index=kilda-*
+
+environment.naming.prefix=
+kafka.bootstrap.server=kafka.pendev:9092
+kafka.topic.topo.disco=kilda.topo.disco.storm
+
 lab-api.endpoint=http://localhost:8288
+lockkeeper.port=5001
 
 opentsdb.endpoint=http://localhost:4242
 opentsdb.metric.prefix=kilda.
 
 spring.profiles.active=virtual
-mode.topology.single=true
 
-reroute.delay=3
-reroute.hardtimeout=15
+reroute.delay=2
+reroute.hardtimeout=8
+
 discovery.interval=3
-discovery.timeout = 15
+discovery.timeout=15
+
 antiflap.min=1
-antiflap.warmup=5
-antiflap.cooldown=10
-burst.coefficient=1.05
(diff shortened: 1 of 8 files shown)


src-java/testing/performance-tests/kilda.properties.example
```diff
@@ -14,42 +14,62 @@ floodlight.regions=1,2
...
+
+grpc.endpoint=http://localhost:8091
+grpc.username=kilda
+grpc.password=kilda
+#please use the correct ip address and port for the remote log server
+grpc.remote.log.server.ip=1.1.1.1
+grpc.remote.log.server.port=321
+
 use.hs=true
+use.multitable=false
```
andriidovhan: I guess the `use.multitable` option should be `true` by default

I mean if someone decide to copy this config, then they will be sure that use the same configuration as we

src-java/testing/performance-tests/src/test/groovy/org/openkilda/performancetests/spec/EnduranceV2Spec.groovy
```diff
@@ -92,22 +100,17 @@ idle, mass manual reroute, isl break. Step repeats pre-defined number of times"
...
         def allFlows = northbound.getAllFlows()
         def assertions = new SoftAssertions()
-        assertions.checkSucceeds {
-            statsHelper.verifyFlowsWriteStats(allFlows.findAll { it.status == FlowState.UP.toString() }*.id)
-        } ?: true
-
-        and: "All Up flows are pingable"
         def pingVerifications = new SoftAssertions()
         allFlows.findAll { it.status == FlowState.UP.toString() }.forEach { flow ->
             pingVerifications.checkSucceeds {
```
andriidovhan: How about checking that flow is pingable?

src-java/testing/performance-tests/src/test/groovy/org/openkilda/performancetests/spec/EnduranceV2Spec.groovy
```diff
@@ -175,18 +178,127 @@ idle, mass manual reroute, isl break. Step repeats pre-defined number of times"
...
+        }
+
+        and: "Wait for any blinking isls to get UP and flows to finish rerouting"
+        Wrappers.wait(antiflap.getAntiflapCooldown() + discoveryInterval + WAIT_OFFSET * 3) {
+            def isls = northbound.getAllLinks()
+            (topology.isls - brokenIsls).each {
+                assert islUtils.getIslInfo(isls, it).get().state == IslChangeType.DISCOVERED
+            }
+        }
+        Wrappers.wait(WAIT_OFFSET + flows.size() * 0.8) {
```
andriidovhan: Could you explain or leave a comment about magic numbers(`60`, `0.8`, `WAIT_OFFSET * 3`) ?
How did you find them?

src-java/testing/performance-tests/src/test/groovy/org/openkilda/performancetests/spec/EnduranceV2Spec.groovy
rtretyak: Can be added, but doesn't really make profit since we always use virtual switches in this test, it is not designed to be run elsewhere

src-java/testing/performance-tests/kilda.properties.example
rtretyak: ye, thx. fixed

src-java/testing/performance-tests/src/test/groovy/org/openkilda/performancetests/spec/EnduranceV2Spec.groovy
rtretyak: Added a comment. Tbh it is just a big wait which somehow depends on flows amount. Nobody knows the real formula)

rtretyak: '60' is for converting minutes to seconds. do we really need a comment there?

andriidovhan: I just wondered

PR #3308, opened 2020-03-18
Added PacketInOutStats to GRPC API

niksv:


src-java/grpc-speaker/grpc-service/src/main/java/org/openkilda/grpc/speaker/mapper/NoviflowResponseMapper.java
```diff
@@ -44,4 +47,23 @@
...
+        if (yesNo == null) {
+            return null;
+        }
+        switch (yesNo) {
+            case YES:
+                return true;
+            case NO:
+                return false;
+            default:
+                return null;
```
timofei-durakov: it slightly overlaps with line 56 check

src-java/grpc-speaker/grpc-api/src/main/java/org/openkilda/grpc/speaker/model/PacketInOutStatsResponse.java
```diff
@@ -0,0 +1,44 @@
...
+import lombok.NoArgsConstructor;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+@JsonNaming(SnakeCaseStrategy.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class PacketInOutStatsResponse {
+
+    private Long packetInTotalPackets;
```
timofei-durakov: why all of these are wrappers, not primitives?

niksv: fixed, except Boolean field because it can be null in some cases

src-java/grpc-speaker/grpc-service/src/main/java/org/openkilda/grpc/speaker/mapper/NoviflowResponseMapper.java
niksv: We can't remove check in line 57 because it will cause NPE in line 60.
Also we can't remove `default` section because:
1. stylecheck will raise a warning
2. YesNo enum has other values like `UNRECOGNIZED` and `YESNO_RESERVED`. We must handle them

PR #3315, opened 2020-03-18
Extrach ethernet packet body extractor into separate tool

surabujin:
This change is [Reviewable](https://reviewable.io/reviews/telstra/open-kilda/3315)

### DIFF
--- a/src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/utils/EthernetPacketToolbox.java
+++ b/src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/utils/EthernetPacketToolbox.java
@@ -1,0 +1,48 @@
+/* Copyright 2020 Telstra Open Source
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/utils/EthernetPacketToolbox.java
```diff
@@ -0,0 +1,50 @@
...
+        while (payload instanceof VlanTag) {
+            short vlanId = ((VlanTag) payload).getVlanId();
+            vlanStack.add((int) vlanId);
+            payload = payload.getPayload();
+        }
+
+        return payload;
+    }
+
+    protected EthernetPacketToolbox() {
```
niksv: why protected? not private?

surabujin: em. Can't say, replacing it with "private".

src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/utils/EthernetPacketToolbox.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.openkilda.floodlight.shared.packet.VlanTag;
+
+import net.floodlightcontroller.packet.Ethernet;
+import net.floodlightcontroller.packet.IPacket;
+
+import java.util.List;
+
+public final class EthernetPacketToolbox {
+    public static final EthernetPacketToolbox INSTANCE = new EthernetPacketToolbox();
```
sergii-iakovenko, line 26: nit: this could be a helper class with just static methods, no need to introduce a singleton. This will be less verbose.

PR #3317, opened 2020-03-20
Teach lockkeeper to add latency between FL and switches

rtretyak:


src-python/lock-keeper/lockkeeper.py
```diff
@@ -131,6 +132,32 @@ def fl_restart():
...
+                       'tc qdisc add dev eth0 root handle 1: prio priomap 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2',
+                       'tc qdisc add dev eth0 parent 1:1 handle 10: netem delay {}ms'.format(
+                           body['tcData']['egressDelayMs'])]
+    switches_by_container = defaultdict(list)
+    for sw in body['affectedAddresses']:
+        switches_by_container[sw['containerName']].append(sw)
+    for container, switches in switches_by_container.items():
+        commands = common_commands.copy()
+        for switch in switches:
+            commands.extend(['tc filter add dev eth0 protocol ip parent 1:0 prio 1 u32 match ip dst {}/32 match ip '
```
andriidovhan: Do we always use the `eth0` interface?
Probably would be better to use separate function for getting interface name, for example: 
`ip route show to match switch_address | awk  '{ print $5 }'`

src-python/lock-keeper/lockkeeper.py
```diff
@@ -131,6 +132,32 @@ def fl_restart():
...
+                             'dport {} 0xffff flowid 1:1'.format(switch['ip'], switch['port'])])
+        execute_commands_in_container(commands, container)
+    return jsonify({'status': 'ok'})
```
andriidovhan: Do you mind extending text response(for virtual and hardware implementations)?
for instance: 
* after executing `execute_commands_in_container`
* create list([[containerNames, ip, port],] 
* return it in response
`"tc filter is applied in the following containers: {}".format(<generated_list>)`

I guess it can be useful during investigating.

src-java/testing/test-library/src/main/java/org/openkilda/testing/service/lockkeeper/LockKeeperVirtualImpl.java
```diff
@@ -168,19 +168,34 @@ public void restartFloodlight(String region) {
...
         log.debug("Unblock floodlight access to {} by removing iptables rules", address);
         restTemplate.exchange(getCurrentLabUrl() + "/lock-keeper/floodlight/unblock",
                 HttpMethod.POST, new HttpEntity<>(address, buildJsonHeaders()), String.class);
     }
 
+    @Override
+    public void shapeSwitchesTraffic(List<Switch> switches, TrafficControlData tcData) {
+        List<FloodlightResourceAddress> swResources = switches.stream()
+                .map(sw -> LockKeeperService.toFlResource(sw, mgmtManager)).collect(Collectors.toList());
+        restTemplate.exchange(getCurrentLabUrl() + "/lock-keeper/floodlight/tc", HttpMethod.POST,
```
andriidovhan: please, add some log

src-java/testing/test-library/src/main/java/org/openkilda/testing/service/lockkeeper/LockKeeperVirtualImpl.java
```diff
@@ -83,23 +85,21 @@ public void setController(Switch sw, String controller) {
      * changes are reverted.
      */
     @Override
-    public BlockRequest knockoutSwitch(Switch sw, MultiFloodlightManager manager) {
-        log.debug("Block Floodlight access to switch '{}' by adding iptables rules", sw.getName());
```
andriidovhan: Why don't you like `log.debug`?)

src-python/lock-keeper/lockkeeper.py
rtretyak: We always use eth0. This code is executed inside floodlight container, not lab service

src-java/testing/test-library/src/main/java/org/openkilda/testing/service/lockkeeper/LockKeeperVirtualImpl.java
rtretyak: It is misleading since we are not using iptables in virtual implementation, we set dummy controller instead. The setController already has a debug log for changing controller

src-python/lock-keeper/lockkeeper.py
andriidovhan: agree

src-java/testing/test-library/src/main/java/org/openkilda/testing/service/lockkeeper/LockKeeperVirtualImpl.java
rtretyak: done

src-python/lock-keeper/lockkeeper.py
rtretyak: done

src-java/testing/test-library/src/main/java/org/openkilda/testing/service/lockkeeper/LockKeeperVirtualImpl.java
andriidovhan: agree

PR #3326, opened 2020-03-24
Add ip ad tc tools for traffic control

sirmax123:


rtretyak:
merged it to `functional/lockkeper_add_tc`

Issue #3256, opened 2020-03-03
v2 Delete: unable to remove flow with excess flow paths


rtretyak:
Given a flow that has excess/unused/garbage flow paths (see #3255 or #3090 or #3299 for reproduce).
- Try deleting such flow via V2 API
**Expected:** Flow is removed with all related resources
**Actual:** System starts the removal process (flow moves to 'In Progress'), but fails to remove the flow and changes status to previous (Down or Up). Related errors:
```
Flow perfload_28Feb133829_333_apples1366 error - RemoveRulesAction failed: No encapsulation resources found for flow path perfload_28Feb133829_333_apples1366_b30ac7dd-9078-4017-b692-f38bdac56c83 (opposite: perfload_28Feb133829_333_apples1366_062fa7ca-89cd-4984-8b10-ac3597b2972d)
```

PR #3360, opened 2020-04-02
minor fixes in DefaultRulesSpec

andriidovhan:
- ignore ISL related rules in DefaultRulesSpec

### DIFF
--- a/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowCrudV2Spec.groovy
+++ b/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowCrudV2Spec.groovy
@@ -1129,10 +1129,12 @@ class FlowCrudV2Spec extends HealthCheckSpecification {
         }
 
         and: "The new and old dst switches pass switch validation"
-        [dstSwitch, newDstSwitch]*.dpId.each { switchId ->
-            with(northbound.validateSwitch(switchId)) { validation ->
-                validation.verifyRuleSectionsAreEmpty(["missing", "excess", "misconfigured"])
-                validation.verifyMeterSectionsAreEmpty(["missing", "excess", "misconfigured"])
+        Wrappers.wait(RULES_DELETION_TIME / 2) {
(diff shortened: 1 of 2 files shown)


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowCrudV2Spec.groovy
```diff
@@ -1131,10 +1131,12 @@ class FlowCrudV2Spec extends HealthCheckSpecification {
         }
 
         and: "The new and old dst switches pass switch validation"
-        [dstSwitch, newDstSwitch]*.dpId.each { switchId ->
-            with(northbound.validateSwitch(switchId)) { validation ->
-                validation.verifyRuleSectionsAreEmpty(["missing", "excess", "misconfigured"])
-                validation.verifyMeterSectionsAreEmpty(["missing", "excess", "misconfigured"])
+        Wrappers.wait(RULES_DELETION_TIME / 2) {
```
rtretyak, line 1134: why?

andriidovhan: it is difficult to reproduce. Probably we need to run it around 30 times and one them is failed.
it is a rare case, but sometimes it happens.

andriidovhan: I often see failed build on jenkins due to this test

rtretyak: Ok, we can try, but the excess rules you're seeing may be due to #3256, and this test is just the first in the sequence that does the rules verificaiton and shows the problem

andriidovhan: this issue is not caused by other test.
I've rerun this test locally a lot of times and the issue sometimes is reproduced

PR #3366, opened 2020-04-06
Added GRPC service stub

niksv:


confd/templates/docker-compose/docker-compose.tmpl
```diff
@@ -4,6 +4,21 @@
 version: '2.1'
 services:
 
+  grpc-stub:
```
rtretyak, line 7: tbh this doesn't look like a mandatory part of the system, thus should be wrapped with `{{if not (exists "/no_grpc_stub")}}` thing

niksv: fixed

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/DockerHelper.groovy
```diff
@@ -0,0 +1,44 @@
...
+    void restartContainer(String containerId) {
+        dockerClient.restartContainer(containerId)
+    }
+
+    void waitContainer(String containerId) {
+        dockerClient.waitContainer(containerId)
+    }
+
+    private String getNetworkName() {
+        dockerClient.listNetworks()*.name().find { it.contains('_default')}
```
timofei-durakov, line 42: this could lead to wrong network finding

niksv: fixed

Issue #3291, opened 2020-03-13
Delete Flow Topology


timofei-durakov:
Is partially replaced by FLOW HS, the full replacement is consist of:
1.  Read Operations Moving to NB Worker Topology:
  * read flow by flow id[2d]
  * dump all flows[2d]
2. Bulk Update 2 Flows(Swap Endpoints op)[3+1w]  - Nb Worker Topology
3. Path Swap Within a Flow [2w]  - Flow H&S Topology
4. Flow Push/Unpush - proposed for removal

PR #3371, opened 2020-04-08
Neo4j changes for migrating swap endpoints.

dpoltavets:
Added methods for getting path ids by flow set and for calculating the sum of bandwidth of the flow set.

Related with: #3291


src-java/kilda-persistence-api/src/main/java/org/openkilda/persistence/repositories/FlowRepository.java
```diff
@@ -91,4 +92,6 @@
      * components must use this method, which guarantee safety such flows status.
      */
     void updateStatusSafe(String flowId, FlowStatus flowStatus);
+
+    long computeFlowsBandwidthSum(Set<String> flowIds);
```
timofei-durakov, line 96: nit: javadoc

PR #3376, opened 2020-04-08
Remove potential reordering of OF events

### LINKED ISSUE #3369: Periodic fl router sync in rare cases may disrupt ISL discovery process on port
**Scenario:**
given: port is in DOWN state
1. Floodlight router periodic sync happens
2. simultaneously with sync a port UP event happens
At this point race appears. Because sync uses correlationId as key when writing to kafka and portUp event uses switchId, the order of events may disrupt, so that portUp event gets immediately followed buy portDown state (comming from sync), causing the related port to be considered by the system as 'Down'. This in turn will prevent a related ISL on that port to be discovered until next periodic sync happens.
**Expected:** Disruption of port events order should not happen

surabujin:
Use "correct" kafka-key for each switch-dump event produced by
OF-speaker on network-dump request. It guarantees correct selection of
kafka-partition for such messages and as a result, it guarantees correct
ordering of produced events.

Close #3369

### DIFF
--- a/src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchTrackingService.java
+++ b/src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchTrackingService.java
@@ -28,7 +28,6 @@ import org.openkilda.floodlight.utils.CorrelationContext;
 import org.openkilda.floodlight.utils.FloodlightDashboardLogger;
 import org.openkilda.floodlight.utils.NewCorrelationContextRequired;
 import org.openkilda.messaging.Message;
-import org.openkilda.messaging.info.ChunkedInfoMessage;
 import org.openkilda.messaging.info.InfoData;
 import org.openkilda.messaging.info.InfoMessage;
 import org.openkilda.messaging.info.discovery.NetworkDumpSwitchData;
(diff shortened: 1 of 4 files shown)


src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/switchmanager/SwitchTrackingService.java
```diff
@@ -167,20 +166,11 @@ public void setup(FloodlightModuleContext context) {
...
-                producerService.sendMessageAndTrack(discoveryTopic,
-                                                correlationId,
-                                                new ChunkedInfoMessage(swData, System.currentTimeMillis(),
-                                                        correlationId, switchCounter, iofSwitches.size(), region));
-            } catch (Exception e) {
-                logger.error("Failed to send network dump for {}", sw.getId());
-            }
-            switchCounter++;
+            NetworkDumpSwitchData payload = new NetworkDumpSwitchData(buildSwitch(sw));
+            emitDiscoveryEvent(sw.getId(), payload);
```
sergii-iakovenko, line 173: nit: Why is "emit discovery event" used? This is not related to it.

surabujin: I can't produce a better name at this moment. Can you propose some? `emitDiscoveryEvent` is used to send any event into network-topology. Most of them are "discovery" events, and only this one "dump" is not exactly discovery event. But from the consumer perspective the switch-dump is a discovery message too. So I have named it as `emitDiscoveryEvent`.

PS I open to propositions.

PR #3381, opened 2020-04-10
adjust grpc tests to work with grpc-stub

andriidovhan:
- adjust grpc tests to work with grpc-stub;
- add DockerHelper;
- fix wfmManipulator;
- unignore StormLcmSpec;
- add new endpoint to lockkeeper(`get-container-ip/<string:name>`)
- fix MultiRerouteSpec(sequence of cleanup)

### DIFF
--- a/confd/templates/atdd-staging/kilda.properties.tmpl
+++ b/confd/templates/atdd-staging/kilda.properties.tmpl
@@ -77,3 +77,4 @@ latency.update.time.range = {{ getv "/kilda_latency_update_time_range" }}
 latency.discovery.interval.multiplier = {{ getv "/kilda_latency_discovery_interval_multiplier" }}
 
 flow.ping.interval=5
+docker.host=localhost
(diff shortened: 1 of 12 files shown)


src-java/testing/test-library/src/main/java/org/openkilda/testing/service/lockkeeper/LockKeeperService.java
```diff
@@ -69,6 +69,8 @@
 
     void reviveFloodlight(String region);
 
+    String getContainerIp(String containerName);
```
rtretyak, line 72: why on earth will you put this local-only code into lockkeeper and request container name over http? Do it directly in test classes
`def dockerClient = DefaultDockerClient.fromEnv().build()`

andriidovhan: in this case we will not be able to run grpc tests from remote machine

rtretyak: Maybe it's cleaner to do sth like `DefaultDockerClient.fromEnv().uri(whereTheDockerIs).build()` in this case? I mean it fixes the issue with container ip for grpc, but we have other tests that manipulate with docker containers and they remain unfixed

PR #3382, opened 2020-04-13
Added switch properties for server 42 Flow RTT feature

niksv:
* server_42_ping
* server_42_port
* server_42_mac_address

### DIFF
--- a/src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/model/SwitchPropertiesDto.java
+++ b/src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/model/SwitchPropertiesDto.java
@@ -16,6 +16,7 @@
 package org.openkilda.messaging.model;
 
 import org.openkilda.messaging.payload.flow.FlowEncapsulationType;
+import org.openkilda.model.MacAddress;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.Data;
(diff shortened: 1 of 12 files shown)


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchPropertiesSpec.groovy
```diff
@@ -104,6 +105,59 @@ class SwitchPropertiesSpec extends HealthCheckSpecification {
         null                          | "Supported transit encapsulations should not be null or empty"
     }
 
+    @Tidy
+    def "Error is returned when trying to #data.desc"() {
```
rtretyak, line 109: it is mandatory to explicitly annotate the test with `@Unroll` to get the name pattern unrolled into multiple tests. Otherwise it will look like single test with `#data.desc` not being replaced with proper value

niksv: fixed

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/model/SwitchPropertiesDto.java
```diff
@@ -36,4 +37,13 @@
 
     @JsonProperty("switch_arp")
     private boolean switchArp;
+
+    @JsonProperty("server_42_ping")
```
nikitamarchenko, line 41: flow_rtt

niksv: fixed

PR #3393, opened 2020-04-15
Expose discovery latency as separate log field

surabujin:


src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/pathverification/PathVerificationService.java
```diff
@@ -601,7 +604,7 @@ private void handleDiscoveryPacket(OfInput input, IOFSwitch destSwitch, Discover
                 source.getPortNo(),
                 destination.getSwitchId(),
                 destination.getPortNo(),
-                latency,
+                scaleLatencyMsToNs(latencyMs),  // TODO(surabujin): do we really need fake ns accuracy?
```
niksv, line 607: yes, we need in because this value could be used in
1. Isl latency database field
2. opentsdb `rtt` metric
both these metrics are in nano seconds

src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/pathverification/PathVerificationService.java
```diff
@@ -740,4 +740,14 @@ private long getSwitchPortSpeed(IOFSwitch sw, OFPort inPort) {
     private long getAvailableBandwidth(long speed) {
         return (long) (speed * islBandwidthQuotient);
     }
+
+    /**
+     * Convert milliseconds into nanoseconds, taking in account illegal (-1) latency value.
+     */
+    public static long scaleLatencyMsToNs(long latencyMs) {
```
sergii-iakovenko, line 747: Shouldn't the method be private? I don't a reason for  PathVerificationService to expose such methods.

surabujin: yes, it should. There was a plan to use it outside, but this is not required anymore.

src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/pathverification/PathVerificationService.java
surabujin: So we should make this conversion right before storing it in the mentioned storage. At least the fake ns accuracy conversion reason will be more obvious.

PR #3395, opened 2020-04-20
Fixed incorrect Connected devices packet logging

niksv:


src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/service/connected/ConnectedDevicesService.java
```diff
@@ -149,13 +149,16 @@ private void handlePacketIn(OfInput input) {
...
-        logger.info("Receive connected device packet from {} OF-xid:{}", input.getDpId(), input.getMessage().getXid());
         long cookie = rawCookie.getValue();
         SwitchId switchId = new SwitchId(input.getDpId().getLong());
 
         if (isLldpRelated(cookie)) {
+            logger.info("Receive connected device LLDP packet from {} OF-xid:{}, cookie: {}",
+                    input.getDpId(), input.getMessage().getXid(), cookie);
             handleSwitchLldp(input, switchId, cookie);
         } else if (isArpRelated(cookie)) {
+            logger.info("Receive connected device ARP packet from {} OF-xid:{}, cookie: {}",
```
timofei-durakov, line 160: even though we have a meter on packet in rule, won't we end up with a bunch of ARP?

niksv: log level changed to debug

PR #3416, opened 2020-04-27
SpeakerRequestBuildContext refactoring

niksv:
* all variables were placed into `forward` and `reverse` fiends
* `RemoveSharedRulesContext` was renamed to `RulesContext` because it will contain other options for server 42.

### DIFF
--- a/src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/model/SpeakerRequestBuildContext.java
+++ b/src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/model/SpeakerRequestBuildContext.java
@@ -25,10 +25,19 @@ import lombok.EqualsAndHashCode;
 @AllArgsConstructor
 @EqualsAndHashCode
 public class SpeakerRequestBuildContext {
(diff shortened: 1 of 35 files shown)


src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/model/SpeakerRequestBuildContext.java
```diff
@@ -25,10 +25,28 @@
...
+            .reverse(PathContext.builder().build())
+            .build();
+
+    private PathContext forward;
+    private PathContext reverse;
+
+    /**
+     * Swap forward and reverse PathContext.
+     */
+    public void swap() {
```
dpoltavets, line 39: I am not sure if this is a good idea. It is easy to forget in what condition we were and in what condition we will be after this operation. For me, an explicit constructor is better. And instead of this method, call the constructor to get a new instance.

niksv: okay

PR #3422, opened 2020-04-28
Remove Connected Devices Service log

niksv:
This log was deleted in #3395 but PR #3313 brought it back because of rebasing

### DIFF
--- a/src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/service/connected/ConnectedDevicesService.java
+++ b/src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/service/connected/ConnectedDevicesService.java
@@ -138,7 +138,6 @@ public class ConnectedDevicesService implements IService, IInputTranslator {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/service/connected/ConnectedDevicesService.java
```diff
@@ -134,11 +135,11 @@ ArpPacketData deserializeArp(Ethernet eth, SwitchId switchId, long cookie) {
     private void handlePacketIn(OfInput input) {
         U64 rawCookie = input.packetInCookie();
 
-        if (rawCookie == null) {
+        if (rawCookie == null || rawCookie.getValue() == UNKNOWN_COOKIE) {
```
surabujin, line 138: You should use https://github.com/telstra/open-kilda/blob/develop/src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/model/OfInput.java#L75 here.

PR #3423, opened 2020-04-28
Bitfield numberic enum types safe access

surabujin:
Provide save interface for bit fields represented as enum types.

### DIFF
--- a/src-java/kilda-model/src/main/java/org/openkilda/model/ServiceCookie.java
+++ b/src-java/kilda-model/src/main/java/org/openkilda/model/ServiceCookie.java
@@ -22,6 +22,8 @@ import org.openkilda.model.bitops.NumericEnumField;
 import lombok.Builder;
 import org.apache.commons.lang3.ArrayUtils;
 
+import java.util.Optional;
+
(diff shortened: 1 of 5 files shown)


src-java/kilda-model/src/main/java/org/openkilda/model/ServiceCookie.java
```diff
@@ -63,7 +65,18 @@ public void validate() throws InvalidCookieException {
...
     // TODO - replace existing meter-id generation approach
     public MeterId getMeterId() {
-        return new MeterId(getServiceTag().getValue());
+        return new MeterId(getField(SERVICE_TAG_FIELD));
+    }
+
+    /**
+     * Extract and return "service tag" field is save way (return empty {@link Optional} object if tag is invalid).
+     */
+    public Optional<ServiceCookieTag> getServiceTagSafe() {
```
timofei-durakov, line 74: I don't like idea having Safe method here, why not to change existing one?

surabujin: It can be done only if we don't want simple/raw access.

PR #3428, opened 2020-04-29
group cookie classes in single package

timofei-durakov:


src-java/kilda-model/src/main/java/org/openkilda/model/bitops/BitField.java
```diff
@@ -15,7 +15,7 @@
 
 package org.openkilda.model.bitops;
 
-import org.openkilda.model.Cookie;
+import org.openkilda.model.cookie.Cookie;
```
surabujin, line 18: (nip) I would like to remove this dependency.

PR #3432, opened 2020-04-29
Dispatch v1 flow create to flow hs

timofei-durakov:
- changed worker for flow v1 create request
- fixed compile time warnings

### DIFF
--- a/src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/converter/FlowMapper.java
+++ b/src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/converter/FlowMapper.java
@@ -28,6 +28,7 @@ import org.openkilda.messaging.model.Ping;
 import org.openkilda.messaging.model.SwapFlowDto;
 import org.openkilda.messaging.nbtopology.response.FlowValidationResponse;
 import org.openkilda.messaging.payload.flow.DetectConnectedDevicesPayload;
+import org.openkilda.messaging.payload.flow.FlowCreatePayload;
 import org.openkilda.messaging.payload.flow.FlowEncapsulationType;
 import org.openkilda.messaging.payload.flow.FlowEndpointPayload;
 import org.openkilda.messaging.payload.flow.FlowIdStatusPayload;
@@ -73,7 +74,7 @@ public abstract class FlowMapper {
     /**
      * Map FlowReadResponse.
      *
-     * @param r  {@link FlowReadResponse} instance.
+     * @param r {@link FlowReadResponse} instance.
      * @return {@link FlowResponsePayload} instance.
      */
     public FlowResponsePayload toFlowResponseOutput(FlowReadResponse r) {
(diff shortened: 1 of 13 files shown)


src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/converter/FlowMapper.java
```diff
@@ -159,6 +160,42 @@ default FlowResponsePayload toFlowResponseOutput(FlowReadResponse r) {
     @Mapping(target = "type", ignore = true)
     FlowRequest toFlowRequest(FlowRequestV2 request);
 
+
+    @Mapping(target = "flowId", source = "id")
+    @Mapping(target = "sourceSwitch", expression = "java(request.getSource().getSwitchDpId())")
```
surabujin, line 165: You don't need eval expression here - instead, you can use `source = "source.switchDpId"`.

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/service/impl/FlowServiceImpl.java
```diff
@@ -170,17 +169,17 @@
...
             logger.error("Can not parse arguments: {}", e.getMessage(), e);
             throw new MessageException(correlationId, System.currentTimeMillis(), ErrorType.DATA_INVALID,
                     e.getMessage(), "Can not parse arguments of the create flow request");
         }
 
         CommandMessage request = new CommandMessage(
-                payload, System.currentTimeMillis(), correlationId, Destination.WFM);
+                flowRequest, System.currentTimeMillis(), correlationId, Destination.WFM);
 
         return messagingChannel.sendAndGet(topic, request)
```
surabujin, line 184: Should it be placed into `flowHsTopic`?

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/service/impl/FlowServiceImpl.java
```diff
@@ -170,17 +169,17 @@
         final String correlationId = RequestCorrelationId.getId();
         logger.info("Create flow: {}", input);
 
-        FlowCreateRequest payload;
+        FlowRequest flowRequest;
         try {
-            payload = new FlowCreateRequest(new FlowDto(input), input.getDiverseFlowId());
+            flowRequest =  flowMapper.toFlowCreateRequest(input);
```
surabujin, line 174: Are we ready for a complete switch to H&S create? I.e. remove toggle for a fast rollback on v1 create? (Current mapping have such toggle).

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/converter/FlowMapper.java
```diff
@@ -159,6 +160,42 @@ default FlowResponsePayload toFlowResponseOutput(FlowReadResponse r) {
...
+    /**
+     * Map FlowCreatePayload.
+     *
+     * @param source {@link FlowCreatePayload} instance.
+     * @return {@link FlowRequest} instance.
+     */
+    default FlowRequest toFlowCreateRequest(FlowCreatePayload source) {
+        FlowRequest target = toFlowRequest(source).toBuilder().type(Type.CREATE).build();
+        DetectConnectedDevicesDto detectConnectedDevicesDto = new DetectConnectedDevicesDto();
+        if (source.getSource().getDetectConnectedDevices() != null) {
```
surabujin, line 186: I would like to have a copy constructor for `DetectConnectedDevicesDto` or mapper method to copy such objects (at least it remove code duplication).

rtretyak:
This will result in a slightly different format of some of the error messages (e.g. bad request). Don't see it as a problem at this point, but maybe sbody can think of any issues related to this

rtretyak:
Due to this change simultaneous v1 create(which is now v2) + v1 delete may cause some rule inconsistencies. Not a blocker as well, just observations

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/converter/FlowMapper.java
rtretyak: fixed

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/service/impl/FlowServiceImpl.java
rtretyak: fixed

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/converter/FlowMapper.java
timofei-durakov: in v1 detect info was part of flow endpoint, in v2 it's a separate entity. 
v1: FlowCreatePayload::FlowEndpointPayload::DetectConnectedDevicesPayload(separation based on endpoint level
v2: FlowRequest::DetectConnectedDevicesDto - single object, separation based on field level, 
e.g. srcLldp/dstLldp

Does it make sense to implement such mapper? Could see little gain here.

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/service/impl/FlowServiceImpl.java
timofei-durakov: It should be just tested, whole point to change worker on permanent basis

Issue #940, opened 2018-06-27
Moving off of Neo4J


carmine:
ONLY Neo4J EE supports scaling and HA. Proof: https://neo4j.com/docs/operations-manual/current/ha-cluster/basic-installation/

> Follow these steps in order to configure a Neo4j HA cluster:
> 
>     Download and install the **Neo4j Enterprise Edition** on each of the servers to be included in the cluster.

surabujin:
@carmine what benefits do we have from neo4j?

sergii-iakovenko:
Created two PoC with benchmarks: #2747 (Persistence / storage replacement PoC for Ferma + OrientDB) and #2770 (PCE replacement PoC for Ferma + OrientDB). Git branches: `neo4j-replacement-poc` and `neo4j-replacement-poc-2`.

sergii-iakovenko:
The next steps:
- Clustered OrientDB:
	* Run clustered OrientDB in dev & all-in-one enviroments.
	* Options for GUI / client to access data in the storage.
	* Data-model: is Ferma + data gateway pattern the only option? How to perform partial data reads / entity views?
	* Multi-threaded benchmark tests: deadlocks, locks, read-only access.
	* Options for PCE: native, Gremlin Server, Kilda in-memory. The requirements the same as https://github.com/telstra/open-kilda/pull/2770

- Other storages?

sergii-iakovenko:
The migration is completed. PCE improvements and minor issues related to OrientDB will be addressed separately.

PR #3433, opened 2020-04-30
Re-implement the data model for Tinkerpop-based persistence layer

sergii-iakovenko:
The changes:
- The data model has been revised and reimplemented to support frames of Ferma / Tinkerpop-based persistence layer.
- The implementation follows Row Data Gateway pattern and relies on composition of an entity and data gateway objects.

Related to #940

**MERGE NOTES: this PR must be merged along with #3434 and #3435**


src-java/kilda-model/src/main/java/org/openkilda/model/BfdSession.java
```diff
@@ -15,83 +15,138 @@
...
+        String remoteIpAddress;
+        @NonNull Integer port;
+        Integer discriminator;
+    }
+
+    /**
+     * A cloner for BfdSession entity.
+     */
+    @Mapper
+    public interface BfdSessionCloner {
```
surabujin, line 138: Why you prefer interface as the base for mapper? If it will be abstract class, it will allow restricting access to method that can be restricted (there is not too many such methods now, but they exists). Also in opposite to weird syntax hack (default methods in interfaces), code in methods of abstract classes are natural.

src-java/kilda-model/src/main/java/org/openkilda/model/FeatureToggles.java
```diff
@@ -15,100 +15,246 @@
...
-        this.deleteFlowEnabled = deleteFlowEnabled;
-        this.pushFlowEnabled = pushFlowEnabled;
-        this.unpushFlowEnabled = unpushFlowEnabled;
-        this.useBfdForIslIntegrityCheck = useBfdForIslIntegrityCheck;
-        this.floodlightRoutePeriodicSync = floodlightRoutePeriodicSync;
-        this.flowsRerouteViaFlowHs = flowsRerouteViaFlowHs;
-        this.flowsRerouteUsingDefaultEncapType = flowsRerouteUsingDefaultEncapType;
-        this.collectGrpcStats = collectGrpcStats;
+    @Data
+    @Builder
```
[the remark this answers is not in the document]
sergii-iakovenko: Right, it can be just a constructor. But IDEA will not help with pulling constructor arguments from FeatureTogglesDataImpl up to FeatureToggles constructor. Developers have to do it manually. And idea behind usage of builders here - to avoid mixing up the arguments (with a builder, you see the names)

src-java/kilda-model/src/main/java/org/openkilda/model/BfdSession.java
sergii-iakovenko: With interfaces it's a bit less verbose:
```
    public interface BfdSessionCloner {
        BfdSessionCloner INSTANCE = Mappers.getMapper(BfdSessionCloner.class);

        void copy(BfdSessionData source, @MappingTarget BfdSessionData target);

        default BfdSessionData copy(BfdSessionData source) {
            ...
        }
    }
```
vs
```
    public static abstract class BfdSessionCloner {
        public static final BfdSessionCloner INSTANCE = Mappers.getMapper(BfdSessionCloner.class);

        public abstract void copy(BfdSessionData source, @MappingTarget BfdSessionData target);

        public BfdSessionData copy(BfdSessionData source) {
            ...
        }
    }
```

No other preferences. Also, I found only 2 entities where we need non-public methods in a cloner - those can be changed to abstract classes.

PR #3434, opened 2020-04-30
Implement Tinkerpop-based persistence layer

sergii-iakovenko:
The changes:
- Ferma / Tinkerpop-based implementation for the persistence layer has been introduced (kilda-persistence-tinkerpop).
- The FetchStrategy approach has been removed (as entity relations in Ferma frames are lazy-loaded)
- The persistence context is a dedicated concept now (PersistenceContextRequired, PersistenceContextManager, etc.).

Related to #940

**MERGE NOTES: this PR must be merged only after #3433**


src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaBfdSessionRepository.java
```diff
@@ -0,0 +1,92 @@
...
+        super(graphFactory, transactionManager);
+    }
+
+    @Override
+    public Collection<BfdSession> findAll() {
+        return framedGraph().traverse(g -> g.V()
+                .hasLabel(BfdSessionFrame.FRAME_LABEL))
+                .toListExplicit(BfdSessionFrame.class).stream()
+                .map(BfdSession::new)
+                .collect(Collectors.toList());
```
surabujin, line 48: Why no to return Stream<?> in such methods? (just because we already have it and most times caller needs to make only one iteration over objects in response... so converting it into List is resource wasting)

sergii-iakovenko: It's a good point. When the initial implementation of the persistence layer was introduced, we inherited the Neo4j-OGM approach - fetching via collections. 
However, I don't think this task is a right time to refactor the api.

PR #3443, opened 2020-05-05
use signleTable by default for func tests

andriidovhan:


rtretyak:
Please fix in performance tests 'example' file as well

PR #3456, opened 2020-05-08
Changed alert level of mapstruct warning "unmappedTargetField" from WARN to ERROR

niksv:
PR #3276 fixed all unmappedTargetField warnings, but future patches bring new warns.

This patch changes alert level for WARN to ERROR to do not pass such warns in future


src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/DetectConnectedDevicesMapper.java
```diff
@@ -19,9 +19,10 @@
 import org.openkilda.model.DetectConnectedDevices;
 
 import org.mapstruct.Mapper;
+import org.mapstruct.ReportingPolicy;
 import org.mapstruct.factory.Mappers;
 
-@Mapper
+@Mapper(unmappedTargetPolicy = ReportingPolicy.ERROR)
```
sergii-iakovenko, line 25: I think we can set the config globally. Add the following argument to the compiler -Amapstruct.unmappedTargetPolicy=ERROR

Issue #3028, opened 2019-12-11
V2 Delete is unable to remove a rolled back DOWN flow


rtretyak:
1. Repeat steps from #2954 to receive a DOWN flow
2. Remove the target flow using V2 API
**Expected:** Flow is removed, switches are clean
**Actual:** Flow is stuck in 'In progress' state
[flow history](https://github.com/telstra/open-kilda/files/3950765/unable_to_delete_history.txt)

rtretyak:
fixed in scope of #2954

PR #3459, opened 2020-05-08
Fix NPE in H&S reroute and update thrown for a flow with no paths

### LINKED ISSUE #2954: V2 Flow Creation creates damaged flow if one of the endpoints disconnects in the process

sergii-iakovenko:
The changes:

- Proper handling of a flow with no paths or with incomplete pairs.

Closes #2954, #3028, #3467

### DIFF
--- a/src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/FlowPathMapper.java
+++ b/src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/FlowPathMapper.java
@@ -41,19 +41,20 @@ public abstract class FlowPathMapper {
      */
     public PathInfoData map(FlowPath path) {
         PathInfoData result = new PathInfoData();
-        result.setLatency(path.getLatency());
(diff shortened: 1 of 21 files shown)


sergii-iakovenko:
> Should we also carry about FlowPathSwapFsm?
FlowPathSwapFsm just holds the paths, there's no code which need to be updated. All refactoring code will be delivered as a separate PR.

Issue #3469, opened 2020-05-14
Global Reroute Retry Could not happen.


timofei-durakov:
According to logs in some case reroute procedure could end up with error. In that case there should be global retry happen.
According to logs:
`Flow f1 reroute result failed with error null` - which means there are no reroute error in fsm context.
OnFinishedWithErrorAction has a handler to send response back with reroute result:
``` @Override
    public void sendRerouteResultStatus(String flowId, RerouteError rerouteError, String correlationId) {
        RerouteResultInfoData rerouteResult = RerouteResultInfoData.builder()
                .flowId(flowId)
                .success(rerouteError == null)
                .rerouteError(rerouteError)
                .build();
        Message message = new InfoMessage(rerouteResult, System.currentTimeMillis(), correlationId);
        emitWithContext(Stream.HUB_TO_REROUTE_RESPONSE_SENDER.name(), getCurrentTuple(),
                new Values(currentKey, message));
    }```
Since rerouteError is null, success flag is set to true.

PR #3470, opened 2020-05-14
Fix for global re-route retry.

timofei-durakov:
This patch fills Reroute Error on each error event in fsm.

Closes: #3469


src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/BaseResourceAllocationAction.java
```diff
@@ -143,6 +146,15 @@ protected abstract void allocate(T stateMachine)
      */
     protected abstract void onFailure(T stateMachine);
 
+
+    /**
+     * Called in a case of allocation failure.
+     */
+    protected void recordFailureReason(T stateMachine, String reason) {
```
surabujin, line 153: As for me - the abstract method is better. It will not allow forgetting to define it in new code.

niksv:
@timofei-durakov please fix conflicts

sergii-iakovenko (approved):

PR #3471, opened 2020-05-14
Do not take into account orphaned paths while returning flows for end…

timofei-durakov:
…point

Added additional filtering of flow by orphaned paths


src-java/kilda-model/src/main/java/org/openkilda/model/Flow.java
```diff
@@ -274,6 +274,14 @@ public final void addPaths(FlowPath... paths) {
                 .findAny();
     }
 
+    /**
+     * Checks whether path is is orphaned or not.
+     */
+    public final boolean isValidPath(PathId pathId) {
```
surabujin, line 280: Em... it is a more limited copy of `isActualPathId()`. Why we need it?

timofei-durakov: good catch

Issue #3461, opened 2020-05-12
error in kibana while updating portProperties


andriidovhan:
**Step to reproduce:**
update portProperties

**Actual result:**
Feature works as expected, 
but in kibana we can see an error:
```
Could not resolve type id 'org.openkilda.messaging.nbtopology.request.UpdatePortPropertiesRequest' as a subtype of [simple type, class org.openkilda.messaging.info.InfoData]: no such class found
 at [Source: (byte[])"{"clazz":"org.openkilda.messaging.info.InfoMessage","payload":{"clazz":"org.openkilda.messaging.nbtopology.request.UpdatePortPropertiesRequest","switch_id":"00:00:00:00:00:00:00:04","port":4,"discovery_enabled":false,"timestamp":1589282902693},"timestamp":1589282902693,"correlation_id":"a1b2621e-ff37-4e32-a8af-5c93010f29f6 : adovhan_postman","destination":null,"region":null}"; line: 1, column: 72] (through reference chain: org.openkilda.messaging.info.InfoMessage["payload"])
```

PR #3481, opened 2020-05-18
Add missing dependency to port state topo

timofei-durakov:
UpdatePortPropertiesRequest is a part of nbworker-messaging.
Added dependency to build.gradle

Closes: #3461


src-java/portstate-topology/portstate-storm-topology/build.gradle
```diff
@@ -13,6 +13,7 @@ dependencies {
     runtimeClasspath project(path: ':base-storm-topology', configuration: 'releaseArtifacts')
     implementation project(':kilda-reporting')
     implementation project(':network-messaging')
+    implementation project(':nbworker-messaging')
```
surabujin, line 16: If some object used outside specific topology (that owns messaging packet) should it be moved into a more "common/shared" messaging packet?

PR #3484, opened 2020-05-18
Various test stability fixes

rtretyak:


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/AutoRerouteV2Spec.groovy
```diff
@@ -110,6 +110,8 @@ class AutoRerouteV2Spec extends HealthCheckSpecification {
         then: "The flow becomes 'Up'"
         Wrappers.wait(rerouteDelay + WAIT_OFFSET) {
             assert northboundV2.getFlowStatus(flow.flowId).status == FlowState.UP
+            assert northbound.getFlowHistory(flow.flowId).last().histories
+                .find { it.action == "The flow status was reverted to UP" }
```
andriidovhan, line 114: why not `"The flow status was set to UP"` or `"Flow was rerouted successfully"` ?

rtretyak: ye, makes sense

rtretyak: no, doesn't) It will always be the same path as it was before, so it will always be 'status reverted to up'
the history entry for this event is kinda strange =(
```
{
                "timestamp": 1589790868,
                "action": "Flow was validated successfully",
                "details": null
            },
            {
                "timestamp": 1589790868,
                "action": "Found the same primary path. Skipped creating of it",
                "details": null
            },
            {
                "timestamp": 1589790868,
                "action": "The flow status was reverted to UP",
                "details": null
            },
            {
                "timestamp": 1589790868,
                "action": "Failed to reroute the flow",
                "details": "Reroute is unsuccessful. Couldn't find new path(s)"
            }
```

PR #3490, opened 2020-05-20
Add "cost and available bandwidth" path computation strategy

rozdy:
Add new path computation strategy that finds path with best cost but if two paths has the same cost then choose the one with min available bandwidth.

### DIFF
--- a/src-java/kilda-pce/src/main/java/org/openkilda/pce/finder/BestWeightAndShortestPathFinder.java
+++ b/src-java/kilda-pce/src/main/java/org/openkilda/pce/finder/BestWeightAndShortestPathFinder.java
@@ -24,6 +24,7 @@ import org.openkilda.pce.exception.UnroutableFlowException;
 import org.openkilda.pce.impl.AvailableNetwork;
 import org.openkilda.pce.model.Edge;
 import org.openkilda.pce.model.Node;
+import org.openkilda.pce.model.PathWeight;
 import org.openkilda.pce.model.WeightFunction;
 
 import com.google.common.collect.Lists;
@@ -227,7 +228,7 @@ public class BestWeightAndShortestPathFinder implements PathFinder {
         if (path.isEmpty()) {
             return Long.MAX_VALUE;
         }
-        return path.stream().mapToLong(weightFunction::apply).sum();
+        return path.stream().map(weightFunction).mapToLong(PathWeight::toLong).sum();
     }
 
     private void removeEdge(Edge edge) {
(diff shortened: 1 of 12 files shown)


src-java/kilda-pce/src/main/java/org/openkilda/pce/model/PathWeight.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * Path weight.
+ */
+public class PathWeight implements Comparable<PathWeight> {
+
+    private List<Long> params;
```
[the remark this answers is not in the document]
niksv: I guess we need so generic param because PathWeight will be used with different strategies. And different strategies can have different param names.

src-java/kilda-pce/src/main/java/org/openkilda/pce/finder/BestWeightAndShortestPathFinder.java
```diff
@@ -227,7 +228,7 @@ private long getTotalWeight(List<Edge> path, WeightFunction weightFunction) {
         if (path.isEmpty()) {
             return Long.MAX_VALUE;
         }
-        return path.stream().mapToLong(weightFunction::apply).sum();
+        return path.stream().mapToLong(edge -> weightFunction.apply(edge).toLong()).sum();
```
niksv, line 231: nit: I think you can use 
`path.stream().map(weightFunction::apply).map(weightFunction::toLong()).sum()`

src-java/kilda-pce/src/main/java/org/openkilda/pce/model/PathWeight.java
```diff
@@ -0,0 +1,84 @@
...
+            if (first != second) {
+                return first > second ? 1 : -1;
+            }
+            i++;
+        }
+        if (firstSize == secondSize) {
+            return 0;
+        } else {
+            return firstSize > secondSize ? 1 : -1;
+        }
```
niksv, line 82: nit: you can use just `Integer.compare(firstSize, secondSize)`

src-java/kilda-pce/src/main/java/org/openkilda/pce/finder/BestWeightAndShortestPathFinder.java
rozdy: Fixed.

src-java/kilda-pce/src/main/java/org/openkilda/pce/model/PathWeight.java
rozdy: Fixed.

PR #3518, opened 2020-06-01
Suppress stale discovery requests on FL side

surabujin:


andriidovhan:
**How to test:**

1. create a topology;
2. stop FL container(the rest components of kilda is up);
3. observe `kibana`, `docker stats`, `ISLs status` (shouldn't see any anomalies, all ISLs should be up);
4. wait up to 15 minutes;
5. start FL container;
6. make sure that FL is up and system allows to CRUD flow;
7. observe kibana, docker stats, ISLs status (shouldn't see any anomalies, all ISLs should be up);


This PR was tested on two topologies

-  sw 50, isl 300
-  sw 70, isl 420

with two different options of `org.openkilda.floodlight.kafka.KafkaMessageCollector.discovery-flush-delay-millis` (100(default) and 250)
Each topology was tested 3 times with each option.

Topology sw/isl |  100(default) |  250
-- | -- | --
50/300 | 2/3 - ok; 1/3 - fail (-2 ISL) | 3/3 - ok
70/420 | 2/3 - ok; 1/3 - fail (-18 ISL) | 3/3 - ok

Issue #2977, opened 2019-11-28
Bump storm and kafka verisons


rozdy:
We need to bump storm and kafka versions to fix security vulnerabilities.

Changes related to fix this issue can be found in `bump-storm-and-kafka-libs` branch.

sergii-iakovenko:
The changes in #3519 cause the following issues:
-   All spouts have the "complete latency" and "acked" statistics equal to 0 (the stats can be checked in Storm UI). This may be caused by the "__acker" component, which lost associated streams ("__ack_init" and "__ack_ack") from bolts and spouts after the storm lib upgrade.

- On the development and staging environments, we observe uneven growth of "OTSDB ConsummerGroup Lag" (from 1 to 40 millions during 10 hours, then a steep drop) and gaps in metrics stored in OpenTSDB.

- There're Kafka related errors in the log: 
```
[Consumer clientId=consumer-6, groupId=floodlightrouter__SPEAKER_DISCO_KAFKA_SPOUT] Offset commit failed on partition kilda.speaker.disco.storm-1 at offset 192769769: The request timed out.
[Consumer clientId=consumer-4, groupId=floodlight_2] Offset commit failed on partition kilda.speaker.disco_2-0 at offset 138816474: The request timed out.
[Consumer clientId=consumer-4, groupId=floodlight_2] Offset commit failed on partition kilda.speaker.disco_2-1 at offset 80522141: The request timed out.
```

sergii-iakovenko:
Please, check the changes in Storm 1.2.1 vs 1.1.0 related to Kafka Spout:

PR #3519, opened 2020-06-01
Bump kafka and storm client libs versions

rozdy:


src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/kafka/InfoDataDeserializer.java
```diff
@@ -15,18 +15,40 @@
...
+    }
+
+    @Override
+    public InfoData deserialize(String topic, byte[] data) {
+        try {
+            return SerializationUtils.MAPPER.readValue(data, InfoData.class);
+        } catch (IOException e) {
+            log.error(format("Failed to deserialize data: %s from topic %s",
+                    StringUtils.toEncodedString(data, Charset.defaultCharset()), topic), e);
+            return InfoData.CORRUPTED_INFO_DATA;
```
timofei-durakov, line 46: are there way not to return anything in this case?

rozdy: Reworked to return InfoData with error details.

sergii-iakovenko:
Related to #2977

PR #3522, opened 2020-06-03
Fix log types for grpc speaker and server42

rozdy:


docker/logstash/etc/logstash/conf.d/logstash.conf.example
```diff
@@ -1,45 +0,0 @@
-input {
```
timofei-durakov, line 1: why did you remove example?

rozdy: Because it isn't used any more. This config is generated by corresponding template.

timofei-durakov: but this one is an example file, not a config itself

rozdy: Ok, reverted file back

PR #3551, opened 2020-06-18
refactor SwitchMaintenanceSpec

andriidovhan:


andriidovhan:
This test is already refactored in #3525

PR #3561, opened 2020-06-24
Add more tests to smoke_switches suite

rtretyak:


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/SwitchActivationSpec.groovy
```diff
@@ -41,7 +43,7 @@ class SwitchActivationSpec extends HealthCheckSpecification {
     @Autowired
     SwitchHelper switchHelper
 
-    @Tags(VIRTUAL)
+    @Tags([VIRTUAL, SMOKE, SMOKE_SWITCHES])
```
andriidovhan, line 46: looks like we can delete VIRTUAL tag

rtretyak: should be removed in the parallel pr about virtual tags. will fix conflicts here if they appear

PR #3568, opened 2020-06-25
Add test that tries to create a single-sw flow with periodic pings

rtretyak:


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowPingSpec.groovy
```diff
@@ -311,6 +314,26 @@ class FlowPingSpec extends HealthCheckSpecification {
         flow && flowHelperV2.deleteFlow(flow.flowId)
     }
 
+    @Tidy
```
andriidovhan, line 317: maybe `low_priority`?

rtretyak: Don't have a strong opinion here. Added low_priority

Issue #3571, opened 2020-06-29
Add new fields to Switch DTO


dpoltavets:
Need to add the `location` field to the Switch DTO, which contains `latitude`, `longitude`, `street`, `city` and `country`.

PR #3589, opened 2020-07-02
Added `location` field to Switch DTO

dpoltavets:
Closes: #3571

### DIFF
--- a/src-java/kilda-model/src/main/java/org/openkilda/model/Switch.java
+++ b/src-java/kilda-model/src/main/java/org/openkilda/model/Switch.java
@@ -108,10 +108,17 @@ public class Switch implements Serializable {
 
     private String pop;
 
+    private double latitude;
+    private double longitude;
+    private String street;
+    private String city;
+    private String country;
+
     @Builder(toBuilder = true)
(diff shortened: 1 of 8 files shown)


src-java/kilda-model/src/main/java/org/openkilda/model/Switch.java
```diff
@@ -123,6 +130,11 @@ public Switch(@NonNull SwitchId switchId, SwitchStatus status, InetSocketAddress
         this.timeModify = timeModify;
         setFeatures(features);
         this.pop = pop;
+        this.latitude = latitude;
+        this.longitude = longitude;
+        this.street = street;
+        this.city = city;
+        this.country = country;
```
niksv, line 137: Should we put these 5 fields into `SwitchLocation` class too?

timofei-durakov: No, we shouldn't, from db perspective it's a straightfowrward node attributes without grouping

dpoltavets: We discussed that we should store these fields directly on the switch in the database.

Issue #3572, opened 2020-06-29
Add PATCH update for `location` and `pop` fields in the Switch DTO


dpoltavets:
Create new endpoint: `PATCH /v2/switches/{switch_id}`
with request body:
```
{
    "pop": "string",
    "location": {
        "latitude": 0,
        "longitude": 0,
        "street" : "string", 
        "city" : "string", 
        "country" : "string"
    }
}
```

PR #3595, opened 2020-07-03
Added PATCH Switch operation for `location` and `pop` fields.

dpoltavets:
Closes: #3572


src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/controller/v2/SwitchControllerV2.java
```diff
@@ -140,4 +143,18 @@
...
+    /**
+     * Update switch.
+     *
+     * @param switchId the switch
+     * @return switch.
+     */
+    @ApiOperation(value = "Update switch", response = SwitchDtoV2.class)
+    @PatchMapping(value = "/{switch_id}")
+    @ResponseStatus(HttpStatus.OK)
+    public CompletableFuture<SwitchDtoV2> getSwitch(@PathVariable("switch_id") SwitchId switchId,
```
niksv, line 156: name should be updateSwitch or patchSwitch

dpoltavets: Fixed.

PR #3596, opened 2020-07-03
update partialUpdateSpec according to #3590

andriidovhan:


andriidovhan:
it is already implemented in #3597

PR #3599, opened 2020-07-06
Change format of History records in DB from ISO to epoch

niksv:


services/src/neo4j/migrations/1.25-change-type-of-flow-event-timestamp/rollback.cql
```diff
@@ -0,0 +1,3 @@
+MATCH (n:flow_event) set n.timestamp=apoc.date.toISO8601(n.timestamp)
```
sergii-iakovenko, line 1: I'd take precautions - like "toInteger(n.timestamp) == n.timestamp" in the "where" section.

src-java/kilda-persistence-neo4j/src/test/java/org/openkilda/persistence/repositories/impl/Neo4jFlowEventRepositoryTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+    private static FlowEventRepository repository;
+
+    @BeforeClass
+    public static void setUp() {
+        repository = new Neo4jFlowEventRepository(neo4jSessionFactory, txManager);
+    }
+
+    @Test
+    public void findByFlowIdAndTimeFrameOrderTest() {
```
sergii-iakovenko, line 50: I'd add test for the case with 0 millisecond part of the timestamp.

niksv: It's already there.  `TIME_1` is a timestamp with 0 milliseconds. On previous code version this test  fails.

services/src/neo4j/migrations/1.25-change-type-of-flow-event-timestamp/rollback.cql
niksv: ok. will add

PR #3613, opened 2020-07-08
add/refactor tests for smart discovery

andriidovhan:


andriidovhan:
pr is recreated (#3635 3635)

PR #3614, opened 2020-07-09
enable server42FlowRtt in feature toogle by default for func test

andriidovhan:


nikitamarchenko:
commit message is confusing, please add tests or something to commit message

PR #3636, opened 2020-07-15
fix "Traffic counters in ingress rule are reset on flow rerouting(multiTable mode)"

andriidovhan:
The test was unstable.

Reason:
Traffgen doesn't delete interface, this interface has auto-assigned IPv6 address, then as a result this interface sends an autodiscovery packet.
```
interface:
2: vlan.2686@gw: <BROADCAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether b2:a8:1c:f5:7f:c6 brd ff:ff:ff:ff:ff:ff
    inet6 fe80::b0a8:1cff:fef5:7fc6/64 scope link
       valid_lft forever preferred_lft forever


tcpdump:
2020-07-17 06:44:39.958198 b2:a8:1c:f5:7f:c6 (oui Unknown) > 33:33:00:00:00:02 (oui Unknown), ethertype 802.1Q (0x8100), length 74: vlan 2686, p 0, ethertype IPv6, fe80::b0a8:1cff:fef5:7fc6 > ip6-allrouters: ICMP6, router solicitation, length 16
```

### DIFF
--- a/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/FlowRulesSpec.groovy
+++ b/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/FlowRulesSpec.groovy
@@ -641,7 +641,6 @@ class FlowRulesSpec extends HealthCheckSpecification {
         flowHelperV2.deleteFlow(flow.flowId)
     }
 
-    @Ignore
     @Tidy
     @Tags([SMOKE, SMOKE_SWITCHES])
     def "Traffic counters in ingress rule are reset on flow rerouting(singleTable mode)"() {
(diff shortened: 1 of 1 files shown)


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/FlowRulesSpec.groovy
```diff
@@ -742,9 +742,10 @@ class FlowRulesSpec extends HealthCheckSpecification {
             [examVlanFlow.forward, examVlanFlow.reverse].eachParallel { direction ->
                 def resources = traffExam.startExam(direction)
                 direction.setResources(resources)
-                assert traffExam.waitExam(direction).hasTraffic()
+                assert traffExam.waitExam(direction)
             }
         }
+        sleep(4000) //packetCount can be written with some delay
```
rtretyak, line 748: you can do 'wait' on below operation instead of bare sleeping

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/FlowRulesSpec.groovy
```diff
@@ -742,9 +742,10 @@ class FlowRulesSpec extends HealthCheckSpecification {
             [examVlanFlow.forward, examVlanFlow.reverse].eachParallel { direction ->
                 def resources = traffExam.startExam(direction)
                 direction.setResources(resources)
-                assert traffExam.waitExam(direction).hasTraffic()
+                assert traffExam.waitExam(direction)
```
rtretyak, line 745: why change? It makes perfect sense to see traffic here so that later you expect packet counters growth

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/FlowRulesSpec.groovy
andriidovhan: We can't. We don't know exact number of packetCount which should we wait.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/FlowRulesSpec.groovy
andriidovhan: As for me it allows us to be sure that flow is not rerouted while traffic flow is active.
refactored to:
```
assert !traffExam.isFinished(direction)
traffExam.waitExam(direction)
```

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/FlowRulesSpec.groovy
rtretyak: no need to change any verifications. Just wrap the below 'then' inside wait.

andriidovhan:
The root cause is fixed in #3641

PR #3655, opened 2020-07-22
Remove default controller role hack

surabujin:
Correct way to define default switch role was added into floodlight
itself, so we can remove existing hack.


surabujin:
Do not merge until https://github.com/kilda/floodlight/pull/7 is merged.

Issue #3656, opened 2020-07-23
Discovery poll timeout is not changed after deleting failed ISL


andriidovhan:
**Steps to reproduce:**

1. find an active ISL _**(local env: isl with a-switch, hardware evn: isl acton-acton)**_
2. do portDown on the src side of the ISL
3. make sure the ISL is FAILED
4. delete the failed ISL
5. check discovery interval for the dst side

**Actual result:**
discovery interval is not changed

**Expected result:**
discovery interval should be changed to `discovery.exhausted.interval`

**NOTE:** 
The issue is NOT reproduced on a direct link(without a-switch) on ovs switches and on a link between noviflow switches

PR #3660, opened 2020-07-25
Fix link deleting in the Network topology

dpoltavets:
Closes: #3656


src-java/network-topology/network-storm-topology/src/main/java/org/openkilda/wfm/topology/network/storm/bolt/speaker/SpeakerRouter.java
```diff
@@ -157,7 +163,12 @@ private void proxySpeaker(Tuple input, InfoData payload) throws PipelineExceptio
             emit(STREAM_PORT_ID, input, makePortTuple(
                     new UpdatePortPropertiesCommand((UpdatePortPropertiesRequest) payload)));
         } else if (payload instanceof DeactivateIslInfoData) {
-            emit(STREAM_ISL_ID, input, makeIslTuple(input, new IslDeleteCommand((DeactivateIslInfoData) payload)));
+            DeactivateIslInfoData data = (DeactivateIslInfoData) payload;
+            emit(STREAM_ISL_ID, input, makeIslTuple(input, new IslDeleteCommand(data)));
+            emit(STREAM_UNI_ISL_ID, input,
+                    makeUniIslTuple(input, new UniIslSetupCommand(new Endpoint(data.getSource()))));
```
surabujin, line 169: Em... you reuse setup command to remove data. As for me such code reuse... makes more harm than good. At least you will have an extreme mess in logs - there is no person who will treat the message `"Uni-ISL service receive SETUP request for {}"` as "remove" of uni-isl data.

One more even bigger issue here - you "clean" uniIsl without any checks... IslFsm can reject or postpone removal, but you remove any way to communicate to it in advance, so all ISL clean logic becomes corrupted. You must clean data in cascade and only when the current step is done you can go to the next step.

So - only IslFsm/service can call further cleanup when it has done with its own cleanup. The easiest way here is to make the switch to the exhausted discovery mode from the Isl side as part of the removal process.

dpoltavets: You're right. I've added the switch to exhausted mode to ISL FSM.

src-java/network-topology/network-storm-topology/src/main/java/org/openkilda/wfm/topology/network/controller/isl/IslFsm.java
```diff
@@ -342,6 +343,11 @@ private void sendInstallMultiTable(IIslCarrier carrier, Endpoint ingress, Endpoi
         }
     }
 
+    private void sendEnableExhaustedPollMode(IIslCarrier carrier) {
+        carrier.exhaustedPollModeUpdateRequest(reference.getSource(), true);
```
surabujin, line 347: Why auxiliary mode controlled via bfdManger but exaustem mode controlled directly by Islsm?

PR #3674, opened 2020-08-03
Use floodlightrouter sw tracking for stats requests routing

surabujin:


src-java/floodlightrouter-topology/floodlightrouter-storm-topology/src/main/java/org/openkilda/wfm/topology/floodlightrouter/model/RegionMappingAdd.java
```diff
@@ -0,0 +1,29 @@
...
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+package org.openkilda.wfm.topology.floodlightrouter.model;
+
+import org.openkilda.model.SwitchId;
+
+public class RegionMappingAdd extends RegionMappingUpdate {
```
timofei-durakov, line 20: Do we really need a command pattern here?

src-java/floodlightrouter-topology/floodlightrouter-storm-topology/src/main/java/org/openkilda/wfm/topology/floodlightrouter/model/RegionMappingAdd.java
```diff
@@ -0,0 +1,29 @@
...
+    public RegionMappingAdd(SwitchId switchId, String region, boolean readWriteMode) {
+        super(switchId, region, readWriteMode);
+    }
+
+    @Override
+    public void apply(RegionMappingStorage storage) {
+        storage.update(this);
```
timofei-durakov, line 27: this inverse of action make things harder to track without any pros

src-java/floodlightrouter-topology/floodlightrouter-storm-topology/src/main/java/org/openkilda/wfm/topology/floodlightrouter/model/RegionMappingStorage.java
```diff
@@ -13,12 +13,12 @@
  *   limitations under the License.
  */
 
-package org.openkilda.wfm.topology.statsrouter.service;
+package org.openkilda.wfm.topology.floodlightrouter.model;
 
-import org.openkilda.messaging.Message;
+public interface RegionMappingStorage {
```
timofei-durakov, line 18: single implementation without any signs that any other will come in future, can't see any value from this

src-java/floodlightrouter-topology/floodlightrouter-storm-topology/src/main/java/org/openkilda/wfm/topology/floodlightrouter/model/RegionMappingAdd.java
surabujin: This is a visitor, not command.

src-java/floodlightrouter-topology/floodlightrouter-storm-topology/src/main/java/org/openkilda/wfm/topology/floodlightrouter/model/RegionMappingAdd.java
surabujin: This is a visitor pattern. What do you propose to use here?

src-java/floodlightrouter-topology/floodlightrouter-storm-topology/src/main/java/org/openkilda/wfm/topology/floodlightrouter/model/RegionMappingStorage.java
surabujin: In this specific case - this is part of visitor pattern implementation. I can say the same about all NB architecture and most part of switchmanager topology (and this is only what I can remember right now).

src-java/floodlightrouter-topology/floodlightrouter-storm-topology/src/main/java/org/openkilda/wfm/topology/floodlightrouter/model/RegionMappingAdd.java
timofei-durakov: my bad, was thinking about the command pattern there at that moment). looks like it won't work here since you need to inject storage in it somehow, and finally end up with visitor.

timofei-durakov (approved):

PR #3684, opened 2020-08-11
Add retries for grpc calls in functional tests

rtretyak:


src-java/testing/test-library/src/main/java/org/openkilda/testing/service/grpc/GrpcServiceImpl.java
```diff
@@ -63,13 +66,15 @@ public SwitchInfoStatus getSwitchStatus(String switchAddress) {
     }
 
     @Override
+    @Retryable
```
andriidovhan, line 69: unfortunately `@Retryable` won't work with negative tests
we need to add some exclude

src-java/testing/test-library/build.gradle
```diff
@@ -20,7 +20,9 @@ dependencies {
     api 'org.springframework:spring-context'
     api 'org.springframework:spring-test'
     api 'org.springframework:spring-web'
+    api 'org.springframework.retry:spring-retry:1.3.0'
```
andriidovhan, line 23: do we need this import in this spec?

andriidovhan: please, ignore this comment
my bad

src-java/testing/test-library/src/main/java/org/openkilda/testing/service/grpc/GrpcServiceImpl.java
rtretyak: added conditions

timofei-durakov (approved):

PR #3705, opened 2020-08-31
Add design for flow loop feature

rozdy:


docs/design/flow-loop/flow-loop.md
```diff
@@ -0,0 +1,44 @@
...
+
+## API
+
+* Get existing flow loops `GET /v2/flows/loop` with optional params `flow_id` and `switch_id`. Response example:
+
+~~~
+[ 
+    {
+        "switch_id": "00:00:11:22:33:44:55:66",
+        "flow_id": "string"
```
timofei-durakov, line 18: if one switch flows are supported, you need at least port to this api, also, consider one-switch one-port flows as well, maybe it's worth having vlan_vid too

docs/design/flow-loop/flow-loop.md
```diff
@@ -0,0 +1,44 @@
+# Flow loop feature
+
+## Overview
+
+Flow loop feature designed for flow path testing. Loop provides additional flow rules on one of the terminating switch so any flow traffic is returned to switch-port where it was received. 
+Such flow has `looped=true` flag and support all flow operations. When loop removed system should restore original flow rules.
```
timofei-durakov, line 6: where you going to store this flag from db perspective, please describe

timofei-durakov (changes requested):
see comments inline

docs/design/flow-loop/flow-loop.md
```diff
@@ -0,0 +1,44 @@
...
+
+* Delete flow loop and restore regular flow rules `DELETE /v2/flows/loop/{flow_id}/{switch_id}`. 
+
+## Details
+
+From persistence perspective flow loop stored as a relation from flow to switch with type `looped on`.
+
+Flow loop create and delete operations implemented using H&S approach as a separate hubs in `flow-hs` topology. All H&S flow operations use single point of rule generation based on flow parameters. In current implementation it is `SpeakerFlowSegmentRequestBuilder` class. It should be modified to support flow loop flag and reused in flow loop operations.
+
+Additional loop rules has the same cookie as all other flow rules but higher priority. Only these rules are installed or deleted during flow loop operations.
```
timofei-durakov, line 34: it may affect our sync/validate for the rules, since the diff could be invisible for validation procedure

docs/design/flow-loop/flow-loop.md
rozdy: All one-switch and one-port flows will have different `flow_id` and the same `switch_id` so we can separate them. Why do we need extra fields like port or vlan? This API should support filtration by switch, flow and both of them.

docs/design/flow-loop/flow-loop.md
rozdy: It's described in Details section.

docs/design/flow-loop/flow-loop.md
rozdy: As it mentioned in `Additional changes` section flow validation should be changed to expect such loop rules. Flow sync now is a reroute with force flag so it will use the same `single point of rule generation`. And switch validation is check only expected cookie presence so it should work fine. Separately we need to rework switch validation to check actual flow rules.

timofei-durakov: usecase: disabling loop stucked, due to some other issues. How operator should be confident that the flows, installed on a switch are correct? since the cookies are the same, it would be much more harder to figure out and sync in case of any issues.

docs/design/flow-loop/flow-loop.md
timofei-durakov: ok, i have one switch flow on sw 1 how I enable loop only for src endpoint of this flow since the switch id is exactly the same?

docs/design/flow-loop/flow-loop.md
timofei-durakov: no details, which attributes will this looped on should include?

rozdy: Reworked from relation to additional properties to flow node.

docs/design/flow-loop/flow-loop.md
rozdy: Reworked to use separate cookie type.

timofei-durakov (approved):

PR #3707, opened 2020-09-01
More accurate handling of newrok dumps for region mappings

surabujin:
Add read-write mode check for network dumps data. Helps to decrease harm in the case of 2 FL with different roles (stats and management) in the same region.


dpoltavets (approved):

timofei-durakov (approved):

timofei-durakov:
p.s. please add couple unit tests

PR #3728, opened 2020-09-21
update port/vlan should not cause reroute

timofei-durakov:


src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/update/actions/UpdateFlowAction.java
```diff
@@ -159,6 +185,44 @@ private String getOrCreateFlowGroupId(String flowId) throws FlowProcessingExcept
...
+
+        updateEndpointOnly &= originalFlow.isAllocateProtectedPath() == targetFlow.isAllocateProtectedPath();
+        updateEndpointOnly &= originalFlow.getBandwidth() == targetFlow.getBandwidth();
+        updateEndpointOnly &= originalFlow.isIgnoreBandwidth() == targetFlow.isIgnoreBandwidth();
+
+        updateEndpointOnly &= Objects.equal(originalFlow.getMaxLatency(), targetFlow.getMaxLatency());
+        updateEndpointOnly &= Objects.equal(originalFlow.getFlowEncapsulationType(),
+                targetFlow.getFlowEncapsulationType());
+        updateEndpointOnly &= Objects.equal(originalFlow.getPathComputationStrategy(),
+                targetFlow.getPathComputationStrategy());
```
niksv, line 201: It should be fine but I'll ask just  in case. Did you checked that enabling/disabling of connected devices on flow endpoint works correct if we change connected devices flag and src_port (for example)? I'm worried about situation when we install Ingress/egress rules and adds some new rules (like connected devices) in one command.

niksv (approved):

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/update/actions/UpdateFlowAction.java
timofei-durakov: good catch, updated

rozdy (approved):

PR #3738, opened 2020-09-24
Simplify BFD session management

surabujin:
Get rid of the "conflict" concept into BFD LCM - existing tools tracking
error responses on BFD session requests are capable of correctly handle
conflicts between existing and planned BFD sessions.

New toolset to track logical port status inside BfdPortFsm where added.
It simplifies tools responsible for the correct handling of incorrectly
ordered BFD session's create/remove responses and port UP/DOWN
responses. There is no way to order these FL events(response) on FL side
so BfdPortFsm must be ready to handle them in any order.

To be able to fix failed BFD session manually (remove BFD session from
switch via switch CLI console) error handling behavior where changed.
The "session is missing" error not treated as an error for the session's
remove responses. Removing existing(from BfdPortFsm point of view) BFD
session is a mandatory cleanup step. Without this fix, BfdPortFsm can't
complete this cleanup step in cases when BFD session does not exist on
the switch.


src-java/network-topology/network-storm-topology/src/main/java/org/openkilda/wfm/topology/network/controller/bfd/BfdAction.java
```diff
@@ -29,53 +28,48 @@
...
-    protected abstract Optional<ActionResult> evaluateResult();
+    protected abstract ActionResult makeResult(BfdSessionResponse response);
 
     @Value
     static class ActionResult {
-        private boolean success;
+        boolean success;
 
         /**
          * Error core (null in case of timeout).
```
nikitamarchenko, line 58: typo?

nikitamarchenko (approved):

src-java/network-topology/network-storm-topology/src/main/java/org/openkilda/wfm/topology/network/controller/bfd/BfdAction.java
surabujin: yep (fixed)

niksv (approved):

timofei-durakov (approved):

PR #3741, opened 2020-09-25
Hex view for switch field representing cookies

timofei-durakov:


niksv (approved):

src-java/northbound-service/northbound-api/src/main/java/org/openkilda/northbound/dto/v1/switches/HexView.java
```diff
@@ -0,0 +1,32 @@
...
+
+package org.openkilda.northbound.dto.v1.switches;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public interface HexView {
+
+    /**
+     * Converts list of longs to hex.
```
rozdy, line 24: `Converts list of longs to list of hex` is more correct.

rozdy (approved):

src-java/northbound-service/northbound-api/src/main/java/org/openkilda/northbound/dto/v1/switches/RulesSyncResult.java
```diff
@@ -39,4 +40,9 @@ public RulesSyncResult(
 
         this.installedRules = installedRules;
     }
+
+    @JsonGetter("installed-rules-hex")
```
surabujin, line 44: Other fields here are underscore separated, this one is dash-separated.

surabujin (commented):
I don't like the idea to add into API something dedicated to human eyes only. API clients can and should translated responses in any shape it needs.

If it is a fix for JS numbers precision issue - we should explicitly define it and force clients to use "fixed"/"hex"  representation... I am trying to say that the goal of this change is not clear enough as for me.

src-java/northbound-service/northbound-api/src/main/java/org/openkilda/northbound/dto/v1/switches/RulesSyncResult.java
timofei-durakov: in some day at past all agreed to use dash instead of underscores for any new fields in api

timofei-durakov:
@surabujin cookie itself is something dedicated for human yes only, and the only reason decimal representation is left in api is not to break backward compatibility. usage of decimal for it from day 1 was very stupid idea

PR #3748, opened 2020-09-29
Add 'lockkeeper' tag to some tests

rtretyak:


andriidovhan:
looks like some tests(i.e. SwitchDeleteSpec, AutoRerouteV2Spec) with lockkeeper are missing in this PR

andriidovhan (approved):

rtretyak:
was only marking those that are smoke_switches at the same time. This is a quick fix. I hope to try and start lockkeeper locally and check whether it can work with local fl out of the box

PR #3777, opened 2020-10-12
OrientDB schema migration toolset

surabujin:
New docker container `db_migration` responsible for applying DB
migration for devel and for prod environments.


dpoltavets (approved):

docker/db-migration/migrate-develop.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
```
sergii-iakovenko, line 1: I'd propose to distinct migration for development / production envs with separate changelog roots (e.g. migrations/root.yaml and migrations/root-develop-env.yaml).

docker/db-migration/migrations/initial-access-management.yaml
```diff
@@ -0,0 +1,13 @@
+---
+databaseChangeLog:
+  - changeSet:
+      id: 1
```
sergii-iakovenko, line 4: I'd recommend to use descriptive ids for changesets.

docker/db-migration/migrations/initial-access-management.yaml
```diff
@@ -0,0 +1,13 @@
...
+      author: surabujin
+      changes:
+        - sql: "INSERT INTO ORole SET name = 'kilda-role', mode = 0, inheritedRole = (SELECT FROM ORole WHERE name = 'writer')"
```
sergii-iakovenko, line 7: I'd recommend to stick to the "one change per changeset" rule. https://www.liquibase.org/get-started/best-practices

docker/db-migration/migrate.sh
```diff
@@ -0,0 +1,15 @@
...
+
+cd /liquibase
+
+DB_NAME="${KILDA_ORIENTDB_DB_NAME:-kilda}"
+DB_HOST="${KILDA_ORIENTDB_HOST:-odb3.pendev}"
+
+echo "Apply main migrations set (root.yaml)"
+./liquibase --username="${KILDA_ORIENTDB_USER:-kilda}" --password="${KILDA_ORIENTDB_PASSWORD:-kilda}" \
+  --url=jdbc:orient:remote:"${DB_HOST}/${DB_NAME}" \
+  --driver=com.orientechnologies.orient.jdbc.OrientJdbcDriver --changeLogFile=migrations/root.yaml update
```
sergii-iakovenko, line 13: Let's consider tagging as the rollback strategy. See https://docs.liquibase.com/change-types/community/tag-database.html

src-java/build.gradle
```diff
@@ -223,6 +223,8 @@ task buildAndCopyArtifacts(type: Copy) {
     dependsOn floodlightBuild.task(':build')
 }
 
+task buildAnd
```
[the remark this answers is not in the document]
surabujin: It should be here at all, I have changed the approach I use to download dependencies so I have dropped all added artifacts from our gradle files.

docker/db-migration/Dockerfile
```diff
@@ -0,0 +1,78 @@
+# Copyright 2017 Telstra Open Source
```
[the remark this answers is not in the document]
surabujin: A long time ago this idea instance was installed. :) Have no idea how to configure the header for Dockerfile in it.

docker/db-migration/Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+  && echo "$LB_SHA256  liquibase-${LIQUIBASE_VERSION}.tar.gz" | sha256sum -c - \
+  && tar -xzf liquibase-${LIQUIBASE_VERSION}.tar.gz
+
+# Setup GPG
+RUN GNUPGHOME="$(mktemp -d)"
+
+# Download JDBC libraries and plugins
+
+RUN wget -O /liquibase/lib/orientdb-jdbc-3.0.34-all.jar https://repo1.maven.org/maven2/com/orientechnologies/orientdb-jdbc/3.0.34/orientdb-jdbc-3.0.34-all.jar
+RUN wget -O /liquibase/lib/liquibase-orientdb-0.3.0.jar https://dl.bintray.com/till-krullmann/tools/org/unbroken-dome/liquibase-orientdb/liquibase-orientdb/0.3.0/liquibase-orientdb-0.3.0.jar
```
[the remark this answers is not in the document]
surabujin: LiquiBase does not work without this plugin at all. It can't even create its own tables to track applied migrations. Plus I do not insist on using `sql` only changesets.

docker/db-migration/migrate-develop.sh
surabujin: Didn't get. How we will "move" devel migrations into prod migrations in this schema?

docker/db-migration/migrations/initial-access-management.yaml
surabujin: done, maybe.

docker/db-migration/migrations/initial-access-management.yaml
surabujin: I would like to keep meaningful transactions i.e. all required in one transaction should go in one changeset.

docker/db-migration/migrate.sh
surabujin: ok. There is no actual migration here, so I will try to describe it in README.

docker/db-migration/migrations/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+To start DB update by hands you need to build migration container
+```shell script
+docker-compose build db_migration
+```
+
+And execute following command (for DB on some foreign host):
+```shell script
+docker run \
+  --volume /etc/resolv.conf:/etc/resolv.conf --rm --network host \
```
timofei-durakov, line 61: why do we need resolve conf in a container?

timofei-durakov (approved):
see question inline, overall looks good to me

docker/db-migration/migrations/README.md
surabujin: To use the host's system DNS resolver - without it, the container will not be able to resolve private(accesses only via VPN) hostnames.

docker/db-migration/migrations/root.yaml
```diff
@@ -0,0 +1,8 @@
+---
+databaseChangeLog:
+  - changeSet:
+      id: root-tag
+      author: surabujin
+      changes:
+        - tagDatabase:
+            tag: history-starting-point
```
sergii-iakovenko, line 8: I was thinking about tags as a solution for automatic rollback procedure, not manual.
Tags can be aligned to docker images, so the database update procedure  performs upgrade/downgrade in one action using the same docker image. 

Here's the scenario how the migration pipeline may look like:
- db-migration images for RC1 and RC2 are built using the git commit hash as the tag.
- During RC1 deployment, the database is upgraded and tagged with the RC1 hash.
- During RC2 deployment, the database is upgraded and tagged with the RC2 hash.
- During RC1 deployment, the database is downgraded with rolling back to the RC1 hash.

sergii-iakovenko (commented):
In general looks good. But the rollback procedure requires further consideration and additional development.

docker/db-migration/migrate-develop.sh
surabujin: There is an issue with `devel` migrations set. Some subset of queries must be executed from `root` user (to create required user and roles). If I will merge all devel migrations in one set, I had to run them from root user. I would like not to do it and keep grants for devel user and prod user the same, so we can detect lack of access on PR level and not on RC or even prod level.

surabujin: There is an issue with `devel` migrations set. Some subset of queries must be executed from `root` user (to create required user and roles). If I will merge all devel migrations in one set, I had to run them from root user. I would like not to do it and keep grants for devel user and prod user the same, so we can detect lack of access on PR level and not on RC or even prod level.

docker/db-migration/migrations/root.yaml
surabujin: It is good idea, but there is no simple way to implement it. There is at least 2 big challenge:
* to perform rollback we need migrations set used during corresponding update, because rollback instructions stored inside migration's set (i.e if we make update using hash H1 we need same container/code with hash H1 to do rollback)
* we need to find a common root for releases to make rollback (i.e. we have deployed RC1 and we are going to deploy RC2, we don't know how many tags we should rollback in RC1 before starting update to RC2).

Or we can choose another way for rollbacks - use rollback-sql generation during updates, and keep it as artifacts somewhere... and use them, somehow to rollback. There is a lot of challenge here too. :)

nikitamarchenko (approved):

docker/db-migration/apply-prehistory-migrations.sh
```diff
@@ -0,0 +1,38 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+
+DB_NAME="${KILDA_ORIENTDB_DB_NAME:-kilda}"
+DB_HOST="${KILDA_ORIENTDB_HOST:-odb3.pendev}"
```
sirmax123, line 19: odb3 ?

surabujin: hm... forgot to replace :(

docker/db-migration/migrations/root.yaml
sergii-iakovenko: Ok, let's continue rollback / tagging development as a separate task.

sergii-iakovenko (approved):

PR #3779, opened 2020-10-13
adjust func test to work with hex cookies

andriidovhan:


src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/SwitchHelper.groovy
```diff
@@ -338,17 +338,30 @@ class SwitchHelper {
      * Based on that you have to use extra filter to detect these rules in missing/excess/misconfigured sections.
      */
     static void verifyRuleSectionsAreEmpty(SwitchValidationResult switchValidateInfo,
-                                           List<String> sections = ["missing", "proper", "excess", "misconfigured"]) {
+                                           List<String> sections = ["missing", "proper", "excess", "misconfigured",
+                                                                    "properHex", "excessHex", "missingHex",
+                                                                    "misconfiguredHex"]) {
```
rtretyak, line 343: This is fine I guess, but with current implementation you don't really have to add `"properHex", "excessHex", "missingHex", "misconfiguredHex"` here. You will check hex values even if you pass non-hex section and check non-hex section if you only pass hex section. So I'd suggest you either check them truly separately, or just remove them from here and leave the implementation where you check both of them automatically.

andriidovhan: extracted into a separate method

rtretyak (approved):

PR #3786, opened 2020-10-16
Added parallelism for Isl Latency Topology

niksv:


src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/topology/AbstractTopologyConfig.java
```diff
@@ -38,6 +38,9 @@
     @Key("parallelism.new")
     int getNewParallelism();
 
+    @Key("isl.latency.parallelism")
+    int getIslLatencyParallelism();
```
surabujin, line 42: What about the default value? Can topology code start correctly if it will receive 0 here?

surabujin (approved):

timofei-durakov (approved):

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/topology/AbstractTopologyConfig.java
niksv: Value will always has some value because we are getting value of this param from config

```{{ getv "/kilda_storm_isl_latency_parallelism" }}```

without value in the config Kilda wouldn't start

PR #3828, opened 2020-11-05
[Server42] Show warning if dpdk interface is down

nikitamarchenko:


vriabyk (approved):

src-cpp/server42/src/server42.cpp
```diff
@@ -314,9 +314,18 @@ int main(int argc, char *argv[]) {
...
+        // Once per second actions
         if (rte_get_timer_cycles() - last_stats_dump_tsc >= cycles_in_one_second) {
+            // Write stats to file
             save_stats_to_file(config, shared_context);
+
+            // Check interface
+            pcpp::DpdkDevice::LinkStatus link_status {};
+            device->getLinkStatus(link_status);
+            if (!link_status.linkUp) {
+                BOOST_LOG_TRIVIAL(warning) << "Link DOWN";
```
surabujin, line 326: Should we stop sending rtt packages if the interface is down? I.e. will packet send attempt out of down interface produce errors in logs?

surabujin (approved):

src-cpp/server42/src/server42.cpp
nikitamarchenko: somebody will drop packets, no error occurred

PR #3831, opened 2020-11-06
Removed legacy variable OK_TESTS="DISABLE_LOGIN"

niksv:
It was used for tests to turn off authentication during testing. Now tests use real credentials.

### DIFF
--- a/docker/base/hacks/usecase/network.disco.make
+++ b/docker/base/hacks/usecase/network.disco.make
@@ -46,7 +46,7 @@ help:
 up:
 	# NB: just specifying the minimal set .. taking advantage of embedded dependencies to bring
 	#     up the remaining containers
-	OK_TESTS="DISABLE_LOGIN" docker-compose up -d mininet floodlight storm-supervisor \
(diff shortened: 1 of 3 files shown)


docker/base/hacks/usecase/network.disco.make
```diff
@@ -46,7 +46,7 @@ help:
 up:
 	# NB: just specifying the minimal set .. taking advantage of embedded dependencies to bring
 	#     up the remaining containers
-	OK_TESTS="DISABLE_LOGIN" docker-compose up -d mininet floodlight storm-supervisor \
+	docker-compose up -d mininet floodlight storm-supervisor \
```
surabujin, line 49: We don't have mininet and topology-engine... Is this `network.disco.make` actual at al?

niksv: looks like it's not actual > 1 year :(

rtretyak (approved):

andriidovhan (approved):

PR #3844, opened 2020-11-10
Added ability to run Kilda in blue and green modes locally

niksv:


dpoltavets (approved):

confd/templates/makefile/makefile.tmpl
```diff
@@ -44,7 +48,25 @@ up-test-mode:
...
+
+up-test-mode-green:
+	cp -n .env.example .env
+	WFM_TOPOLOGIES_MODE=green docker-compose up -d floodlight_2 northbound-blue-green wfm {{if not (exists "/no_grpc_speaker")}}grpc-speaker-blue-green{{end}}
+	docker-compose logs -f wfm
+	$(MAKE) -C tools/elk-dashboards
+
+up-test-mode-blue:
+	cp -n .env.example .env
+	TAG=$(STABLE_TAG) WFM_TOPOLOGIES_MODE=blue  docker-compose up -d floodlight_1 wfm --scale northbound-blue-green=0 {{if not (exists "/no_grpc_speaker")}}--scale grpc-speaker-blue-green=0{{end}}
```
dpoltavets, line 69:
```
--scale northbound-blue-green=0 {{if not (exists "/no_grpc_speaker")}}--scale grpc-speaker-blue-green=0{{end}}
```
This is not needed here.

dpoltavets (approved):

docker/wfm/app/entry-point.sh
```diff
@@ -19,5 +19,21 @@ set -eu ${DEBUG:+-x}
...
 
 cd /app
-make kill-all
-exec make deploy-all
+
+case ${WFM_TOPOLOGIES_MODE:-} in
+
+  blue)
+    make kill-all-blue
+    exec make deploy-all-blue
```
surabujin, line 27: do not use `exec` in this way. In the original code, it was some kind of optimization - we can "close/replace" the existing bash process with make (at the end of the script). Now you have some logic around this `make` call... and should not close the existing bash process too early.

PS it can be done in a simpler way (if WFM_TOPOLOGIES_MODE can contain only "blue" or "green" or are not defined):
```
make kill-all${WFM_TOPOLOGIES_MODE:+-$WFM_TOPOLOGIES_MODE}
exec make deploy-alll${WFM_TOPOLOGIES_MODE:+-$WFM_TOPOLOGIES_MODE}
```

docker/wfm/app/entry-point.sh
```diff
@@ -19,5 +19,21 @@ set -eu ${DEBUG:+-x}
...
+    ;;
+
+  green)
+    make kill-all-green
+    exec make deploy-all-green
+    ;;
+
+  *)
+    make kill-all
```
surabujin, line 36: Unexpected behavior, at least for me. You completely ignore the value of WFM_TOPOLOGIES_MODE. Such behavior can mask incorrect values into WFM_TOPOLOGIES_MODE.

surabujin (approved):

docker/wfm/app/entry-point.sh
niksv: I left it for backward compatibility. If we do not pass WFM_TOPOLOGIES_MODE typologies will run on blue mode

confd/templates/docker-compose/docker-compose.tmpl
```diff
@@ -441,9 +446,34 @@ services:
          - northbound.pendev
     mem_limit: ${NB_MEM_LIMIT:-1073741824}
 
+  northbound-blue-green:
```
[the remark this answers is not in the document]
niksv: northbound-blue-green is always green. Maybe I should rename it.
yes we still need northbound to run Kilda as usual. Maybe it can be renamed to northbound-blue

PR #3856, opened 2020-11-17
Fix flow create FSM action retry call

surabujin:


niksv (approved):

dpoltavets (approved):

surabujin:
Fixed retry introduces incorrect behavior of flow create operation in case of errors during rules setup. And this retry implementation does not match retry implementation used in other CRUD operations. In long term, it will be completely replaced. In short term this retry is removed - https://github.com/telstra/open-kilda/pull/3875

PR #3860, opened 2020-11-18
Zero downtime version

timofei-durakov:


dpoltavets (approved):

niksv (approved):

src-java/blue-green/src/main/java/org/openkilda/bluegreen/ZkWatchDog.java
```diff
@@ -32,19 +32,28 @@
...
     private Signal signal;
 
+    @VisibleForTesting
+    protected String buildVersionPath;
+    private String buildVersion;
+
     @VisibleForTesting
     Set<LifeCycleObserver> observers = new HashSet<>();
+    @VisibleForTesting
+    Set<BuildVersionObserver> buildVersionObservers = new HashSet<>();
```
surabujin, line 48: As for me, it looks like you need some "generic" (with java generics) class responsible for publishing to a single node (with shared zk-client if it is possible) and some closure for the set of such publishers. This approach is much more scalable than the current one.

surabujin (approved):

Issue #3821, opened 2020-11-03
System is unable to create a protected flow with max_latency strategy


rtretyak:
1. Given a topology, that has at least 2 non-overlapping paths
2. Try creating a protected flow with max_latency strategy and non-zero max_latency
```
{
  "flow_id" : "protected_flow_max_lat",
  "source" : {
    "switch_id" : "00:00:00:00:00:00:00:01",
    "port_number" : 13,
    "vlan_id" : 2043
  },
  "destination" : {
    "switch_id" : "00:00:00:00:00:00:00:02",
    "port_number" : 13,
    "vlan_id" : 4014
  },
  "maximum_bandwidth" : 123,
  "max_latency": 11,
  "path_computation_strategy": "max_latency",
  "allocate_protected_path": true
}
```
**Expected:** System is able to create such flow if there are at least 2 paths that match the max_latency SLA
**Actual:** System not able to create such flow, no matter how big max_latency is specified

note: special case with `max_latency: 0` works fine
autotest that reveals the issue is available

PR #3668, opened 2020-07-30
improvements in func-test, enable server42

rtretyak:
- discovery packet ttl increased to allow bigger delay to switches
- otsdb client noisy output fixed
- enable server42 by default
- adjust functional to work with server42 rules in multiTable mode
- refactor getDefaultCookies according to #3853
- unignore functional test for 3821

PR #3881, opened 2020-11-27
Unignore functional test for #3821

rtretyak:


andriidovhan (commented):
It is already done in #3668 
I hoped it would be merged till today.

Issue #3841, opened 2020-11-10
[KildaGui]- Change kildagui build structure from maven to gradle


neeraj890kumar:
Change kildagui build structure from maven to gradle

PR #3893, opened 2020-12-01
Kilda Gui Maven to Gradle build structure change

neeraj890kumar:
- changing the build structure of kilda gui to gradle
resolves: #3841


sergii-iakovenko (changes requested):
Currently the PR contains the combined changes of "Kilda Gui Maven to Gradle build structure change" and the fix for windows newline char. We strongly encourage to have a PR per each feature / fix / improvement. Please, split the changes into 2 separate PRs.

Issue #3906, opened 2020-12-04
Discovery of one-way ISL leads to missing rules when in multitable mode


rtretyak:
Given: switches in multitable mode. A potential ISL, that can pass traffic only in 1 direction
1. Discover an ISL only in 1 direction (mandatory that it does not pass disco packets in reverse direction)
2. Observe ISL added to the system in Failed state (expected)
3. Validate switches on both ends of ISL
**Expected:** no anomalies, all rules are in 'proper' section
**Actual:** Switches have missing rules.
novi NS missing: 8020000000000007, 8030000000000007, 8040000000000007 
novi WB missing: 80200000000000d1
ovs: not reproducible

sergii-iakovenko:
The same issue affects Server42 ISL RTT rules - if a new ISL is discovered only in 1 direction, then Network topology doesn't install Server42 ISL RTT rules. Therefore the switch validation process reports missing ISL RTT rules for such ISLs.

The team discussed and agreed the possible solution:
- Introduce "resourceState" in ISL entity, and store it in the database. The state indicates whether the corresponding resources are / were installed on switches.
- The validation process takes into account "resourceState" and doesn't report missing rules if they weren't installed yet.

PR #3908, opened 2020-12-07
add test for #3906

andriidovhan:


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/links/LinkSpec.groovy
```diff
@@ -759,13 +759,74 @@ class LinkSpec extends HealthCheckSpecification {
...
-                assert islUtils.getIslInfo(links, isl).get().state == IslChangeType.DISCOVERED
+                assert islUtils.getIslInfo(links, isl.reversed).get().state == DISCOVERED
+                assert islUtils.getIslInfo(links, isl).get().state == DISCOVERED
             }
         }
         database.resetCosts()
     }
 
+    @Tidy
+    def "The src/dst switches are valid when isl is discovered in one direction only"() {
```
rtretyak, line 770: I don't like this naming since it is very narrow. Test should describe a certain situation that we are testing, more broad scenario. Right now this looks like a pin-point test that is created to test a specific defect, very low value.
Describe a real-life situation/use-case: `System detects a 1-way ISL as a Failed ISL` -> then as part of this broad test add required verifications

andriidovhan: Make sense.Renamed.

rtretyak (approved):

PR #3913, opened 2020-12-11
Moved ZkClient initialization out of constructor

niksv:


timofei-durakov (approved):

dpoltavets (approved):

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/zk/ZooKeeperBolt.java
```diff
@@ -72,14 +69,10 @@ protected void init() {
     }
 
     private void initZk() {
```
surabujin, line 71: Do you really need both `init()` and `initZk()`?

src-java/blue-green/src/main/java/org/openkilda/bluegreen/ZkWatchDog.java
```diff
@@ -61,38 +61,39 @@ public ZkWatchDog(String id, String serviceName, String connectionString,
...
-    private void checkData(String path) throws KeeperException, InterruptedException {
+    private void checkData(String path) {
         zookeeper.getData(path, this, this, null);
     }
 
     @Override
-    void initWatch() {
+    public void init() {
         try {
+            initZk();
```
surabujin, line 92: It must be the responsibility of ZkClient...

src-java/blue-green/src/main/java/org/openkilda/bluegreen/ZkWatchDog.java
```diff
@@ -61,38 +61,39 @@ public ZkWatchDog(String id, String serviceName, String connectionString,
...
-        initWatch();
     }
 
     @Override
-    void validateNodes() throws KeeperException, InterruptedException {
+    void validateNodes() throws KeeperException, InterruptedException, IOException {
         super.validateNodes();
         ensureZNode(serviceName, id, SIGNAL);
         ensureZNode(DEFAULT_BUILD_VERSION.getBytes(), serviceName, id, BUILD_VERSION);
+        nodesValidated = true;
```
surabujin, line 71: Terrible logic tangling between this class and it's parent. I can't propose anything to fix it right now... except removing this `nodesValidated` flag.

src-java/blue-green/src/main/java/org/openkilda/bluegreen/ZkClient.java
```diff
@@ -93,12 +101,18 @@ protected void ensureZNode(byte[] value, String... path) throws KeeperException,
...
+    void validateNodes() throws KeeperException, InterruptedException, IOException {
         ensureZNode(serviceName);
         ensureZNode(serviceName, id);
     }
 
+    /**
+     * Checks that connection to zookeeper is alive and nodes were validated.
+     */
     public boolean isConnectionAlive() {
+        if (zookeeper == null || !nodesValidated) {
```
surabujin, line 113: Why we have `!nodesValidated` here? Especially when `ZcClient` is not controlling this flag value?

src-java/blue-green/src/main/java/org/openkilda/bluegreen/ZkWriter.java
```diff
@@ -51,11 +48,21 @@ public void setState(int active) {
         }
     }
 
+    @Override
+    public void init() {
+        try {
+            initZk();
+            validateNodes();
```
surabujin, line 55: `validateNodes()` will never be called, if initZk() on previous line raise exception.

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/zk/ZooKeeperBolt.java
niksv: actually not. We can move initZk to init

src-java/blue-green/src/main/java/org/openkilda/bluegreen/ZkClient.java
niksv: To be ensured that nodes were validated. In ZkWatchDog and ZkClient

src-java/blue-green/src/main/java/org/openkilda/bluegreen/ZkWatchDog.java
niksv: I want to solve following problem. 
We need to check that we connected to ZK and it works fine. We need to check not one, but on every action (when we receive tupple, when we reading massage from kafka, etc)
So this check must be fast. Current check method checks
1. that we connected to zk
2. that we able to create nodes and default nodes were created

If we will check just 1 it would be useless. Because clients wouldn't able to work with Zookeeper. 
Maybe I can rename `isConnectionAlive` method to something else. like `isZooKeeperZeady ToWork`

src-java/blue-green/src/main/java/org/openkilda/bluegreen/ZkWatchDog.java
niksv: `initZk` creates connection to ZK. The whole idea of this patch is to split creating of ZkClient instance and connecting to ZK.

src-java/blue-green/src/main/java/org/openkilda/bluegreen/ZkWriter.java
niksv: And what is the problem? `nodesValidated` wouldn't be set to true? If yes it exactly what I want. To return false in `isCOnnectionAlive` in this case

PR #3924, opened 2020-12-15
Zk Connection clean up

timofei-durakov:


niksv:
We do not need it anymore

PR #3926, opened 2020-12-16
Zookeeper connections safe

timofei-durakov:


dpoltavets (approved):

niksv (approved):

src-java/blue-green/src/main/java/org/openkilda/bluegreen/ZkClient.java
```diff
@@ -141,18 +167,31 @@ protected void ensureZNode(byte[] value, String... path) throws KeeperException,
         }
     }
 
-    void validateNodes() throws KeeperException, InterruptedException, IOException {
+    @VisibleForTesting
+    void validateZNodes() throws KeeperException, InterruptedException, IOException {
         ensureZNode(serviceName);
         ensureZNode(serviceName, id);
+        validateNodes();
+        nodesValidated = true;
```
surabujin, line 175: You are making an assumption that nothing can happen with nodes while there where no connection to zk... as for me it is too optimistic.

surabujin (approved):

src-java/blue-green/src/main/java/org/openkilda/bluegreen/ZkClient.java
timofei-durakov: not exactly, if zk is reconnected the flag is reseted to false

PR #3937, opened 2020-12-20
[Server42] script for generate protobuf java files

nikitamarchenko:


src-cpp/server42/Dockerfile.protobuf
```diff
@@ -0,0 +1,14 @@
+FROM ubuntu:18.04
```
surabujin, line 1: can we have 20.04? Recently we have an issue with "missing" ubuntu repositories for outdated ubuntu release.

src-cpp/server42/Dockerfile.protobuf
```diff
@@ -0,0 +1,14 @@
...
+
+WORKDIR /root
+
+RUN echo 'deb http://archive.ubuntu.com/ubuntu/ bionic-proposed main' > /etc/apt/sources.list.d/bionic-proposed.list \
+  && apt-get update -y && apt-get install -y curl unzip
+
+RUN curl -L https://github.com/protocolbuffers/protobuf/releases/download/v3.8.0/protoc-3.8.0-linux-x86_64.zip --output protobuf.zip
```
surabujin, line 8: Why not use `wget` (it should be available in the base image)?

src-java/server42/server42-stats-messaging/src/main/java/org/openkilda/server42/stats/messaging/flowrtt/Statistics.java
```diff
@@ -1,5 +1,5 @@
 // Generated by the protocol buffer compiler.  DO NOT EDIT!
-// source: server42/statistics.proto
+// source: statistics.proto
```
surabujin, line 2: Should it be removed from repo now?

surabujin (approved):

src-cpp/server42/Dockerfile.protobuf
niksv: +1

niksv: +1

niksv: agree

niksv: agree

sirmax123 (approved):

src-cpp/server42/generate_java_protobuf.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+if [ ! -f "src-java" ]; then
```
sergii-iakovenko, line 3: I think it should be "-d" instead of "-f".

src-cpp/server42/Dockerfile.protobuf
nikitamarchenko: `/bin/sh: 1: wget: not found`

src-cpp/server42/Dockerfile.protobuf
nikitamarchenko: done

src-java/server42/server42-stats-messaging/src/main/java/org/openkilda/server42/stats/messaging/flowrtt/Statistics.java
nikitamarchenko: We need to discuss that. I will do it with a separate PR if somebody needs that.

src-cpp/server42/generate_java_protobuf.sh
nikitamarchenko: good catch

Issue #3941, opened 2020-12-21
Retry during reroute after table mode change leads to excess rules


rtretyak:
Case looks similar to #3940

1. A flow in single-table mode, on single-table mode switches
2. Src switch mode is changed to multi-table (flow remains in single-table mode until next reroute)
3. Current path breaks (ISL down) and reroute happens
4. During reroute, system for some reason is unable to install rules on new path from the first try, start retries and one of the retries is successful 
5. Run switch validation for related switches
**Expected:** no rule discrepancies on switches
**Actual:** Src switch has some excess rules
```
"excess": [
            2305843009213795558,
            4611686018427489510
        ]
```
- Also in Kibana errors are present, similar to #3940: `segment and flow level multi-table flag values are incompatible to each other - flow(true) != segment(false)"`, but flow can be successfully deleted, unlike 3940

rtretyak:
autotest available

rtretyak:
Closed by #3946

Issue #3940, opened 2020-12-21
Flow segment and flow level multitable flag may diverge during reroute, leading to multiple issues


rtretyak:
There is a problem with flow-level 'multitable' state and flow-segment-level state becoming diverge under certain circumstances.

1. A flow in single-table mode, on single-table mode switches
2. Src switch mode is changed to multi-table (flow remains in single-table mode until next reroute)
3. Current path breaks (ISL down) and reroute happens
4. During reroute, system for some reason is unable to install rules on new path and after a number of retries brings the flow down
5. ISL on current path goes back UP, resulting in a new flow reroute
6. Flow lands on the original path and has an UP state
7. Run flow validation, run switch validation for related switches
**Expected:** Flow validation is OK, no rule discrepancies on switches
**Actual:**
- Flow validation show corrupted flow rules on src switch:
```
 "discrepancies": [
            {
                "rule": "{sw:00:00:00:00:00:00:00:01, ck:4611686018427489505, in:20-0, out:1-2202}",
                "field": "inVlan",
                "expected_value": "0",
                "actual_value": "2029"
            }
        ]
```
_^flow endpoint uses 2029 vlan, validation seems to be incorrectly expecting no vlan_

- Switch validation show missing rules on src switch
```
"rules": {
        "missing": [
            -9200854038717923308,
            36028797151936532
        ]
```
_-9200854038717923308 type MULTI_TABLE_INGRESS_RULES
36028797151936532 type  SHARED_OF_FLOW_

- Flow cannot be removed, error `"details": "RemoveRulesAction failed: First flow(id:21Dec140836_187_wholewheatflour5793, path:21Dec140836_187_wholewheatflour5793_56c0e090-a48a-4713-bd8b-fb385de8cc02) segment and flow level multi-table flag values are incompatible to each other - flow(true) != segment(false)"`
- Same errors with 'incompatible flag' in kibana during second reroute

rtretyak:
autotest available

PR #3942, opened 2020-12-21
Add func tests for #3940 and #3941

rtretyak:


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/resilience/RetriesSpec.groovy
```diff
@@ -446,4 +448,170 @@ and at least 1 path must remain safe"
...
+        antiflap.portDown(isl.srcSwitch.dpId, isl.srcPort)
+
+        then: "At least 2 reroute attempts happen, but more retries are about to go"
+        wait(WAIT_OFFSET * 2) {
+            //wait for 2, max is 4
+            assert northbound.getFlowHistory(flow.flowId).count { it.action == REROUTE_ACTION } == 2
+            northboundV2.getFlowStatus(flow.flowId).status == FlowState.DOWN
+        }
+
+        when: "Switch one the new path becomes responsive and allows rule installation"
```
andriidovhan, line 589: The same as in the previous test
Shouldn't we do it in a legal way?
I mean, set the real status for the switch and then invoke the `reviveSwitch` method

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/resilience/RetriesSpec.groovy
```diff
@@ -446,4 +448,170 @@ and at least 1 path must remain safe"
...
+        currentSwitches.each {
+            def validation = northbound.validateSwitch(it.dpId)
+            validation.verifyRuleSectionsAreEmpty(["excess", "missing", "misconfigured"])
+            validation.verifyMeterSectionsAreEmpty(["excess", "missing", "misconfigured"])
+        }
+        def done = true
+
+        cleanup:
+        flow && flowHelperV2.deleteFlow(flow.flowId)
+        swDown && switchHelper.reviveSwitch(brokenSwitch, blockData)
```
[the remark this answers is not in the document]
rtretyak: 1. done
2. I sometimes receive the 'flow is in progress' error when deleting the flow after reviving switch, so I intentionally did it this way

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/resilience/RetriesSpec.groovy
rtretyak: I can agree with the previous situation where we do this in cleanup, but right here the whole situation is kinda illegal. We simulate the case when switch is active but we can't install rules on it - this is already illegal. And right here we simulate that one of the retries suddenly succeed. Switch should not officially go 'offline' in this particular scenario. It just became responsive again in the middle of retries. Making it officially offline may ruin the whole point of test

andriidovhan (approved):

PR #3943, opened 2020-12-22
Minor fixes in zookeeper event handling.

dpoltavets:


niksv (approved):

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/AbstractBolt.java
```diff
@@ -127,16 +127,30 @@ protected void dispatch(Tuple input) throws Exception {
 
     protected void handleLifeCycleEvent(LifecycleEvent event) {
         if (Signal.START.equals(event.getSignal())) {
-            active = true;
             emit(ZkStreams.ZK.toString(), currentTuple, new Values(event, commandContext));
+            try {
+                activate();
+            } finally {
+                active = true;
```
surabujin, line 134: I would like to alter `active` value inside `activate()` / `deactivate()` methods, in this case successors can alter behaviour, if required.

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/AbstractBolt.java
```diff
@@ -127,16 +127,30 @@ protected void dispatch(Tuple input) throws Exception {
...
+            } finally {
+                active = false;
+            }
         } else {
             log.error("Unsupported signal received: {}", event.getSignal());
         }
     }
 
+    protected void activate() {
+    }
```
surabujin, line 149: add commend "no actions required" or something similar. Now it looks like someone has forgotten to write the method body.

surabujin (approved):

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/AbstractBolt.java
dpoltavets: Good point, but there is no such case yet. For now, we want to always update this flag.

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/AbstractBolt.java
dpoltavets: Added.

PR #3964, opened 2020-12-30
add test for #3938

andriidovhan:


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowHistoryV2Spec.groovy
```diff
@@ -298,6 +306,42 @@ class FlowHistoryV2Spec extends HealthCheckSpecification {
         ]
     }
 
+    @Tidy
+    @Tags(LOW_PRIORITY)
+    def "Info about a terminating switch is registered in flow history while rerouting"() {
```
rtretyak, line 311: This is a bit misleading description. We should not really try to reroute a flow when terminating switch is down. We just change its status to DOWN without attempting to find a new path. Technically it's a part of 'reroute' process, but practically we are not even trying to find a new 'route'

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowHistoryV2Spec.groovy
```diff
@@ -298,6 +306,42 @@ class FlowHistoryV2Spec extends HealthCheckSpecification {
...
+        flowHelperV2.addFlow(flow)
+
+        when: "Deactivate the src switch"
+        def blockData = lockKeeper.knockoutSwitch(srcSwitch, RW)
+        def swIsActive = false
+        Wrappers.wait(WAIT_OFFSET) {
+            assert northbound.getSwitch(srcSwitch.dpId).state == SwitchChangeType.DEACTIVATED
+        }
+
+        then: "Flow goes DOWN"
```
rtretyak, line 324:
```
when: "Deactivate the src switch"
then: "Flow goes DOWN"
```
This currently looks like the fact of switch deactivation has brought the flow down, which is not true. Only ISL failure can change the flow status

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowHistoryV2Spec.groovy
```diff
@@ -298,6 +306,42 @@ class FlowHistoryV2Spec extends HealthCheckSpecification {
...
+        and: "The 'Switch is not active' message is registered in flow history"
+        def flowHistory = northbound.getFlowHistory(flow.flowId).find { it.action == REROUTE_ACTION }
+        assert flowHistory.payload[0].action == "Started flow validation"
+        assert flowHistory.payload[1].action == "ValidateFlowAction failed: Flow's $flow.flowId src switch is not active"
+        assert flowHistory.payload[2].action == REROUTE_FAIL
+
+        cleanup:
+        if (!swIsActive) {
+            lockKeeper.reviveSwitch(srcSwitch, blockData)
+            Wrappers.wait(WAIT_OFFSET) { northbound.getSwitch(srcSwitch.dpId).state == SwitchChangeType.ACTIVATED }
```
rtretyak, line 340: It's better to use `switchHelper.reviveSwitch` instead

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowHistoryV2Spec.groovy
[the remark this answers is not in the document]
andriidovhan: good catch
I set the `VIRTUAL` tag.

andriidovhan: refactored
```
when: "Deactivate the src switch"
and: "Related ISLs are FAILED"
then: "Flow goes DOWN"
```

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowHistoryV2Spec.groovy
andriidovhan: fixed

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowHistoryV2Spec.groovy
```diff
@@ -298,6 +307,46 @@ class FlowHistoryV2Spec extends HealthCheckSpecification {
         ]
     }
 
+    @Tidy
+    @Tags([VIRTUAL, LOW_PRIORITY])
+    //it doesn't work on the hardware env due to the 'round trip latency' feature
```
rtretyak, line 312: It will work as it is now, but it's only because of our current setup.
1. You can explicitly pick switches that do not support rtl, so that this behavior is strictly defined
2. or You can force a flow reroute by bringing some port down on the path. This way you will no longer be dependent on whether isls are rtl or not

andriidovhan: agree, refactored.
the first option was implemented

rtretyak (approved):

PR #3975, opened 2021-01-12
Prevention of decreasing the zk status below zero.

dpoltavets:


niksv (approved):

timofei-durakov (approved):

src-java/blue-green/src/main/java/org/openkilda/bluegreen/ZkStateTracker.java
```diff
@@ -62,14 +62,14 @@ private void handleStart(LifecycleEvent event) {
     private void handleShutdown(LifecycleEvent event) {
         if (shutdownUuid != null) {
             if (shutdownUuid.equals(event.getUuid())) {
-                active--;
+                active = active > 0 ? --active : 0;
```
sergii-iakovenko, line 65: Why not just
```
if (active > 0) {
    --active;
}
```

PR #3980, opened 2021-01-13
Update readmes with more relevant information

rtretyak:


README.md
```diff
@@ -67,6 +67,12 @@ wget https://github.com/kelseyhightower/confd/releases/download/v0.16.0/confd-0.
 chmod +x /usr/local/bin/confd
 ```
 
+#### /etc/hosts
+Following entry has to be added to /etc/hosts for _local_ Kilda to work properly
+```
+127.0.0.1	localhost	kafka.pendev logstash.pendev
```
andriidovhan, line 73: Please use `tab` or `space` only, not both of them at the same time

andriidovhan (approved):

PR #3987, opened 2021-01-14
Fix second create flow loop request removes loop rules

### LINKED ISSUE #3960: Flow loops: Installing the exact same loop twice leads to deletion of flow loop rules
1. Create a flow
2. Create a flow loop on src
3. Repeat the same request from previous step
**Expected:** Flow loop rules are just re-installed
**Actual:** Flow loop rules are removed. Switch validation shows missing rules

test: `"Attempt to create the exact same flowLoop twice just re-installs the rules"`

rozdy:
Close #3960

### DIFF
--- a/src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/update/actions/UpdateFlowAction.java
+++ b/src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/update/actions/UpdateFlowAction.java
@@ -234,10 +234,10 @@ public class UpdateFlowAction extends NbTrackableAction<FlowUpdateFsm, State, Ev
     }
 
     private FlowLoopOperation detectFlowLoopOperation(RequestedFlow originalFlow, RequestedFlow targetFlow) {
-        if (originalFlow.getLoopSwitchId() == targetFlow.getLoopSwitchId()) {
+        if (originalFlow.getLoopSwitchId() == null && targetFlow.getLoopSwitchId() == null) {
             return FlowLoopOperation.NONE;
         }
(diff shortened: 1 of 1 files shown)


src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/update/actions/UpdateFlowAction.java
```diff
@@ -234,10 +234,10 @@ private String getOrCreateFlowGroupId(String flowId) throws FlowProcessingExcept
     }
 
     private FlowLoopOperation detectFlowLoopOperation(RequestedFlow originalFlow, RequestedFlow targetFlow) {
-        if (originalFlow.getLoopSwitchId() == targetFlow.getLoopSwitchId()) {
+        if (originalFlow.getLoopSwitchId() == null && targetFlow.getLoopSwitchId() == null) {
```
niksv, line 237: what if `originalFlow.getLoopSwitchId() == targetFlow.getLoopSwitchId()` 
will we create same second rule in this case? Why we shouldn't return `NONE`?

rozdy: Requirements here was to reinstall loop rules when we have second create flow loop request with the same params. To achieve this goal when we detect request to create flow loop for already looped flow with the same switchId we need to process `CREATE` operation.  `FlowLoopOperation.NONE` is used for endpoint update only (when changed port/vlan) and we need to install new rules and delete old ones.

niksv (approved):

dpoltavets (approved):

sergii-iakovenko (commented):
Can we have a unit test to cover the case?

PR #3988, opened 2021-01-15
Restore topologies parallelism for local run

niksv:
PRs #3984 #3986 set topologies parallelism > 1 as a hot fix.
This PR restores parallelism back.


niksv:
parallelism will be restored in #4010

PR #3992, opened 2021-01-18
Enable checkstyle run before unit tests in make unit

timofei-durakov:


niksv (approved):

confd/templates/makefile/makefile.tmpl
```diff
@@ -102,7 +102,7 @@ compile-common: check-java-version
 
 .PHONY: unit
 unit: update-props
-	cd src-java && ./gradlew test --stacktrace
+	cd src-java && ./gradlew checkstyleMain checkstyleTest test --stacktrace
```
sergii-iakovenko, line 105: I'd prefer a dedicated target (make's target) to run checkstyle. Similar to the sonar target. This will keep flexibility, and not slow down the unit target.

timofei-durakov: as you wish, updated

niksv (approved):

sergii-iakovenko (approved):

PR #4061, opened 2021-02-12
Improve performance of H&S flow operations: back-pressure on FLOW spout

sergii-iakovenko:
The changes:
- Introduce back-pressure on FLOW spout. This addresses the issue with flooded hub bolts and slow processing of speaker responses.

### DIFF
--- a/src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/topology/AbstractTopology.java
+++ b/src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/topology/AbstractTopology.java
@@ -54,6 +54,7 @@ import org.apache.storm.Config;
 import org.apache.storm.LocalCluster;
 import org.apache.storm.StormSubmitter;
 import org.apache.storm.flux.model.BoltDef;
+import org.apache.storm.flux.model.PropertyDef;
 import org.apache.storm.flux.model.SpoutDef;
 import org.apache.storm.flux.model.TopologyDef;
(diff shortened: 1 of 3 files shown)


dpoltavets (approved):

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/topology/AbstractTopology.java
```diff
@@ -285,13 +286,22 @@ protected SpoutDeclarer declareKafkaSpout(TopologyBuilder builder, List<String>
...
         if (topologyDef != null) {
             SpoutDef spoutDef = topologyDef.getSpoutDef(spoutId);
             if (spoutDef != null) {
                 spoutParallelism = spoutDef.getParallelism();
                 if (spoutDef.getNumTasks() > 0) {
                     spoutNumTasks = spoutDef.getNumTasks();
                 }
+                if (spoutDef.getProperties() != null) {
+                    for (PropertyDef propertyDef : spoutDef.getProperties()) {
+                        if (propertyDef.getName().equals("max.spout.pending")) {
```
niksv, line 299: maybe `max.spout.pending` should be constant?
what if property is unset? should we use
`"max.spout.pending".equals(propertyDef.getName())` to handle NPE?

sergii-iakovenko: It could be a constant. Changed.

timofei-durakov (approved):

PR #4120, opened 2021-03-12
Upgrade OrientDB docker image and OrientDB client to 3.1 

sergii-iakovenko:
The changes:
- Upgrade OrientDB docker image and OrientDB client to 3.1.x
- Remove the rest of neo4j stuff.


sergii-iakovenko:
This one is blocked by heavy migration of history data (can't make orientdb upgrade in the allocated time window). The work should be resumed once the history data is moved from orientdb.

PR #4147, opened 2021-03-23
add test for 'network/paths' endpoint using LATENCY strategy

andriidovhan:


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/network/PathsSpec.groovy
```diff
@@ -41,11 +43,43 @@ class PathsSpec extends HealthCheckSpecification {
...
+    def "Able to get paths between switches for the LATENCY strategy"() {
+        given: "Two active not neighboring switches"
+        def switchPair = topologyHelper.getNotNeighboringSwitchPair()
+
+        and: "Create a flow to reduce available bandwidth on some path between these two switches"
+        def flow = flowHelperV2.addFlow(flowHelperV2.randomFlow(switchPair))
+
+        when: "Get paths between switches using the LATENCY strategy"
+        def paths = northbound.getPaths(switchPair.src.dpId, switchPair.dst.dpId, null,
+                PathComputationStrategy.LATENCY, 10, null)
```
rtretyak, line 61: I'm not sure if this is on purpose. Why are you specifying maxLatency value when querying 'latency' strategy? That value must not influence the result

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/network/PathsSpec.groovy
```diff
@@ -41,11 +43,43 @@ class PathsSpec extends HealthCheckSpecification {
...
+
+        then: "Paths will be sorted by bandwidth (descending order) and then by latency (ascending order)"
```
rtretyak, line 63: Since we are querying a 'latency' strategy, I would expect the results to be sorted by latency first. was this discussed?

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/network/PathsSpec.groovy
andriidovhan: just in case. The idea was to check that system return some result even when the `max_latency` field is set

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/network/PathsSpec.groovy
andriidovhan: it was refactored.
from now on it is sorted by the `latency` field(ascending order) and then by the `bandwidth` field (descending order).

rtretyak (approved):

PR #4148, opened 2021-03-23
Fix deserialization error in flow-monitoring on ISL delete event

rozdy:


niksv (approved):

sergii-iakovenko (approved):

src-java/network-topology/network-storm-topology/src/main/java/org/openkilda/wfm/topology/network/storm/bolt/isl/IslHandler.java
```diff
@@ -219,17 +220,18 @@ public void islRemovedNotification(Endpoint srcEndpoint, IslReference reference)
...
                         .portNumber(src.map(Endpoint::getPortNumber).orElse(null))
-                        .build())
-                .destination(NetworkEndpoint.builder()
-                        .datapath(dst.map(Endpoint::getDatapath).orElse(null))
-                        .portNumber(dst.map(Endpoint::getPortNumber).orElse(null))
-                        .build())
-                .build();
-        emit(STREAM_FLOW_MONITORING_ID, getCurrentTuple(), makeIslChangedTuple(islChangedInfoData));
+                        .build());
+        if (dst.isPresent()) {
```
surabujin, line 228: How it can be? Isl can't operate on incomplete IslReference, so it can't (must not) appear here.

src-java/network-topology/network-storm-topology/src/main/java/org/openkilda/wfm/topology/network/storm/bolt/isl/IslHandler.java
```diff
@@ -219,17 +220,18 @@ public void islRemovedNotification(Endpoint srcEndpoint, IslReference reference)
...
+            islChangedInfoData.destination(NetworkEndpoint.builder()
+                    .datapath(dst.map(Endpoint::getDatapath).orElse(null))
+                    .portNumber(dst.map(Endpoint::getPortNumber).orElse(null))
```
surabujin, line 231: we need more lambdas... :(

src-java/network-topology/network-storm-topology/src/main/java/org/openkilda/wfm/topology/network/storm/bolt/isl/IslHandler.java
rozdy: It's special case when IslReference with only one side created to notify flow monitoring about isl delete.

src-java/network-topology/network-storm-topology/src/main/java/org/openkilda/wfm/topology/network/storm/bolt/isl/IslHandler.java
rozdy: Reworked.

src-java/network-topology/network-storm-topology/src/main/java/org/openkilda/wfm/topology/network/storm/bolt/isl/IslHandler.java
surabujin: Why not to fill both ends?

src-java/network-topology/network-storm-topology/src/main/java/org/openkilda/wfm/topology/network/storm/bolt/isl/IslHandler.java
surabujin: why not just `if (reference.getDest() == null) { doSomething(); }`? Just curious... do not understand such lambdas popularity in java.

src-java/network-topology/network-storm-topology/src/main/java/org/openkilda/wfm/topology/network/storm/bolt/isl/IslHandler.java
surabujin: So you make a special treatment of `IslChangedInfoData` with only source part filled? As for me must be used new / separate event, to notify about ISL removal instead of such weird unexpected logic built in into existing object even not into object, but in some handler processing this object.

rozdy: I used this way to use only one message type (`IslChangedInfoData`) to flow-monitoring topology about any ISL changes (moved or deleted). I will try to rework it using separate message type.

src-java/network-topology/network-storm-topology/src/main/java/org/openkilda/wfm/topology/network/storm/bolt/isl/IslHandler.java
rozdy: As for me lambdas and other functional-style code in java is more human-readable so I like to use them when I can.

nikitamarchenko (approved):

PR #4197, opened 2021-04-06
Updated version of switch fsm to fix bfd

timofei-durakov:


niksv (approved):

dpoltavets (approved):

docs/design/network-discovery/switch-FSM.puml
```diff
@@ -32,7 +32,7 @@ INIT --> SYNC : online
 INIT --> PENDING : history
 
 PENDING : enter / setup port workers
-PENDING --> ONLINE : online / init ports from history
+PENDING --> SYNC : online / init ports from history
```
surabujin, line 35: There are one side effect here - we are delaying ISL FSM's creation to the moment of first network dump. Can't say how it will affect whole behaviour (we increase a window when we can't track ISL health during deployment).

If "init ports from history" will be done on enter to the PENDING state we will get rid from this delay, but again can't say how it will affect whole behaviour.

surabujin (approved):

PR #4203, opened 2021-04-09
Cleanup deserialization error handling reporting

surabujin:


rozdy (approved):

nikitamarchenko (approved):

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/kafka/MessageDeserializer.java
```diff
@@ -42,15 +38,15 @@ public void configure(Map<String, ?> configs, boolean isKey) {
...
         } catch (IOException e) {
-            String message = StringUtils.toEncodedString(data, Charset.defaultCharset());
-            if (log.isDebugEnabled()) {
-                log.debug(format("Failed to deserialize data: %s from topic %s", message, topic), e);
-            }
-            ErrorData errorData = new ErrorData(ErrorType.INTERNAL_ERROR, "Failed to deserialize message", message);
-            return new DeserializationErrorMessage(errorData, System.currentTimeMillis(), UUID.randomUUID().toString());
+            DeserializationErrorReport errorReport = DeserializationErrorReport.createFromException(
+                    topic, base, data, e);
+            return new ErrorMessage(
```
niksv, line 47: we will lose information about deserialization error without log because a lot of consumers (like `DatapointParseBolt` or `SwitchManagerHub`) doesn't handle Error messages at all. In such cases Operator will see only `ClassSastException` in logs. We need to rework it if we want to remove logging from here

niksv (changes requested):

PR #4206, opened 2021-04-13
Use the same consumer group for both northbound colors

timofei-durakov:
this patch removes color for nb in different envs, since it's a
singleton anyway. So it's reduce operations part on cleaning group
offset

### DIFF
--- a/confd/templates/docker-compose/docker-compose.tmpl
+++ b/confd/templates/docker-compose/docker-compose.tmpl
@@ -489,7 +489,7 @@ services:
       context: docker
       dockerfile: northbound/Dockerfile
     ports:
-      - "8787:8080"
+      - "8080:8080"
     environment:
       REST_USERNAME: 'kilda'
       REST_PASSWORD: 'kilda'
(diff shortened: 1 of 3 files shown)


dpoltavets (approved):

niksv (approved):

dpoltavets (approved):

niksv (approved):

confd/templates/docker-compose/docker-compose.tmpl
```diff
@@ -489,7 +489,7 @@ services:
       context: docker
       dockerfile: northbound/Dockerfile
     ports:
-      - "8787:8080"
+      - "8080:8080"
```
surabujin, line 492: You have same value for "northbound" service (line 467). I will collide.

timofei-durakov: yes, but it won't this is made intentionally to replicate production usecase, check makefile templates

surabujin: I saw them. And didn't get why for we keep northbound, if it is "overriden" by northbound-blue-green ?

PR #4201, opened 2021-04-08
Rework LATENCY path computation strategy

rozdy:
Add support of maxLatency and maxLatencyTier2 params to LATENCY path computation strategy. Flow operations should move flow to DEGRADED state If found path has latency more than maxLatency and to DOWN state if found path has latency more than maxLatencyTier2.

Related to #3968

PR #4208, opened 2021-04-15
add test for a flow with latency according to #4201

andriidovhan:


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/MaxLatencySpec.groovy
```diff
@@ -242,6 +242,58 @@ class MaxLatencySpec extends HealthCheckSpecification {
         database.resetCosts()
     }
 
+    @Tidy
+    def "Able to create DEGRADED flow with LATENCY strategy if flowPath > max_latency"() {
+        given: "2 non-overlapping paths with 11 and 15 latency"
+        setLatencyForPaths(11, 15)
+
+        when: "Create a flow, maxLatency 5 and maxLatencyTier2 10"
```
rtretyak, line 250: description not related to context

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/MaxLatencySpec.groovy
```diff
@@ -242,6 +242,58 @@ class MaxLatencySpec extends HealthCheckSpecification {
...
+        def flow = flowHelperV2.randomFlow(switchPair).tap {
+            allocateProtectedPath = false
+            maxLatency = 9
```
rtretyak, line 253: can be 10?

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/MaxLatencySpec.groovy
```diff
@@ -242,6 +242,58 @@ class MaxLatencySpec extends HealthCheckSpecification {
...
+            maxLatencyTier2 = 16
+            pathComputationStrategy = PathComputationStrategy.LATENCY.toString()
+        }
+        northboundV2.addFlow(flow)
+
+        then: "Flow is created in DEGRADED state because flowPath doesn't satisfy max_latency value"
```
rtretyak, line 259: please add `but satisfies max_latency_tier2`, would make the intended behavior very clear

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/MaxLatencySpec.groovy
```diff
@@ -242,6 +242,58 @@ class MaxLatencySpec extends HealthCheckSpecification {
...
+    @Tidy
+    def "Able to create DEGRADED flow with LATENCY strategy if flowPath > max_latency"() {
+        given: "2 non-overlapping paths with 11 and 15 latency"
+        setLatencyForPaths(11, 15)
+
+        when: "Create a flow, maxLatency 5 and maxLatencyTier2 10"
+        def flow = flowHelperV2.randomFlow(switchPair).tap {
+            allocateProtectedPath = false
+            maxLatency = 9
+            maxLatencyTier2 = 16
```
rtretyak, line 254: can be 12? I don't follow the chosen boundary values. Can you explain?

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/MaxLatencySpec.groovy
```diff
@@ -242,6 +242,58 @@ class MaxLatencySpec extends HealthCheckSpecification {
...
+        flow && flowHelperV2.deleteFlow(flow.flowId)
+    }
+
+    @Tidy
+    @Tags([LOW_PRIORITY])
+    def "Able to create a flow with LATENCY strategy when max_latency > pathLatency > max_latency_tier2"() {
+        given: "2 non-overlapping paths with 11 and 15 latency"
+        setLatencyForPaths(11, 15)
+
+        when: "Create a flow, maxLatency 5 and maxLatencyTier2 10"
```
rtretyak, line 279: same

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/MaxLatencySpec.groovy
```diff
@@ -242,6 +242,58 @@ class MaxLatencySpec extends HealthCheckSpecification {
...
+        def flow = flowHelperV2.randomFlow(switchPair).tap {
+            allocateProtectedPath = false
+            maxLatency = 19
```
rtretyak, line 282: 12?

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/MaxLatencySpec.groovy
```diff
@@ -242,6 +242,58 @@ class MaxLatencySpec extends HealthCheckSpecification {
...
+            maxLatencyTier2 = 8
```
rtretyak, line 283: 10?

rtretyak (changes requested):

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/MaxLatencySpec.groovy
andriidovhan: There was an issue when we were not able to create a flow in case `max_latency > max_latency_tier2`
This test was added just to cover this case.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/MaxLatencySpec.groovy
rtretyak: I still don't get it. This should work the same with `maxLatencyTier2 = 12`. Second path with '15' latency is not even used in this test and we are solely focused on the '11' latency path. So if should not satisfy the `maxLatency=10` restriction but should satisfy the tier2 restriction. Is it the case being tested here? If yes, then `16` value completely obfuscates the goal of the test and messes with the second '15' path which is not used in the test

andriidovhan: Agree, it is not clear. I didn't think from this side.
The system always tries to find a path with the best latency for a flow with the `latency` strategy.
The `max_latency` and `max_latency_tier2` fields are used for setting the status of the flow.

for example:
We have two paths with 11 and 15 latency
also we have two flows (flow_1: mL=10, mlT2=12; flow_2: mL=14, mlT2=16)
Both flows are build via the path with 11 latency.
But the status of the flows are different: flow_1 - degraded, flow_2-up


The `maxLatencyTier2` value in the test was changed to 12, now it should clear.

rtretyak: Good example. The code also looks clear now 👍

rtretyak (approved):
Maybe not related to this particular PR, but I'm now interested in how the system will behave in cases:
1. Latency value on the flow is the same as latency value on the path
2. ml = mlT2 = path

andriidovhan:
> Maybe not related to this particular PR, but I'm now interested in how the system will behave in cases:
> 
>     1. Latency value on the flow is the same as latency value on the path
> 
>     2. ml = mlT2 = path

To tell the truth, I don't know.
I will check it.

andriidovhan:
> > Maybe not related to this particular PR, but I'm now interested in how the system will behave in cases:
> > ```
> > 1. Latency value on the flow is the same as latency value on the path
> > 
> > 2. ml = mlT2 = path
> > ```
 In these cases, the flow will be built via a path that has the same latency as ml and the flow will be in UP state.

PR #4269, opened 2021-05-25
Extend traffgen api to return address stats

rtretyak:


andriidovhan (approved):

surabujin (approved):
Please increase traffexam version (any change in it must increase it's version).

PR #4281, opened 2021-06-01
Sync transit rules via SpeackerCommand

niksv:


dpoltavets (approved):

src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/command/flow/NotIngressFlowSegmentCommand.java
```diff
@@ -15,6 +15,8 @@
 
 package org.openkilda.floodlight.command.flow;
 
+import static org.openkilda.floodlight.switchmanager.SwitchManager.VXLAN_UDP_DST;
```
sergii-iakovenko, line 18: Looks weird - accessing a "service" class (which is higher in the hierarchy) from a shared "action" class (which is lower).

sergii-iakovenko (commented):
Looks good.

rozdy (approved):

PR #4283, opened 2021-06-02
Server 42: fix server stub update based on flow hash

nikitamarchenko:
no production code affected

### DIFF
--- a/src-java/server42/server42-control-server-stub/src/main/java/org/openkilda/server42/control/serverstub/StatsServer.java
+++ b/src-java/server42/server42-control-server-stub/src/main/java/org/openkilda/server42/control/serverstub/StatsServer.java
@@ -47,6 +47,7 @@ public class StatsServer extends Thread {
         String flowId;
         Boolean direction;
         Long baseLatency;
+        long flowHashCode;
     }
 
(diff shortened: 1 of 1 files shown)


dpoltavets (approved):

src-java/server42/server42-control-server-stub/src/main/java/org/openkilda/server42/control/serverstub/StatsServer.java
```diff
@@ -114,15 +115,23 @@ void init() {
      */
     public synchronized void addFlow(Flow flow) {
 
+        FlowKey flowKey = FlowKey.fromFlow(flow);
+
+        if (flows.containsKey(flowKey) && flows.get(flowKey).getFlowHashCode() == flow.getHashCode()) {
```
sergii-iakovenko, line 120: Need to be careful with == operator. The operator works here only because flow.getHashCode() returns a primitive. If it returns wrapper class as getFlowHashCode() does, then the operator will fail.

sergii-iakovenko (approved):

timofei-durakov (approved):

PR #4319, opened 2021-06-17
Set NB worker topology parallelism to 2

niksv:


niksv:
Will fix later. some day..

PR #4371, opened 2021-07-13
Add design for port LAGs

rozdy:


docs/design/LAG-for-ports/README.md
```diff
@@ -0,0 +1,50 @@
...
+}
+~~~
+
+* Delete LAG logical port on the switch `DELETE /v2/switches/{switch_id}/lags/{logical_port_number}`.
+
+
+## Details
+All logical port related commands are sent to the switches using gRPC speaker.
+
+Open-kilda calculate logical port number using the following rule: 2000 + min of the physical ports number in the LAG. It is not allowed to have one physical port in two LAGs so this rule will provide unique logical port number for any port configuration.
```
niksv, line 40: I'm not sure about unique port numbers. 
exaple:
[1, 2] ports will have logical port number 2001
[2, 3] ports will have logical port number 2002

logical ports are unique but both ports use port 2.

docs/design/LAG-for-ports/domain-model.puml
```diff
@@ -0,0 +1,23 @@
+@startuml
+
+title LAGs Domain Model
+
+class LogicalPort {
```
niksv, line 5: I'm not sure about name `LogicalPort` because we have several types of Logical ports (for example BFD port is logical)
Maybe add `Lag` in name? `LagLogicalPort` or `LagPort`?

docs/design/LAG-for-ports/README.md
rozdy: The same physical port can't be a part of two LAG ports. So this configuration is not valid. Added a validation step to the description.

docs/design/LAG-for-ports/domain-model.puml
rozdy: Reworked to `LagLogicalPort`.

niksv (approved):

nikitamarchenko (approved):

PR #4376, opened 2021-07-15
Rearrange history models naming scheme

surabujin:
The goal is to make clear and consistent naming scheme.


src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/HistoryMapper.java
```diff
@@ -186,4 +180,38 @@ protected String mapPath(FlowPath path) {
...
+    public long mapTimestamp(Instant value) {
+        return value.getEpochSecond();
+    }
+
+    /**
+     * Map {@link Cookie} into {@link long}.
+     */
+    public long mapCookie(Cookie value) {
+        if (value == null) {
+            return 0;
```
sergii-iakovenko, line 193: Before, null cookie was mapped to null Long. What is the reason to implicitly introduce 0 as equivalent to null cookie? If null cookie is not acceptable then throw an exception.

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/HistoryMapper.java
```diff
@@ -186,4 +180,38 @@ protected String mapPath(FlowPath path) {
...
+        }
+        return value.getValue();
+    }
+
+    /**
+     * Map {@link MeterId} into {@link long}.
+     */
+    public long mapMeterId(MeterId value) {
+        if (value == null) {
+            return 0;
```
sergii-iakovenko, line 203: Same as above with null cookie ~ 0.

src-java/kilda-model/src/main/java/org/openkilda/model/history/FlowStatusView.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package org.openkilda.model.history;
+
+import lombok.Value;
+
+import java.time.Instant;
+
+@Value
+public class FlowStatusView {
```
sergii-iakovenko, line 23: Please, add javadoc for the class.

sergii-iakovenko (commented):
Looks good. Minor comments left.

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/HistoryMapper.java
surabujin: Reasoin is the possible NPE - target objects use "long" not "Long" in corresponding fields.

dpoltavets (approved):

niksv (approved):

PR #4395, opened 2021-07-30
Make all orientDB enum type converters caseless

surabujin:


dpoltavets (approved):

rozdy (approved):

sergii-iakovenko (commented):
Looks good. Please, adapt this to the blue-green deployment approach.

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/frames/converters/FlowApplicationConverter.java
```diff
@@ -28,14 +28,14 @@ public String toGraphProperty(FlowApplication value) {
         if (value == null) {
             return null;
         }
-        return value.name();
+        return value.name().toLowerCase();
```
sergii-iakovenko, line 31: If it's the first release in the migration series, then how can we make sure that we can rollback this release? Any stored data in the lower case can't be read by the previous one. Am I right?

niksv (approved):

PR #4404, opened 2021-08-05
Use diverse group id property in the flow frame instead of group id property

dpoltavets:


docker/db-migration/migrations/012-copy-from-flow-group-id-field-to-diverse-group-id-field.yaml
```diff
@@ -0,0 +1,17 @@
...
+      author: dpoltavets
+      changes:
+        - tagDatabase:
+            tag: 012-copy-from-flow-group-id-field-to-diverse-group-id-field
+
+  - changeSet:
+      id: copy_from_flow_group_id_field_to_diverse_group_id_field
+      author: dpoltavets
+      changes:
+        - sql: "UPDATE flow SET diverse_group_id = group_id"
```
sergii-iakovenko, line 13: Does it make sense to update only record where diverse_group_id is null?

dpoltavets: Added a condition where we update only records where diverse_group_id is null

sergii-iakovenko (approved):

rozdy (approved):

surabujin (approved):

PR #4413, opened 2021-08-07
fix test mirror+connectedDevice in singleTable

andriidovhan:


rtretyak (approved):

rtretyak:
Ok hotfix. The whole connected devices spec should be 'assume multitable' when we get time to refactor. We don't really care about single table at this point

PR #4425, opened 2021-08-10
Improve mysql init script compatibility

surabujin:


sirmax123 (approved):

niksv (approved):

docker/mysql/setup_data/initial-schema.sql
```diff
@@ -2,58 +2,58 @@
...
+    `task_id_unique_key` char(100)   not null,
+    `action`             text        not null,
+    `unstructured`       text        not null,
+    `time_create`        datetime(3) not null default now(3),
+    `time_modify`        datetime(3),
+    `event_time`         datetime(3),
+    index `ix_event_time` (`event_time`),
+    index `ix_flow_id` (`flow_id`(64)),
+    unique `ix_task_id` (`task_id_unique_key`(100))
+) engine=innodb;
```
sergii-iakovenko, line 17: Are we sure that the engine must be hardcoded?

surabujin: We need transactions, so we need innodb. We are sharing DB with other services, so we can't force any global default settings.

PR #4437, opened 2021-08-16
Introduce Y-Flow Northbound API

sergii-iakovenko:


timofei-durakov (approved):

surabujin (commented):
I would prefer 2 separate PRs. One for dependencies fix, another for y-flow API.

surabujin (approved):

sergii-iakovenko:
> I would prefer 2 separate PRs. One for dependencies fix, another for y-flow API.
I can split, but what is for? Payload validation requires another versions of api & hibernate validator. So, PR with y-flow API requires that changes (including dependency fix).

PR #4440, opened 2021-08-16
Remove one switch flows from flow monitoring

rozdy:


dpoltavets (approved):

src-java/flowmonitoring-topology/flowmonitoring-messaging/src/main/java/org/openkilda/messaging/info/flow/UpdateFlowInfo.java
```diff
@@ -65,4 +68,9 @@ public Long getMaxLatency() {
     public Long getMaxLatencyTier2() {
         return maxLatencyTier2 == null || maxLatencyTier2 == 0 ? Long.MAX_VALUE : maxLatencyTier2;
     }
+
+    public boolean isRemoveFlowFlowMonitoring() {
```
surabujin, line 72: Terrible name. `isRemoveRequest` fit better, but still not perfect.

PS If it is some sort of command, why it is `InfoMessage`? Looks like it should be `CommandMessage`.

surabujin (changes requested):
Where is the "Remove one switch flows from flow monitoring"? Into the change there is some refactoring without logic changes.

src-java/flowmonitoring-topology/flowmonitoring-messaging/src/main/java/org/openkilda/messaging/info/flow/UpdateFlowInfo.java
rozdy: Reworked method name and `InfoMessage` to `CommandMessage`.

Actual bugfix was change `path.isEmpty()` -> `path.size() <= 1`. One switch flows will have path with one segment in this representation. And we need to remove such flows from monitoring.

surabujin (approved):

PR #4450, opened 2021-08-19
Test/dump sw props

andriidovhan:


src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/SwitchHelper.groovy
```diff
@@ -141,7 +145,7 @@ class SwitchHelper {
     }
 
     static List<Long> getDefaultCookies(Switch sw) {
-        def swProps = northbound.get().getSwitchProperties(sw.dpId)
+        def swProps = getSwitchProperties(sw.dpId)
```
rtretyak, line 148: This change should break many tests that rely on `getDefaultCookies` method to verify switch state after some switch property was changed

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/SwitchHelper.groovy
```diff
@@ -530,7 +534,17 @@ class SwitchHelper {
 
     @Memoized
     static boolean isVxlanEnabled(SwitchId switchId) {
-        return northbound.get().getSwitchProperties(switchId).supportedTransitEncapsulation
+        return getSwitchProperties(switchId).supportedTransitEncapsulation
                 .contains(FlowEncapsulationType.VXLAN.toString().toLowerCase())
     }
+
+    @Memoized
+    static List<SwitchPropertiesDto> getAllSwitchProperties() {
```
rtretyak, line 542: This should be called `getAllSwitchPropertiesCached` or sth like that in order to clearly state that we are working with cache and not requesting fresh data from the server

rtretyak (approved):

PR #4455, opened 2021-08-23
Fix updating last-updated flow property when updating path latencies

### LINKED ISSUE #4354: Flow's 'last-updated' field is updated every minute
**Summary:** Kilda flows now have `forward_path_latency_ns` and `reverse_path_latency_ns` fields. These fields are automatically updated by '[flow monitoring](https://github.com/telstra/open-kilda/tree/develop/docs/design/flow-monitoring)' every minute. Every such update leads to an update of `last_updated` field, which is undesirable.
**Steps:**
1. Create a flow
2. Let it stay for couple minutes. Call 'get flow' API and check `last_updated` field
**Expected:** `last_updated` has the same value as `created` field, since no actual flow updates happened
**Actual:** `last_updated` is updated every minute together with `forward_path_latency_ns` and `reverse_path_latency_ns` fields

rozdy:
Introduce Flow Properties entity

Close #4354

### DIFF
--- a/src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/FlowMapper.java
+++ b/src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/FlowMapper.java
@@ -26,6 +26,7 @@ import org.openkilda.model.FlowEncapsulationType;
 import org.openkilda.model.FlowMirrorPath;
 import org.openkilda.model.FlowPath;
 import org.openkilda.model.FlowPathStatus;
+import org.openkilda.model.FlowStats;
 import org.openkilda.model.FlowStatus;
 import org.openkilda.model.KildaConfiguration;
 import org.openkilda.model.MeterId;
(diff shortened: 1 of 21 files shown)


dpoltavets (approved):

timofei-durakov (approved):

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/FlowMapper.java
```diff
@@ -75,15 +76,27 @@
     @Mapping(target = "diverseWith", ignore = true)
     @Mapping(source = "affinityGroupId", target = "affinityWith")
     @Mapping(target = "mirrorPointStatuses", ignore = true)
+    @Mapping(target = "forwardLatency", ignore = true)
+    @Mapping(target = "reverseLatency", ignore = true)
     public abstract FlowDto map(Flow flow);
```
surabujin, line 81: Can we/should we pass `FlowStats` here as second argument?

src-java/kilda-model/src/main/java/org/openkilda/model/FlowStats.java
```diff
@@ -0,0 +1,167 @@
...
+        return Objects.hash(getFlowId(), getForwardLatency(), getReverseLatency());
+    }
+
+    /**
+     * Defines persistable data of the FlowStats.
+     */
+    public interface FlowStatsData {
+        String getFlowId();
+
+        Flow getFlowObj();
```
surabujin, line 111: Name `flowObj` looks weird. Why not just `flow` or something like `owner`?

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/frames/FlowStatsFrame.java
```diff
@@ -0,0 +1,90 @@
...
+import org.apache.tinkerpop.gremlin.structure.Direction;
+import org.apache.tinkerpop.gremlin.structure.Edge;
+
+import java.util.List;
+import java.util.Objects;
+
+public abstract class FlowStatsFrame extends KildaBaseVertexFrame implements FlowStatsData {
+    public static final String FRAME_LABEL = "flow_stats";
+    public static final String HAS_BY_EDGE = "has";
+    public static final String FLOW_ID_PROPERTY = "flow_id";
```
surabujin, line 34: Can we get rid of `flow_id` field into this table/class? Its existence forces us to track changes and provide updates in all copies by "hands".

surabujin (approved):

PR #4456, opened 2021-08-23
Ability exclude specific objects from history dump

surabujin:


dpoltavets (approved):

surabujin:
Depends on #4431 (need to merged after it).

PR #4470, opened 2021-09-03
[Server42] Connection to Test lab

nikitamarchenko:
Add ability to test server42 related rules on local OVS Test lab

- [Documentaion](https://github.com/telstra/open-kilda/blob/23917dff1850f1201497aaba139785fa60c05d0a/docs/design/server42/server42-to-lab/README.md)
- PR:#4468


timofei-durakov (approved):

src-cpp/server42/Dockerfile
```diff
@@ -12,4 +12,6 @@ WORKDIR /root/cmake-build-release/
 
 COPY ./docker_entrypoint.sh /root/cmake-build-release/
```
rtretyak, line 13: should be removed since `.sh` file does not exist anymore. This fails the image build atm

timofei-durakov (commented):
is local_loop.sh still used? If yes, please add a small comment in it with scenario when and where it might be useful

src-cpp/server42/Dockerfile
nikitamarchenko: done

nikitamarchenko:
> is local_loop.sh still used? If yes, please add a small comment in it with scenario when and where it might be useful

tnx, done!

timofei-durakov (approved):

rtretyak (approved):

niksv (approved):

PR #4492, opened 2021-09-23
Adjust config to work with local s42

rtretyak:


andriidovhan (approved):

confd/vars/docker-compose.yaml
```diff
@@ -9,9 +9,9 @@
 #no_elasticsearch: true
 #no_logstash: true
 #no_grpc_stub: true
-no_server42_server: true
```
nikitamarchenko, line 12: you remove all server42 instances stub and real dpdk

nikitamarchenko (changes requested):

andriidovhan:
it is not merged yet due to an issue with `hpet` on a Jenkins slave

confd/vars/docker-compose.yaml
andriidovhan: It was done intentionally, I guess.
(for running real s42 on virtual env)

PR #4494, opened 2021-09-26
tests: Properly wait for result when sending to kafka

rtretyak:


andriidovhan (approved):
Please, add the same fix in the StromHeavyLoadSpec file.

PR #4496, opened 2021-09-28
Adjust tests configs for local env

rtretyak:


confd/templates/functional-tests/kilda.properties.tmpl
```diff
@@ -0,0 +1,100 @@
+# Generated by confd.
```
andriidovhan, line 1: is it really generated file?
I thought the `properties.template` file is predefined, and the real `properties` file is generated

rtretyak: This is added in all template files, so I'm not changing this now. I guess the idea was that this info also gets copied to the actual generated properties file, where it is actually generated by confd

andriidovhan: it makes sense, thx.

confd/vars/test-vars.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+kilda_lab_api_endpoint: "http://localhost"
+kilda_lab_api_port: "8288"
+
+kilda_tests_northbound_endpoint: "http://localhost"
+kilda_tests_northbound_rest_port: "8080"
+
+kilda_tests_logging_logstash_host: "localhost"
+kilda_tests_logging_elasticsearch_proto: "http"
+kilda_tests_logging_elasticsearch_hosts: "localhost:9200"
```
andriidovhan, line 30: How about extracting `9200` into the separate variable(`kilda_tests_logging_elasticsearch_port`) ?
The same can be applied for `kilda_tests_kafka_hosts` as well.
Probably it can be implemented as an improvement in the future.

andriidovhan (approved):

PR #4521, opened 2021-10-08
Remove confd from local dev dependencies

nikitamarchenko:


Makefile
```diff
@@ -1,17 +1,29 @@
 # 'make' will build docker images and try to run it.
-default: update-props build-stable up-test-mode
+#default: update-props build-stable up-test-mode
```
timofei-durakov, line 2: maybe just remove this line at all?

PR #4530, opened 2021-10-14
YFlow changes to persistence layer

sergii-iakovenko:
The changes:
- YFlow changes to persistence (yflow and ysubflow entities, flow meter, feature toggles).
- Update MeterPool correspondingly.

### DIFF
--- a/src-java/kilda-model/src/main/java/org/openkilda/model/YFlow.java
+++ b/src-java/kilda-model/src/main/java/org/openkilda/model/YFlow.java
@@ -1,0 +1,355 @@
+/* Copyright 2021 Telstra Open Source
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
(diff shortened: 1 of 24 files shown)


dpoltavets (approved):

src-java/kilda-model/src/main/java/org/openkilda/model/YFlow.java
```diff
@@ -0,0 +1,335 @@
...
+
+        @Override
+        public Set<YSubFlow> getSubFlows() {
+            return Collections.unmodifiableSet(subFlows);
+        }
+
+        @Override
+        public void setSubFlows(Set<YSubFlow> subFlows) {
+            subFlows.forEach(flow -> {
+                flow.setYFlow(yFlow);
```
surabujin, line 299: Should we clear y-flow reference for exising on this moment list of sub fows?

surabujin (approved):

PR #4538, opened 2021-10-20
Rule manager: Ingress rules

niksv:


dpoltavets (approved):

src-java/kilda-model/src/main/java/org/openkilda/model/IPv4Address.java
```diff
@@ -0,0 +1,58 @@
+/* Copyright 2020 Telstra Open Source
```
nikitamarchenko, line 1: wtf

nikitamarchenko (approved):

src-java/kilda-model/src/main/java/org/openkilda/model/IPv4Address.java
niksv: copy paste :)

timofei-durakov: updated

timofei-durakov (approved):

PR #4541, opened 2021-10-21
Spec proposal for SLA of the flow

timofei-durakov:
Related to #578

### DIFF
--- a/docs/design/flow-sla/README.md
+++ b/docs/design/flow-sla/README.md
@@ -1,0 +1,96 @@
+# SLA for flows in Kilda
+
+## Problem
+
+Since the first release there were only one SLA for Kilda flows - bandwidth. But now there are more sla's that are taken
+into account for the flows. E.g. latency, latency_tier2. The purpose of this doc is to rethink and redesign how these 
+SLAs are visible outside the system and define strategy for adding new ones.
+
+## Current State
+
+Here is the list of flow fields that are taken into account by path computer:
+- maximum bandwidth
+- max_latency
+- max_latency_tier2 
+- diverse_flow_id
+- affinity_flow_id
+- allocate_protected_path
+- ignore_bandwidth
+- strict_bandwidth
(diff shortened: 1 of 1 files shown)


docs/design/flow-sla/README.md
```diff
@@ -0,0 +1,96 @@
...
+  "sla": {
+    "bandwidth_tier1": 0,
+    "bandwidth_tier2": 0,
+    "latency_tier1": 0, 
+    "latency_tier2": 0,
+    "affinity_flow_id": "string",
+    "diversity_flow_id": "string",
+    "protected_path": true,
+    "strict_bandwidth": true,
+  }
```
rtretyak, line 63: If we are to rethink this completely, then I would have waived the tier1/tier2 approaches and introduced  'fallback' SLA objects. If prio:1 SLA has failed, try to apply next priority SLAs and apply Degraded status to the flow (maybe even introduce the "degradationLevel" value that will be scored according to the SLA priority that was used to build the flow). If none can be applied - flow creation is denied completely (or flow reroute makes it Down)
```
  "slas": [
         {
             "bandwidth": 100,
             "latency": 100, 
             "affinity_flow_id": "string",
             "diversity_flow_id": "string",
             "protected_path": true,
             "ignore_bandwidth": false
             "priority": 1
         },
          {
             "bandwidth": 50,
             "latency": 200, 
             "affinity_flow_id": "string",
             "protected_path": false,
             "ignore_bandwidth": false,
             "priority": 2
         },
          {
             "bandwidth": 50,
             "latency": 300, 
             "protected_path": false,
             "ignore_bandwidth": true,
             "priority": 3
         }
]
```

rtretyak: This may also be used to simplify all our 'retry' strategies. (retry with ignore bandwidth, retry with latency_tier2 etc.). System will just cycle through available SLAs

PR #4534, opened 2021-10-18
Implement YFlow delete feature

sergii-iakovenko:
The changes:
- Implement the delete operation for Y-flow.

PR #4542, opened 2021-10-22
Adapt existing Flow API to support YFlow sub-flows

sergii-iakovenko:
The changes:
- Extend Flow responses in Flow APIs with y-flow field.		
- Forbid amend (update, delete, etc.) ops for YFlow sub-flows.

**MERGE NOTE:** merge after #4534 and change the base correspondingly.

### DIFF
--- a/src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/FlowMapper.java
+++ b/src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/FlowMapper.java
@@ -79,6 +79,7 @@ public abstract class FlowMapper {
     @Mapping(target = "forwardLatency", ignore = true)
     @Mapping(target = "reverseLatency", ignore = true)
     @Mapping(target = "latencyLastModifiedTime", ignore = true)
+    @Mapping(target = "yFlowId", source = "YFlowId")
     public abstract FlowDto map(Flow flow);
 
     /**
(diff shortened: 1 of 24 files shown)


src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/FlowMapper.java
```diff
@@ -79,6 +79,7 @@
     @Mapping(target = "forwardLatency", ignore = true)
     @Mapping(target = "reverseLatency", ignore = true)
     @Mapping(target = "latencyLastModifiedTime", ignore = true)
+    @Mapping(target = "yFlowId", source = "YFlowId")
```
surabujin, line 82: Looks excess, mapstruct can find this relation by himself.

surabujin (approved):

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/FlowMapper.java
sergii-iakovenko: No. In our Mapstruct configuration, the AccessorNamingStrategy is case sensitive. And here, the source is a getter method "getYFlowId", so the mapper sees it as "YFlowId". The target is a builder method "yFlowId". They are different from the mapper standpoint.

PR #4558, opened 2021-10-29
[test] replace getSwProps by getCachedSwProps

andriidovhan:


rtretyak (approved):
Haven't reviewed each change in too many details. If gate passes then it's ok

PR #4560, opened 2021-11-01
Feature/rule manager persistence adapter

rozdy:


dpoltavets (approved):

src-java/rule-manager/rule-manager-implementation/src/main/java/org/openkilda/rulemanager/adapter/PersistenceDataAdapter.java
```diff
@@ -0,0 +1,125 @@
...
+        if (flowCache == null) {
+            flowCache = flowPathRepository.findFlowsByPathIds(pathIds);
+        }
+        return flowCache.get(pathId);
+    }
+
+    @Override
+    public FlowTransitEncapsulation getTransitEncapsulation(PathId pathId) {
+        if (encapsulationCache.get(pathId) == null) {
+            encapsulationCache = new HashMap<>();
```
niksv, line 95:
```
getTransitEncapSulation("id1") 
getTransitEncapSulation("id2")
```

Second call will remove information about "id1" from cache because `encapsulationCache.get("id2")` is null and cache will be set to empty set

nikitamarchenko (approved):

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaFlowPathRepository.java
```diff
@@ -78,6 +80,34 @@ public FermaFlowPathRepository(FermaPersistentImplementation implementation) {
...
+        Set<String> graphPathIds = pathIds.stream()
+                .map(PathIdConverter.INSTANCE::toGraphProperty)
+                .collect(Collectors.toSet());
+        List<? extends FlowPathFrame> flowPathFrames = framedGraph().traverse(g -> g.V()
+                .hasLabel(FlowPathFrame.FRAME_LABEL)
+                .has(FlowPathFrame.PATH_ID_PROPERTY, P.within(graphPathIds)))
+                .toListExplicit(FlowPathFrame.class);
+        return flowPathFrames.stream()
+                .map(FlowPath::new)
+                .collect(Collectors.toMap(FlowPath::getPathId, FlowPath::getFlow));
```
sergii-iakovenko, line 108: Not optimal. You can traverse further from FlowPath vertex (node) to Flow vertex, and get it.

src-java/rule-manager/rule-manager-implementation/src/main/java/org/openkilda/rulemanager/adapter/PersistenceDataAdapter.java
```diff
@@ -0,0 +1,124 @@
...
+    private final FlowPathRepository flowPathRepository;
+    private final SwitchRepository switchRepository;
+    private final SwitchPropertiesRepository switchPropertiesRepository;
+    private final TransitVlanRepository transitVlanRepository;
+    private final VxlanRepository vxlanRepository;
+
+    private final Set<PathId> pathIds;
+    private final Set<SwitchId> switchIds;
+
+    private Map<PathId, Flow> flowCache;
```
sergii-iakovenko, line 54: What is the purpose of these caches? Even changes within the same persistence context (transaction) wont invalidate them.

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaFlowPathRepository.java
rozdy: If I traverse further to Flow vertex I will loose selected pathIds so additional backward filtration will be required.

src-java/rule-manager/rule-manager-implementation/src/main/java/org/openkilda/rulemanager/adapter/PersistenceDataAdapter.java
rozdy: PersistenceDataAdapter class is designed to be used for one operation only. So consequent calls to it's API should return the same result and not producing a lot of DB calls. No invalidation is designed.

PR #4600, opened 2021-11-26
Rule Manager: Flow loop

niksv:


dpoltavets (approved):

src-java/rule-manager/rule-manager-implementation/src/main/java/org/openkilda/rulemanager/factory/FlowRulesGeneratorFactory.java
```diff
@@ -96,6 +98,17 @@ public RuleGenerator getIngressYRuleGenerator(
         }
     }
 
+    /**
+     * Get ingress loop loop generator.
```
rozdy, line 102: loop loop -> loop rule

rozdy (approved):

PR #4637, opened 2021-12-28
improve defaultRuleSpec(temporary add sleep)

andriidovhan:
default rule is installing(install_default/syncSw) twice by the same floodlight

kibana request: i.e. `"installing" AND "<switchId>" AND "0x8030000000000004"`
it can cause instability during running func tests on jenkins
add `sleep` to avoid instability, 
this pr is going to reverted once ruleManager is implemented/merged.


rtretyak (approved):

nikitamarchenko:
nice

PR #4638, opened 2021-12-28
Of match converter

timofei-durakov:


src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/converter/rulemanager/OfMatchConverter.java
```diff
@@ -32,150 +34,168 @@
...
-                fieldMatches.add(FieldMatch.builder().field(Field.UDP_DST).value(udpDst.getPort()).build());
-            }
-        }
+    private FieldMatch getMasked(Match match, MatchField field) {
+        FieldMatchBuilder builder = FieldMatch.builder();
+        switch (field.id) {
+            case ETH_SRC:
+                Masked<MacAddress> ethSrcMasked = match.getMasked(MatchField.ETH_SRC);
+                builder.field(Field.ETH_SRC);
+                builder.value(ethSrcMasked.getValue().getLong());
```
surabujin, line 129: Value and mask field object have same class and can be handled in same way. It is possible to regroup processing to get rid from mostly same handlers for exacd/masked fields. I am not insist on it, but it would be good.

surabujin (approved):

rozdy (approved):

PR #4682, opened 2022-01-27
[yflow] add tests for protected path

andriidovhan:


andriidovhan:
close as duplicate of #4808

PR #4684, opened 2022-01-27
Replace meter pool with generic pool manager

surabujin:


dpoltavets (approved):

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/utils/PoolEntityAdapter.java
```diff
@@ -17,12 +17,10 @@
...
 
-public interface PoolEntityAdapter<T> {
-    long getNumericSequentialId(T entity);
+public interface PoolEntityAdapter {
+    boolean allocateSpecificId(long entityId);
 
-    Optional<T> allocateSpecificId(long entityId);
-
-    Optional<T> allocateFirstInRange(long idMinimum, long idMaximum);
+    Optional<Long> allocateFirstInRange(long idMinimum, long idMaximum);
```
rozdy, line 23: minor: I don't like naming here. As for me this methods doesn't actually allocate entities, they only check ability to do so or find best suitable value. But looks like we already have used this interface and naming in other code.

rozdy (approved):

PR #4687, opened 2022-01-31
[test] check yFlow periodicPings

andriidovhan:


niksv (approved):
Looks good. 
I think we can add one improvement. 
We can check otsdb metric `flow.latency` and check that it has correct tags `flowid` and `y_flow_id`

Issue #4629, opened 2021-12-22
deleteAllFlows endpoint is returning 'timeout' error in case flow is in the "in_progress" state


andriidovhan:
**Steps to delete:**

1. create a flow;
2. change the flow status(via db) to "in_progress";
3. try to delete the flow using the `deleteAllFlows` endpoint.

**Expected result:**
Human-readable error is returned.
for example:
```
    "error-message": "Could not delete flow",
    "error-description": "Flow <flow-id> is in progress now"
```

**Actual result:**
```
   "error-message": "No response received",
    "error-description": "Timeout exceeded"
```
no error in kibana.

PR #4718, opened 2022-03-05
Fixed deleteAllFlows endpoint timeout while flow in progress

nrydanov:
Response now includes all unsuccessful flow delete requests.
Closes: #4629

### DIFF
--- a/src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/service/impl/FlowServiceImpl.java
+++ b/src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/service/impl/FlowServiceImpl.java
@@ -404,7 +404,7 @@ public class FlowServiceImpl implements FlowService {
                                 System.currentTimeMillis(),
                                 ErrorType.REQUEST_INVALID,
                                 "Couldn't delete flow",
-                                errorBuilder.append(String.join(",", failedFlows))
+                                errorBuilder.append(String.join(", ", failedFlows))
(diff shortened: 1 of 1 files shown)


dpoltavets (approved):

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/service/impl/FlowServiceImpl.java
```diff
@@ -385,12 +386,29 @@
...
-                .thenApply(result::complete));
-
+                .thenApply(results -> {
+                    if (!failedFlows.isEmpty()) {
+                        StringBuilder errorBuilder = new StringBuilder("The following flows haven't been deleted: ");
+                        result.completeExceptionally(new MessageException(correlationId,
+                                System.currentTimeMillis(),
+                                ErrorType.REQUEST_INVALID,
+                                "Couldn't delete flow",
+                                errorBuilder.append(String.join(",", failedFlows))
```
niksv, line 407: I think it's better to have `", "` instead of `","` but it's a minor comment

niksv (approved):

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/service/impl/FlowServiceImpl.java
nrydanov: Thanks, agree with you.

PR #4733, opened 2022-03-21
GRPC stub: return errors on invalid requests

### LINKED ISSUE #4728: GRPC stub: return errors on invalid requests:
GRPC stub must emulate work with real GRPC. For that we need to return error messages on following requests:

- [ ]  Error code 190. `Ports of the LAG can not be duplicated.` on duplicate ports in request
- [ ]  Error code 31. `Invalid portno value.` on non existent physical port value
- [ ]  Error code 185. `Valid logicalportno range is 100 to 63487.` if requested LAG port number is not in range [100, 63487]

vshakirova:
Close #4728

### DIFF
--- a/src-python/grpc-stub/server.py
+++ b/src-python/grpc-stub/server.py
@@ -47,6 +47,16 @@ class Greeter(noviflow_pb2_grpc.NoviFlowGrpcServicer):
         return noviflow_pb2.CliReply(reply_status=0)
 
     def SetConfigLogicalPort(self, request, context):
+        if request.logicalportno not in range(100, 63488):
+            return noviflow_pb2.CliReply(reply_status=185)
+
+        if len(request.portno) > len(set(request.portno)):
+            return noviflow_pb2.CliReply(reply_status=190)
+
+        for port in request.portno:
+            if port not in range(1, 100):
+                return noviflow_pb2.CliReply(reply_status=31)
+
         if request.logicalportno in self.storage.logical_ports:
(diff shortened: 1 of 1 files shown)


src-python/grpc-stub/server.py
```diff
@@ -47,6 +47,16 @@ def SetLoginDetails(self, request, context):
         return noviflow_pb2.CliReply(reply_status=0)
 
     def SetConfigLogicalPort(self, request, context):
+        if request.logicalportno not in range(100, 63488):
```
niksv, line 50: I think `100 <= request.logicalportno <= 63488:` is better because `range` generated an iterator and check elements in range form 100 to 63488. Instead `a <= x <= b` uses 2 comparisons. So it is faster.

src-python/grpc-stub/server.py
```diff
@@ -47,6 +47,16 @@ def SetLoginDetails(self, request, context):
...
+            return noviflow_pb2.CliReply(reply_status=185)
+
+        if len(request.portno) > len(set(request.portno)):
+            return noviflow_pb2.CliReply(reply_status=190)
+
+        for port in request.portno:
+            if port not in range(1,100):
```
niksv, line 57: nit: you missed space between 1 and 100

niksv, line 57: nit: you missed space between 1 and 100

niksv (approved):

dpoltavets (approved):

PR #4740, opened 2022-03-23
Replace flow cookie pool with generic pool manager

surabujin:
Continue #4684


sergii-iakovenko (changes requested):
Is it really worth it? -114 lines of CookiePool, but +51 of FlowCookieEntityAdapter and +47 of FlowResourcesManager.java... BTW, where is unit testing for cookie pool operations which involve FlowCookieEntityAdapter, FlowResourcesManager and PersistenceManager?

PR #4743, opened 2022-03-27
[test] ignore wb for qinq in YFlowCreateSpec

andriidovhan:


andriidovhan:
it is going to be obsolete once #4448 is merged

PR #4746, opened 2022-03-30
Use MySQL DB to store history on local env

niksv:


sergii-iakovenko (approved):

andriidovhan:
We had an issue with MySQL on Jenkins previously.
It should be retested before merging.

PR #4771, opened 2022-04-13
Changed examples of flow history endpoint

nrydanov:
A new additional field with human-readable timestamp is now required


niksv:
Looks good, but could you also move `flow-history-get` directory out of `outdated`. For example you can put it in `docs/design` directory

dpoltavets (approved):

sergii-iakovenko (approved):

PR #4778, opened 2022-04-19
Add test to validate non existing lag port creation

vkhalin:


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/LagPortSpec.groovy
```diff
@@ -405,6 +405,11 @@ class LagPortSpec extends HealthCheckSpecification {
                         description: "is more than lagOffset",
                         portNumber : { 2008 },
                         errorMsg   : "Physical port number %d can't be greater than LAG port offset $lagOffset."
+                ],
+                [
+                        description: "with non existing physical port number",
```
andriidovhan, line 410: I would change the description. (for example just `not exist`)
During execution, it will be concatenated with the test name
`Unable to create a LAG port in case port is` + `with non existing physical port number`

andriidovhan: nothing changed

andriidovhan (approved):

vshakirova (approved):

PR #4784, opened 2022-04-20
Add LAG ports test for #4725

vkhalin:
Add LAG ports test with duplicated ports during create operation to
check that duplicates are deleted and port created successfully

### DIFF
--- a/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/LagPortSpec.groovy
+++ b/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/LagPortSpec.groovy
@@ -521,6 +521,71 @@ class LagPortSpec extends HealthCheckSpecification {
         lagPort && northboundV2.deleteLagLogicalPort(sw.dpId, lagPort)
     }
 
+    @Tidy
+    def "Able to create/update LAG port with duplicated port numbers on the #sw.hwSwString switch"() {
+        given: "Switch and two ports"
+        def sw = getTopology().getActiveSwitches().get(0)
+        def testPorts = topology.getAllowedPortsForSwitch(sw).take(2)
+        assert testPorts.size > 1
+
+        when: "Create LAG port with duplicated port numbers"
+        def switchPortToCreate = testPorts.get(0)
+        def swAddress = northbound.getSwitch(sw.dpId).address
+        def portListToCreate = [switchPortToCreate, switchPortToCreate]
+        def createPayload = new LagPortRequest(portNumbers: portListToCreate)
+        def lagPortCreateResponse = northboundV2.createLagLogicalPort(sw.dpId, createPayload)
+
+        then: "Response shows that LAG port created successfully"
+        with(lagPortCreateResponse) {
+            logicalPortNumber > 0
+            portNumbers == [switchPortToCreate]
+        }
+        def lagPort = lagPortCreateResponse.logicalPortNumber
+
+        and: "Request on user side shows that LAG port created"
+        with(northboundV2.getLagLogicalPort(sw.dpId)[0]) {
+            logicalPortNumber == lagPort
+            portNumbers == [switchPortToCreate]
+        }
+
+        and: "Created port exists in a list of all LAG ports from switch side (GRPC)"
+        with(grpc.getSwitchLogicalPortConfig(swAddress, lagPort)) {
+            logicalPortNumber == lagPort
+            name == "novi_lport" + lagPort.toString()
+            portNumbers == [switchPortToCreate]
+            type == LogicalPortType.LAG
+        }
+
+        when: "Update the LAG port with duplicated port numbers"
+        def switchPortToUpdate = testPorts.get(1)
+        def portListToUpdate = [switchPortToUpdate, switchPortToUpdate]
+        def updatePayload = new LagPortRequest(portNumbers: portListToUpdate)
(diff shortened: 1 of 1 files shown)


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/LagPortSpec.groovy
```diff
@@ -516,6 +516,70 @@ class LagPortSpec extends HealthCheckSpecification {
...
+            name == "novi_lport" + lagPort.toString()
+            portNumbers == [switchPortToUpdate]
+            type == LogicalPortType.LAG
+        }
+
+        cleanup:
+        lagPort && northboundV2.deleteLagLogicalPort(sw.dpId, lagPort)
+
+        where:
+        sw << getTopology().getActiveSwitches().unique(false)
```
andriidovhan, line 580: There is no sense to iterate over all switches. We just need to test that code validation works fine, it doesn't depend on a model of a switch.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/LagPortSpec.groovy
```diff
@@ -516,6 +516,70 @@ class LagPortSpec extends HealthCheckSpecification {
...
+        and: "GRPC shows that created port exist in a list of all LAG ports"
+        with(grpc.getSwitchLogicalPortConfig(swAddress, lagPort)) {
+            logicalPortNumber == lagPort
+            name == "novi_lport" + lagPort.toString()
+            portNumbers == [switchPortToCreate]
+            type == LogicalPortType.LAG
+        }
+
+        when: "Update the LAG port with duplicated port numbers"
+        def switchPortToUpdate = topology.getAllowedPortsForSwitch(sw)[1]
```
andriidovhan, line 550: `topology.getAllowedPortsForSwitch(sw)` you can extract it in a separate variable, it is already used twice.

andriidovhan: also, there is some chance that `topology.getAllowedPortsForSwitch(sw)` return just one element. It can be handled early.
I would check the size of the array at the beginning of the test  and remove L551 `assert switchPortToCreate != switchPortToUpdate`

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/LagPortSpec.groovy
```diff
@@ -516,6 +516,70 @@ class LagPortSpec extends HealthCheckSpecification {
...
+        def updatePayload = new LagPortRequest(portNumbers: portListToUpdate)
+        def lagPortUpdateResponse = northboundV2.updateLagLogicalPort(sw.dpId, lagPort, updatePayload)
+
+        then: "Response shows that LAG port updated successfully"
+        with(lagPortUpdateResponse) {
+            logicalPortNumber == lagPort
+            portNumbers == [switchPortToUpdate]
+        }
+
+        and: "Northbound shows that LAG port updated successfully"
```
andriidovhan, line 562: `northbound` is just the name of service. Theoretically, it can be renamed. If so, the test step should be updated.
rephrase, please. for example: `LAG port is really updated` or `LAG port is really updated in system/db`

GRPC - is a protocol to send requests directly to a switch.
The step with grpc also can be updated. ie: `LAG port is really updated on the switch (check via grpc)`

vkhalin: Lets use something like "user side" for Northbound and "switch side" for GRPC. That will reflect the situation better

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/LagPortSpec.groovy
```diff
@@ -516,6 +516,71 @@ class LagPortSpec extends HealthCheckSpecification {
         lagPort && northboundV2.deleteLagLogicalPort(sw.dpId, lagPort)
     }
 
+    @Tidy
+    def "Able to create/update LAG port with duplicated port numbers on the #sw.hwSwString switch"() {
+        given: "Switch and two ports"
+        def sw = getTopology().getActiveSwitches().get(0)
+        def testPorts = topology.getAllowedPortsForSwitch(sw).take(2)
```
andriidovhan, line 523: fyi:
in groovy you can get elements like in python:
```
groovy:000> a = [1,2,3,4,5]
===> [1, 2, 3, 4, 5]
groovy:000> a[0..1]
===> [1, 2]
groovy:000> a[0]
===> 1
```

andriidovhan (approved):

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/LagPortSpec.groovy
vkhalin: yes ofc, but in this case it is error friendly when you use .get or .take. That is why I prefer to use it instead of [0..1]. If our code style requires to use direct work with arrays like [0] or [0..1] I will fix that

andriidovhan: no specific code style, it is up to you

vshakirova (approved):

PR #4795, opened 2022-04-28
decrease kilda_flow_sla_check_shard_count(10->1) for local env

andriidovhan:
due to the instability on Jenkins


niksv (approved):

niksv:
I need to test some cases when we can lose stats. That is why please do not merge it for now

niksv:
Here is the fix which must fix the problem https://github.com/telstra/open-kilda/pull/4819

PR #4819, opened 2022-05-23
Fixed incorrect flow sharding in FlowMonitoring topology

niksv:
Following expression was used for sharding:
`hashCode() % shardCount == shardNumber`

But it's incorrect because hash can be negative. In this case
`hashCode() % shardCount` will be negative too and never be equal to
positive `shardNumber`

### DIFF
--- a/src-java/flowmonitoring-topology/flowmonitoring-storm-topology/src/main/java/org/openkilda/wfm/topology/flowmonitoring/service/ActionService.java
+++ b/src-java/flowmonitoring-topology/flowmonitoring-storm-topology/src/main/java/org/openkilda/wfm/topology/flowmonitoring/service/ActionService.java
@@ -135,13 +135,19 @@ public class ActionService implements FlowSlaMonitoringCarrier {
             log.debug("Processing flow SLA checks for shard {}", shardNumber);
         }
         for (FsmKey key : fsms.keySet()) {
-            if (key.flowId.hashCode() % shardCount == shardNumber) {
+            if (needToCheckSla(key.flowId.hashCode(), shardNumber)) {
                 if (log.isTraceEnabled()) {
                     log.trace("Processing SLA check for flow FSM {}: Shard number: {}", key, shardNumber);
                 }
                 fsmExecutor.fire(fsms.get(key), Event.TICK, context);
             }
         }
+    }
+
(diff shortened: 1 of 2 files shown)


src-java/flowmonitoring-topology/flowmonitoring-storm-topology/src/main/java/org/openkilda/wfm/topology/flowmonitoring/service/ActionService.java
```diff
@@ -135,7 +135,7 @@ public void processTick(int shardNumber) {
             log.debug("Processing flow SLA checks for shard {}", shardNumber);
         }
         for (FsmKey key : fsms.keySet()) {
-            if (key.flowId.hashCode() % shardCount == shardNumber) {
+            if (needToCheckSla(key.flowId.hashCode(), shardNumber)) {
```
rozdy, line 138: Maybe it would be easier to use `Math.abs(hashCode) % shardCount`?

src-java/flowmonitoring-topology/flowmonitoring-storm-topology/src/main/java/org/openkilda/wfm/topology/flowmonitoring/service/ActionService.java
```diff
@@ -144,6 +144,12 @@ public void processTick(int shardNumber) {
         }
     }
 
+    @VisibleForTesting
+    boolean needToCheckSla(int hashCode, int shardNumber) {
+        // hashCode can be negative, so we can't use expression `hashCode() % shardCount == shardNumber`
+        return (hashCode + shardNumber) % shardCount == 0;
```
rozdy, line 150: Didn't get it. As far as I can see this sum still may be negative.

rozdy (approved):

src-java/flowmonitoring-topology/flowmonitoring-storm-topology/src/main/java/org/openkilda/wfm/topology/flowmonitoring/service/ActionService.java
sergii-iakovenko: Nice trick.

PR #4859, opened 2022-06-29
Fix flow monitoring issue

### LINKED ISSUE #4810: flows are not rerouted by flowMonitoring in case main path does not satisfy SLA (shrad_count > 0)
**Steps to reproduce:**

1. update `confd/vars/main.yaml`
```
kilda_flow_sla_check_shard_count: 2
```
2. create 3 the same flows with the `max_latency` strategy;
3. make sure that the path of all 3 flows are the same;
4. update path ( path does not satisfy SLA)

**Expected result:**
All 3 flows are rerouted due to `"details": "Reason: Flow latency become healthy",`

**Actual result:**
sometimes only one flow is rerouted, the rest flows are untouched

**NOTE:**
The issue can be reproduced by autotest `def "Able to detect and reroute a flow with MAX_LATENCY strategy when main path does not satisfy latency SLA"() {` from the `test/repr_issue_with_fl_monitoring_shard_count` branch

rozdy:
Close #4810 

Make flow monitoring stable for case when sla_timeout less then sla_check_interval param.


dpoltavets (approved):

niksv (approved):

rozdy:
Merge after #4853

PR #4876, opened 2022-07-15
Fixed Forward/Reverse cookid ID displaying in UI

nrydanov:
Closes: #2120

### DIFF
--- a/src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/payload/history/FlowDumpPayload.java
+++ b/src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/payload/history/FlowDumpPayload.java
@@ -40,9 +40,9 @@ public class FlowDumpPayload {
 
     private boolean strictBandwidth;
 
-    private long forwardCookie;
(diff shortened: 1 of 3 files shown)


src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/payload/history/FlowDumpPayload.java
```diff
@@ -40,9 +40,9 @@ public class FlowDumpPayload {
 
     private boolean strictBandwidth;
 
-    private long forwardCookie;
+    private String forwardCookie;
 
-    private long reverseCookie;
+    private String reverseCookie;
```
niksv, line 45: As it is existing API you can't change field names/types because clients of this API will get incorrect response (swith strings instead of longs). Correct way is to Create a new API or add new field into existing API. 

Previously we dace with same cases. To solve the problem you can add new fields 
`forwardCookieHex` and `reverseCookie`. You can fill them in mapper

niksv (changes requested):

PR #4877, opened 2022-07-15
Added query parameter to /api/v1/network/paths for limiting amount of computed paths

Nexgen90:
To avoid "Operation has timed out" on /api/v1/network/paths, added a query parameter max_path_count to reduce the time for calculations.

If max_path_count is not specified, then will be used default value which can be configured via topology.properties.

### DIFF
--- a/src-java/nbworker-topology/nbworker-storm-topology/src/test/java/org/openkilda/wfm/topology/nbworker/services/PathsServiceTest.java
+++ b/src-java/nbworker-topology/nbworker-storm-topology/src/test/java/org/openkilda/wfm/topology/nbworker/services/PathsServiceTest.java
@@ -17,12 +17,13 @@ package org.openkilda.wfm.topology.nbworker.services;
 
 import static junit.framework.TestCase.assertEquals;
 import static junit.framework.TestCase.assertTrue;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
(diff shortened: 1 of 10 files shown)


niksv (approved):

dpoltavets (approved):

vkhalin (approved):

src-java/nbworker-topology/nbworker-storm-topology/src/test/java/org/openkilda/wfm/topology/nbworker/services/PathsServiceTest.java
```diff
@@ -206,34 +209,45 @@ public void findNPathsByTransitVlanAndNullMaxLatency()
...
 
+    @Test
+    public void findNPathsByMaxPathCount()
+            throws UnroutableFlowException, SwitchNotFoundException, RecoverableException {
+        int maxPathCount = 3;
+        List<PathsInfoData> paths = pathsService.getPaths(SWITCH_ID_1, SWITCH_ID_2, VXLAN, COST, null,
+                null, maxPathCount);
+        assertThat(paths.size(), equalTo(maxPathCount));
+    }
+
```
vkhalin, line 240: do you need to check out of boundaries value like 0 or -1 here in the next unit test or should I?

Nexgen90: Ok, I have added the validation and unit test for it

PR #4879, opened 2022-07-19
Added hex field, which should display correctly in Swagger UI

nrydanov:


src-java/base-topology/base-storm-topology/src/test/java/org/openkilda/wfm/share/mappers/HistoryMapperTest.java
```diff
@@ -60,7 +60,7 @@ public class HistoryMapperTest {
     public static PathComputationStrategy PATH_COMPUTATION_STRATEGY = PathComputationStrategy.COST;
     public static boolean PERIODIC_PINGS = false;
     public static boolean PINNED = false;
-    public static int REVERSE_COOKIE = 14;
+    public static long REVERSE_COOKIE = 14L;
```
niksv, line 63: I think you should add test here to check that mapper works with hex

nrydanov: Ok, agree with you

dpoltavets (approved):

niksv (approved):

PR #4884, opened 2022-07-20
Replace opentsdb with victoriametrics

surabujin:
Use victoriametrtics opentsdb API wrapper to inject data in the same way
as we done with opentsdb.


src-java/testing/atdd-staging/kilda.properties
```diff
@@ -0,0 +1,91 @@
+# Generated by confd.
```
dpoltavets, line 1: Should this file be added to `.gitignore`?

src-java/testing/atdd-staging/log4j2.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Generated by confd.
```
dpoltavets, line 3: Should this file be added to `.gitignore`?

niksv:
Covered by https://github.com/telstra/open-kilda/pull/4940

PR #4887, opened 2022-07-20
Add test for limiting amount of computed paths #4877

vkhalin:
Test for #4877 and fix for NB service test implementation

### DIFF
--- a/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/network/PathsSpec.groovy
+++ b/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/network/PathsSpec.groovy
@@ -29,7 +29,8 @@ class PathsSpec extends HealthCheckSpecification {
         def flow = flowHelperV2.addFlow(flowHelperV2.randomFlow(switchPair))
 
         when: "Get paths between switches"
-        def paths = northbound.getPaths(switchPair.src.dpId, switchPair.dst.dpId, null, null, null, null)
+        def paths = northbound.getPaths(switchPair.src.dpId, switchPair.dst.dpId,
+                null, null, null, null, null)
 
         then: "Paths will be sorted by bandwidth (descending order) and then by latency (ascending order)"
         paths.paths.size() > 0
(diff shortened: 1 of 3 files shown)


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/network/PathsSpec.groovy
```diff
@@ -45,6 +46,13 @@ class PathsSpec extends HealthCheckSpecification {
             }
         }
 
+        then: "Maximum count of paths can be changed during PCE calculations"
+        def limited_paths = northbound.getPaths(switchPair.src.dpId, switchPair.dst.dpId,
+                null, null, null, null, 1)
+        assert limited_paths.paths.size() == 1
+        assert paths.paths.size() > limited_paths.paths.size()
```
andriidovhan, line 53: L26 should be improved.
you need to be sure that `switchPair` has two paths at least.

andriidovhan (approved):

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/network/PathsSpec.groovy
vkhalin: I do not think that we need exactly what you asking for in this particular case. We need to check that max_path_count var can limit response output to some number of available paths. In our case all switch pairs will have more than 1 path, so our asserts will be always correct

andriidovhan: agree, I forgot implementation of `getNotNeighboringSwitchPair`. 
It returns switchPair with two paths at least.

rtretyak: except that it is not)

rtretyak: it returns a switchpair with the shortest path being of size >2 (means that switches are not neighbours, just as method name implies)

niksv (approved):

PR #4931, opened 2022-08-19
Tool for dump/restore OpenTSDB data

surabujin:
This pull request includes a tool that allows for the configuration and restoration of opentsdb data. This tool will be useful for migrating data from opentsdb to a victoria metrics databse with opentsdb capabilities.

It is part of https://github.com/telstra/open-kilda/issues/4925


tools/otsdb-dump-restore/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Usage
+The tool is inside a Docker container. To use the it, run the following command in your terminal:
+```
+docker run kilda-otsdb-dump-restore <command_to_use> 
+```
+Note: if you need to run it in localhost, you need to add the `--network="host"` docker flag.
+
+### Commands:
+### `kilda-otsdb-dump --help`
```
dmitrii-beliakov, line 22: This doesn't look good as H3. I suggest you use some text form for the header and the command as paragraph:
### Dump data from OpenTSDB
Use this command `kilda-otsdb-dump` ...

the same for the next header. To me, `Commands` header is redundant here

tools/otsdb-dump-restore/build-tools/setup-devel-user.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
+
+set -eux
+
+echo "$@"
+
+DEVEL_NAME="${1}"
+DEVEL_UID="${2}"
+DEVEL_GID="${3}"
```
dmitrii-beliakov, line 9: Maybe some checking for parameters and some help message here saying to execute it with sudo?

tools/otsdb-dump-restore/setup.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python
```
dmitrii-beliakov, line 1: I believe all we need to extend the requirements for java sources to python as well. All py files need to have a copyright and some pythondocs where appropriate.
Also I don't see tests here. Could you please add some or add some guide that could be used to understand the reproduction steps. What you expect after doing certain steps.

dmitrii-beliakov (changes requested):

dmitrii-beliakov:
Please add some description, labels, and link this PR to appropriate issue. Thanks

tools/otsdb-dump-restore/build-tools/setup-devel-user.sh
pablomuri: This is not supposed to be used by the user directly. The user must use the Makefile, but I think I should explain this better.

tools/otsdb-dump-restore/README.md
```diff
@@ -1,26 +1,24 @@
...
 
-# Create docker image
-
-```bash
-make
-````
+## Build
+`make prod` - builds a container image with the application, it labels it with kilda-otsdb-dump-restore
+`make devel` - builds a container and starts an interactive shell inside it (app directory is attached inside the container as a volume,
+so all changes made to the application code are available inside the container without the need to restart/rebuild it). Useful for the application debug.
```
dmitrii-beliakov, line 11: I'd recommend not to use so long sentences in parentheses. Those are 4 separate thoughts/sentences:
-  builds a container ... 
- App directory is attached ...
- All changes made ...
- Useful for debugging ...

Also, `debug` is a verb. It is better to say: It is useful for ... debugging.

tools/otsdb-dump-restore/README.md
```diff
@@ -60,3 +58,55 @@ Options:
...
+3. Then, we can restore the data:
+```bash
+docker run --rm -v opentsdb-data:/tmp kilda-otsdb-dump-restore kilda-otsdb-restore http://victoria:4242
+```
+4. Remove volume
+```bash
+docker volume remove opentsdb-data
+```
+
+Another option could be a loop that iterates by date range, for instance days. This could be useful when there is much information to migrate.
```
dmitrii-beliakov, line 95: more common collocations are: iterate over a range or iterate through a range.

tools/otsdb-dump-restore/README.md
```diff
@@ -60,3 +58,55 @@ Options:
                                 [default: 4096]
   --help                        Show this message and exit.
 ```
+### Format time
```
dmitrii-beliakov, line 61: Date and time formats. 
`2023-03-22` is a date

tools/otsdb-dump-restore/build-tools/setup-devel-user.sh
dmitrii-beliakov: There is no test for it. I thought it would better for the script to fail explicitly rather than continue for example with an empty UID or GID.

tools/otsdb-dump-restore/README.md
pablomuri: done

tools/otsdb-dump-restore/README.md
pablomuri: done

tools/otsdb-dump-restore/build-tools/setup-devel-user.sh
pablomuri: I was adding some lines to check if the variables are empty. When I tested, it does fail before checking:
Screenshot 2023-03-22 at 16 47 48
I'd say it's enough with this error. If not, I can add some ifs before the argument assignment but maybe it is redundant. What do you think?

tools/otsdb-dump-restore/setup.py
pablomuri: I've added a guide for two possible workflows. Is this what you mean?
Regarding the copyright and the pythondocs, let's discuss them with the team.

tools/otsdb-dump-restore/build-tools/setup-devel-user.sh
dmitrii-beliakov: aah, that's the `set -u` does the magic: 
```
-u  Treat unset variables as an error when substituting.
```

all is good then

tools/otsdb-dump-restore/setup.py
dmitrii-beliakov: yes, that's great now

tools/otsdb-dump-restore/src/kilda/tsdb_dump_restore/dump.py
```diff
@@ -69,11 +77,19 @@ def main(opentsdb_endpoint, time_start, **options):
...
+
+def concurrent_dump(all_metrics_iterator, statistics, client, dump_frame, dump_dir, query_frame_size, need_remove_meta, concurrent_executors):
+    with concurrent.futures.ThreadPoolExecutor(max_workers=concurrent_executors) as executor:
+        futures = []
+        for metric in all_metrics_iterator:
+            futures.append(executor.submit(dump, statistics, client, dump_frame, dump_dir, metric, query_frame_size, need_remove_meta))
+        # concurrent.futures.wait(futures)
+        concurrent.futures.as_completed(futures)
+        # for _ in concurrent.futures.as_completed(futures):
+        #     print("OK")
```
dmitrii-beliakov, line 92: This is for debugging or why it is here?

dmitrii-beliakov (approved):
It complains about the end of line at the end of file in couple of places.

tools/otsdb-dump-restore/src/kilda/tsdb_dump_restore/dump.py
pablomuri: This is just here because I forgot to remove it 😬. I'll remove it

dmitrii-beliakov (approved):

tools/otsdb-dump-restore/README.md
```diff
@@ -0,0 +1,146 @@
...
+```bash
+docker run --rm -v opentsdb-data:/tmp kilda-otsdb-dump-restore kilda-otsdb-restore http://victoria:4242
+```
+4. Finally, we can remove the volume using the following command:
+```bash
+docker volume remove opentsdb-data
+```
+
+### How to use the `otsdb-to-vm.sh` script
+The `otsdb-to-vm.sh` script is a bash wrapper around the `kilda-otsdb-dump-restore` tool. It is used to dump data from an OpenTSDB and restore it to a VictoriaMetrics service.
```
dmitrii-beliakov:
```suggestion
The `otsdb-to-vm.sh` script is a bash wrapper around the `kilda-otsdb-dump-restore` tool. It is used for dumping data from an OpenTSDB and restore it to a VictoriaMetrics service.
```
`is used to` is a special collocation that means "get familiar with`, without to be it means "did something in the past, but not now"

tools/otsdb-dump-restore/README.md
```diff
@@ -0,0 +1,146 @@
...
+### Usage examples:
+```bash
+# pump data using one day batch
+./otsdb-to-vm.sh -s http://opentsdb.example.com:4242 -d http://victoria-metrics.example.com:4242 --start_date 2022-01-01 --end_date 2022-01-31 --metrics_prefix 'kilda.' --interval day
+
+# pump data using one hour batch
+./otsdb-to-vm.sh -s http://opentsdb.example.com:4242 -d http://victoria-metrics.example.com:4242 --start_date 2022-01-01T00:00:00 --end_date 2022-01-01T23:59:59 --metrics_prefix 'kilda.' --interval hour
+```
+
+### How it working:
```
dmitrii-beliakov:
```suggestion
### How it works:
```

dmitrii-beliakov (approved):
just a few more suggestions for the docs

niksv (approved):

niksv:
We already using this script for data migration so I think it is ready to merge

PR #4970, opened 2022-10-31
Test/true y flow sync

artHalMirantis:
test for the appropriate fix (true-y-flow-sync)
delivered separately to avoid all-in-one job rerun


niksv:
@artHalMirantis Are you sure that this PR is about y flow sync? looks like code style improvements

Issue #4890, opened 2022-07-21
Remove  up-test-mode script as redundant 


artHalMirantis:
**up-test-mode** used the env variable which was removed
 and it can be removed and user can use **up-stable** instead

niksv:
Sorry, but some of our scripts still use it. Also @yuliiamir faced with an issue related to this script (there were 2 network topologies because Kilda was run with up-stable instead of up-test-mode). So I have to close the PR.

PR #4974, opened 2022-10-31
#4890: Removed obsolete up-test-mode make script

pkazlenka:
* Replaced up-test-mode make script with up-stable as the ones do the same recently


artHalMirantis (approved):

niksv:
I need to check if some of our scripts works with it

niksv:
Sorry, but some of our scripts still use it. Also @yuliiamir faced with an issue related to this script (there were 2 network topologies because Kilda was run with `up-stable` instead of `up-test-mode`). So I have to close the PR.

PR #4977, opened 2022-11-02
Set lacp_reply to true if value is not specified

### LINKED ISSUE #4976: Create LAG without `lacp_reply` in request causes 500 response
**Steps to reproduce:**
1. send request `POST /api/v2/switches/{switch_id}/lags` with body `{ "port_numbers": [12, 14]}` (without `lacp_reply`)

**Expected result:**
1. lacp_reply will be set to true by default 
2. LAG will be created

**Actual result:**
500 Internal server error

niksv:
Closes #4976


Nexgen90 (approved):
Perfect!

PR #4992, opened 2022-11-10
Increased timeout in unit-tests

Nexgen90:
Sometimes tests are failed by a lack of time


niksv:
I guess we can find better way to wait till topology will ready, but it's okay for temporary solution

PR #4993, opened 2022-11-12
Split huge validation messages

niksv:


src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/Utils.java
```diff
@@ -173,5 +181,71 @@ public static boolean validateSwitchId(SwitchId switchId) {
...
+
+    /**
+     * Splits lists on chunks.
+     */
+    public static <E> List<List<E>> split(List<E> list, int chunkSize) {
+        return split(list, chunkSize, chunkSize);
+    }
+
+    /**
+     * Splits lists on chunks.
```
pkazlenka, line 240: ... with first chunk size different to the rest.

src-java/base-topology/base-messaging/src/test/java/org/openkilda/messaging/info/switches/v2/GroupsValidationEntryV2Test.java
```diff
@@ -0,0 +1,207 @@
...
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class GroupsValidationEntryV2Test {
+
+    @Test
+    public void splitAndUniteEmptyGroupEntryTest() {
+        GroupsValidationEntryV2 entry = GroupsValidationEntryV2.builder()
```
pkazlenka, line 32: Here and in test classes below:
I'd recommend to add inner class GroupsValidationEntryV2Factory, which would return test entries with meaningful names, e.g.
`GroupsValidationEntryV2Factory.getEmpty()` or `GroupsValidationEntryV2Factory.getHugeChunk()`
This will let you:
1) Increase test readability
2) Do not describe same test input twice.

src-java/base-topology/base-messaging/src/test/java/org/openkilda/messaging/info/switches/v2/GroupsValidationEntryV2Test.java
```diff
@@ -0,0 +1,207 @@
...
+                .proper(new ArrayList<>())
+                .build();
+        List<GroupsValidationEntryV2> list = entry.split(4, 4);
+        assertEquals(1, list.size());
+        GroupsValidationEntryV2 united = GroupsValidationEntryV2.unite(list);
+        assertEquals(entry, united);
+    }
+
+    @Test
+    public void splitAndUniteNullGroupEntryTest() {
```
pkazlenka, line 46: Combine the tests into one ParametrizedTest. This would significantly decrease amount of code.

src-java/floodlight-service/floodlight-api/src/main/java/org/openkilda/messaging/info/group/GroupDumpResponse.java
```diff
@@ -26,21 +29,38 @@
...
+    public List<GroupDumpResponse> split(int chunkSize) {
+        return Utils.split(groupSpeakerData, chunkSize).stream()
+                .map(GroupDumpResponse::new)
+                .collect(Collectors.toList());
+    }
+
+    /**
+     * Unites several responses into one.
+     */
+    public static GroupDumpResponse unite(List<GroupDumpResponse> dataList) {
```
pkazlenka, line 58: Can this be moved to Chunkable as default method?

src-java/floodlight-service/floodlight-api/src/main/java/org/openkilda/messaging/info/group/GroupDumpResponse.java
```diff
@@ -26,21 +29,38 @@
...
     @Builder
-    public GroupDumpResponse(@JsonProperty("switch_id") SwitchId switchId,
-                             @JsonProperty("group_speaker_data") List<GroupSpeakerData> groupSpeakerData) {
-        this.switchId = switchId;
+    public GroupDumpResponse(@JsonProperty("group_speaker_data") List<GroupSpeakerData> groupSpeakerData) {
         this.groupSpeakerData = groupSpeakerData;
     }
+
+    @Override
+    public List<GroupDumpResponse> split(int chunkSize) {
```
pkazlenka, line 49: Can this be moved to Chunkable as default method?

src-java/swmanager-topology/swmanager-storm-topology/src/main/java/org/openkilda/wfm/topology/switchmanager/service/impl/SpeakerWorkerService.java
```diff
@@ -148,14 +151,25 @@ private void completeChunkedResponse(String key, List<ChunkedInfoMessage> messag
...
             if (pending.getPayload() instanceof DumpRulesForSwitchManagerRequest) {
-                FlowDumpResponse entries = new FlowDumpResponse(
-                        messages.stream()
-                                .map(InfoMessage::getData)
-                                .map(SingleFlowDumpResponse.class::cast)
-                                .map(SingleFlowDumpResponse::getFlowSpeakerData)
-                                .collect(Collectors.toList()));
-                InfoMessage response = new InfoMessage(entries, key, pending.getCookie());
-                carrier.sendResponse(key, response);
+                List<FlowDumpResponse> responses = messages.stream()
```
pkazlenka, line 154: Seems like a good place for generic method or Strategy pattern.

pkazlenka (changes requested):

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/Utils.java
```diff
@@ -173,5 +181,71 @@ public static boolean validateSwitchId(SwitchId switchId) {
...
+     */
+    public static boolean joinBooleans(Collection<Boolean> booleans) {
+        Set<Boolean> set = booleans.stream().filter(Objects::nonNull).collect(Collectors.toSet());
+        if (set.size() > 1) {
+            throw new IllegalArgumentException(String.format("Stream %s contains true and false booleans. "
+                    + "It must contain only one value of boolean or null values.", booleans));
+        }
+        if (set.isEmpty()) {
+            throw new IllegalArgumentException(String.format("Stream %s has no non-null values.", booleans));
+        }
```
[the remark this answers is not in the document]
niksv: Have common exception like "we have one of two problems" is bad idea. User must know exact problem (if we can return such information).

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/info/switches/v2/GroupsValidationEntryV2.java
```diff
@@ -15,21 +15,71 @@
...
+@AllArgsConstructor
 @JsonNaming(value = SnakeCaseStrategy.class)
 public class GroupsValidationEntryV2 implements Serializable {
     private boolean asExpected;
     private List<GroupInfoEntryV2> excess;
     private List<GroupInfoEntryV2> proper;
     private List<GroupInfoEntryV2> missing;
     private List<MisconfiguredInfo<GroupInfoEntryV2>> misconfigured;
+
+    static GroupsValidationEntryV2 unite(List<GroupsValidationEntryV2> entryList) {
```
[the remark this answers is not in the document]
niksv: renamed

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/info/switches/v2/GroupsValidationEntryV2.java
```diff
@@ -15,21 +15,71 @@
...
+                .collect(Collectors.toList())));
+        builder.excess(joinLists(nonNullList.stream().map(GroupsValidationEntryV2::getExcess)));
+        builder.proper(joinLists(nonNullList.stream().map(GroupsValidationEntryV2::getProper)));
+        builder.missing(joinLists(nonNullList.stream().map(GroupsValidationEntryV2::getMissing)));
+        builder.misconfigured(joinLists(nonNullList.stream().map(GroupsValidationEntryV2::getMisconfigured)));
+        return builder.build();
+    }
+
+    List<GroupsValidationEntryV2> split(int firstChunkSize, int chunkSize) {
+        List<GroupsValidationEntryV2> result = new ArrayList<>();
```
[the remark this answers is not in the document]
niksv: 1. what we should check for nulls? oll checks are made in ValidationEntry.split

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/info/switches/v2/MetersValidationEntryV2.java
```diff
@@ -32,4 +39,44 @@ public class MetersValidationEntryV2 implements Serializable {
...
+                    .excess(entry.getExcess())
+                    .proper(entry.getProper())
+                    .misconfigured(entry.getMisconfigured())
+                    .build());
+        }
+        return result;
+    }
+
+    int size() {
+        return getSize(excess) + getSize(proper) + getSize(missing) + getSize(misconfigured);
```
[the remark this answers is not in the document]
niksv: ValidationEntry is not a parent for MetersValidationEntryV2
so calling of this method will look like
```
int size() {
        return ValidationEntry.getSize(excess, proper, missing, misconfigured);
        }
```

So it wouldn't save as any code lines

src-java/floodlight-service/floodlight-api/src/main/java/org/openkilda/messaging/info/group/GroupDumpResponse.java
niksv: I wish I can, but after trying for several days to do it I failed. Base interface 
1. can't create new instance of child class
2. do not know how to each type of child. 

I tried to add non static unite method to override it in each class, but it bring more problems. 
If you have any idea - I'll be happy to hear it, because I'm run out of ideas :(

src-java/floodlight-service/floodlight-api/src/main/java/org/openkilda/messaging/info/group/GroupDumpResponse.java
niksv: No, because each object is unique. one contains a list inside and the other one contains 4 other objects. Each of them need to have separate split method. 

Moreover to make it default Chunkable must have access to inner field `groupSpeakerData`. But it has no such access because it specific filed for only this class. other classes have other fields

src-java/floodlight-service/floodlight-api/src/main/java/org/openkilda/messaging/info/group/GroupDumpResponse.java
pkazlenka: Fair enough.

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/Utils.java
niksv: added description

src-java/swmanager-topology/swmanager-storm-topology/src/main/java/org/openkilda/wfm/topology/switchmanager/service/impl/SpeakerWorkerService.java
pkazlenka: 1. I don't remember if the one can use generic/parametrized methods without making the class parametrized. If no - then you're right, my first suggestion doesn't make sense.
2. Can't think of a good usage of Strategy here. It was a bad suggestion.

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/info/switches/v2/SwitchValidationResponseV2.java
```diff
@@ -23,14 +25,120 @@
...
     LogicalPortsValidationEntryV2 logicalPorts;
+
+    @Override
+    public List<SwitchValidationResponseV2> split(int chunkSize) {
+        List<SwitchValidationResponseV2> result = new ArrayList<>();
+        SwitchValidationResponseV2Builder current = SwitchValidationResponseV2.builder().asExpected(asExpected);
+        boolean addCurrent = true;
+        int currentSize = chunkSize;
+
+        if (groups != null) {
```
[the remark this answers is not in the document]
niksv: I tried to simplify it but I have one problem. We have global variables here like  `current` `addCurrent` `currentSize`
After each addition we need to update it. How we can do it in sub methods? 

How exactly we can use strategy here? have some parent class for all validation entries?

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/Utils.java
```diff
@@ -173,5 +181,71 @@ public static boolean validateSwitchId(SwitchId switchId) {
...
+            return null;
+        }
+        return nonNullLists.stream().flatMap(Collection::stream).collect(Collectors.toList());
+    }
+
+    /**
+     * Joins several booleans.
+     */
+    public static boolean joinBooleans(Collection<Boolean> booleans) {
+        Set<Boolean> set = booleans.stream().filter(Objects::nonNull).collect(Collectors.toSet());
```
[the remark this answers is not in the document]
niksv: fixed

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/info/switches/v2/SwitchValidationResponseV2.java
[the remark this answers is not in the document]
pkazlenka: >  We have global variables here like current addCurrent currentSize
In theory, we can use inner class like SwitchValidationResponseTemporaryContext and pass, update it and return between methods. But I doubt that this would significantly increase the readability of code.

src-java/base-topology/base-messaging/src/test/java/org/openkilda/messaging/info/switches/v2/GroupsValidationEntryV2Test.java
niksv: I'm not sure about readability. When you see name "empty" you do not know what does it mean (e.g. excess rules are null or empty list). What "huge" means? 10? 100? 1000? to check both names you will go to factory, at the end of the file, then you will go back to check chunk sizes and expected chunk count, then you will go to the end of file again to check if this huge object must be split on exactly X chunks or not.

At this moment test has 10 lines. You can see all of them in one screen. I think it has better readability.

src-java/base-topology/base-messaging/src/test/java/org/openkilda/messaging/info/switches/v2/GroupsValidationEntryV2Test.java
niksv: what kind of parameters will we have?
list of arguments like

```
static Stream<Arguments> arguments = Stream.of(
  Arguments.of(LogicalPortsValidationEntryV2.builder()
                .asExpected(true)
                .missing(buildLogicalPortsInfo(1, 500))
                .excess(buildLogicalPortsInfo(1000, 600))
                .proper(buildLogicalPortsInfo(2000, 700))
                .misconfigured(buildMisconfiguredLogicalPortsInfo(3000, 800))
                .build(), 4, 3, 5), // object, first chunk, other chunks, chunk count
  Arguments.of(LogicalPortsValidationEntryV2.builder()
                .asExpected(true)
                .missing(buildLogicalPortsInfo(1, 500))
                .excess(buildLogicalPortsInfo(1000, 600))
                .proper(buildLogicalPortsInfo(2000, 700))
                .misconfigured(buildMisconfiguredLogicalPortsInfo(3000, 800))
                .build(), 4, 3, 5), // object, first chunk, other chunks, chunk count
 ****);
```

looks not very readable.


same list, but with factory methods?

```
 Arguments.of(getHugeObject, 4, 3, 5), // object, first chunk, other chunks, chunk count
 Arguments.of(getEmptyObject, 4, 3, 5), // object, first chunk, other chunks, chunk count
``` 

Better, but again, not very readable.  You need to jump to test, factory and parameters to understand what is going on.
getEmptyObject and other methods will be used only one time. Why we need to create a method for it? Usually we create methods when code will be called several times or to reduce other method size. But test method has 10 lines size

Nexgen90 (approved):

src-java/base-topology/base-messaging/src/test/java/org/openkilda/messaging/info/switches/v2/GroupsValidationEntryV2Test.java
pkazlenka: I'm sorry, but logic 'if it fits my screen size, breaking it would worsen readability' is wrong. Methods must be minimal logical units, not minimal screen size code units.

The readable naming rule for these particular test is quite simple, IMHO. What you have in test name is what your input name should be, e.g.:
```
@Test
void splitAndUniteEmptyGroupEntryTest() {
    expectedGroup = GroupsValidationEntryV2Factory.getEmpty()
...
@Test
void splitAndUniteGroupEntryWithMalformedObjectsTest() {
     expectedGroup = GroupsValidationEntryV2Factory.getGroupWithMalformedObjects()
... 
etc.
```
> What "huge" means? 10? 100? 1000? 
It doesn't matter, which amount of objects you defined as 'huge'. The goal is to understand what your test data tries to mimic (big array, corrupted archive, unchunkable amount of bytes, signal from Mars, whatever). Also it doesn't matter what you have in other fields of test data object, so you can hide these details under factory method name (if you don't want to describe a factory, make it Enum).

src-java/base-topology/base-messaging/src/test/java/org/openkilda/messaging/info/switches/v2/GroupsValidationEntryV2Test.java
pkazlenka: Yes, something like option #2 is what I was talking about. It is a java problem that you can not describe test input with variable names, unlike, e.g., kotlin. But with complete code it would look better:
```
@ParametrizedTest(Stream.of(
Arguments.of(getHugeObject, 4, 3, 5),
Arguments.of(getObjectWhichWontSplitToEqualParts, 5, 6, 7),
...
Arguments.of(getCorruptedListsObject, 1, 0, null)))
void splitAndUniteTest(GroupsValidationEntryV2 entryToSplit, int firstChunk, int otherChunks, int expectedChunksAmount) {...
```
> You need to jump to test, factory and parameters to understand what is going on.

You won't if your naming would be clear. Test arguments with description are below the list of arguments, on the same screen, within amount of lines equal to number of your tests (i.e. if you have 6 tests, you'll have parameters description `GroupsValidationEntryV2 entryToSplit, int firstChunk, int otherChunks, int expectedChunksAmount` within max 6 lines from your 'unreadable' parameter).

> getEmptyObject and other methods will be used only one time

I agree, that during review I mistakenly assumed that you use almost each GroupsValidationEntryV2 several times. But it says about bad readability of your approach. How much time do you need to realize, why do you need a different tests for these objects:
```
TypicalSwitch.builder()
.setPorts(24)
.setIsls(4)
.setManufacturer("Cisco")
.setMaxDowntime(2400)
.setConsumedPower(500)
.build()
vs
TypicalSwitch.builder()
.setPorts(4)
.setIsls(2)
.setManufacturer("Cisco")
.setMaxDowntime(3400)
.setConsumedPower(10)
.build()
```
?
Now, how much time do you need to realize, why do you need a different tests for these objects:
```
Switches.WITH_SHORT_DOWNTIME
vs
Switches.WITH_LONG_DOWNTIME
```
?

pkazlenka (approved):

PR #5002, opened 2022-11-24
Simplified and partially refactored tests in MultitableFlowsSpec

pkazlenka:
* Two long tests were split to 5 smaller ones
* Common code parts moved to methods (for now inside test class)

### DIFF
--- a/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
+++ b/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
@@ -77,9 +77,9 @@ class MultitableFlowsSpec extends HealthCheckSpecification {
 
         and: "Involved switches pass switch validation"
         "validate switches"(involvedSwitches)
-        assert "is flow pingable?"(flow), FLOW_ISNT_PINGABLE_ASSERTION_MESSAGE
 
-        and: "The flow allows traffic"
+        and: "The flow is pingable and allows traffic"
+        assert "is flow pingable?"(flow), FLOW_ISNT_PINGABLE_ASSERTION_MESSAGE
         assert "does flow pass traffic?"(flow), FLOW_DOESNT_PASS_TRAFFIC_ASSERTION_MESSAGE
 
         when: "Delete flow"
@@ -107,6 +107,7 @@ class MultitableFlowsSpec extends HealthCheckSpecification {
         given: "A potential flow on a path of 4 switches: multi -> single -> multi -> single"
         def (involvedSwitches, swPair, desiredPath, initSwProps) =
         "get a potential flow on a path of 4 switches: multi and single"()
+
         when: "Create the prepared hybrid flow"
         def flow = flowHelperV2.randomFlow(swPair)
         flowHelperV2.addFlow(flow)
@@ -161,9 +162,11 @@ class MultitableFlowsSpec extends HealthCheckSpecification {
     @Tidy
     @Tags([LOW_PRIORITY])
     def "System can hold flows of different table-mode types"() {
-        given: "A potential flow on a path of 4 switches: multi -> single -> multi -> single"
+        given: "A switch pair in single-table mode"
         def (involvedSwitches, swPair, desiredPath, initSwProps) =
-        "get a potential flow on a path of 4 switches: multi and single"()
+        "get triplet of traffgen switches in single table mode with flow path"()
+        ArrayList<Switch> switches = [swPair.getSrc(), swPair.getDst()]
+
         when: "Create the prepared hybrid flow"
         def flow = flowHelperV2.randomFlow(swPair)
         flowHelperV2.addFlow(flow)
@@ -172,7 +175,14 @@ class MultitableFlowsSpec extends HealthCheckSpecification {
         then: "Created flow is valid"
         northbound.validateFlow(flow.flowId).each { assert it.asExpected }
 
-        when: "Create one more similar flow on the target path"
+        when: "Set switches to multi-table mode"
+        withPool {
+            involvedSwitches.eachParallel {
+                northbound.updateSwitchProperties(it.dpId, changeSwitchPropsMultiTableValue(initSwProps[it.dpId], true))
+            }
+        }
+
+        and: "Create one more similar flow on the target path"
         def flow2 = flowHelperV2.randomFlow(swPair).tap {
             it.source.portNumber = flow.source.portNumber
             it.source.vlanId = flow.source.vlanId - 1
@@ -180,37 +190,48 @@ class MultitableFlowsSpec extends HealthCheckSpecification {
             it.destination.vlanId = flow.destination.vlanId - 1
         }
         flowHelperV2.addFlow(flow2)
+        def flows = [flow, flow2]
 
         then: "Both existing flows are valid"
-        [flow, flow2].each {
-            northbound.validateFlow(it.flowId).each { assert it.asExpected }
+        withPool {
+            flows.eachParallel {
+                northbound.validateFlow(it.flowId).each { assert it.asExpected }
+            }
         }
+
+        and: "First flow is in single table and the second is in multiple"
+        assert "flow rules are in tables"(swPair.getSrc(), flow) == [SINGLE_TABLE_ID]
+        assert "flow rules are in tables"(involvedSwitches[1], flow) == [SINGLE_TABLE_ID]
+        assert "flow rules are in tables"(swPair.getDst(), flow) == [SINGLE_TABLE_ID]
+        assert "flow rules are in tables"(swPair.getSrc(), flow2) == [INGRESS_RULE_MULTI_TABLE_ID, EGRESS_RULE_MULTI_TABLE_ID]
+        assert "flow rules are in tables"(involvedSwitches[1], flow2) == [TRANSIT_RULE_MULTI_TABLE_ID]
+        assert "flow rules are in tables"(swPair.getSrc(), flow2) == [INGRESS_RULE_MULTI_TABLE_ID, EGRESS_RULE_MULTI_TABLE_ID]
 
         and: "Involved switches pass switch validation"
         "validate switches"(involvedSwitches)
 
         and: "Both flows are pingable and allow traffic"
         // This hangs (probably, blocks each other) if run in parallel mode
-        [flow, flow2].each {
+        flows.each {
             assert "is flow pingable?"(it), FLOW_ISNT_PINGABLE_ASSERTION_MESSAGE
             assert "does flow pass traffic?"(it), FLOW_DOESNT_PASS_TRAFFIC_ASSERTION_MESSAGE
         }
 
         when: "Delete flows"
         def flowsCookies = getFlowCookies(swPair, flow).addAll(getFlowCookies(swPair, flow2))
-        [flow, flow2].each { flowHelperV2.deleteFlow(it.flowId) }
+        flows.each { flowHelperV2.deleteFlow(it.flowId) }
         def flowsAreDeleted = true
 
         then: "Flow rules are deleted from switches"
-        involvedSwitches.each { sw ->
+        switches.each { sw ->
             with(northbound.getSwitchRules(sw.dpId).flowEntries) { rules ->
                 rules.findAll { it.cookie in flowsCookies }.empty
             }
         }
 
         cleanup: "Revert system to original state"
-        !flowsAreDeleted && [flow, flow2].each { it && flowHelperV2.deleteFlow(it.flowId) }
-        initSwProps && revertSwitchesToInitState(involvedSwitches, initSwProps)
+        !flowsAreDeleted && flows.each { it && flowHelperV2.deleteFlow(it.flowId) }
+        initSwProps && revertSwitchesToInitState(switches, initSwProps)
         northbound.deleteLinkProps(northbound.getLinkProps(topology.isls))
     }
 
@@ -253,12 +274,50 @@ class MultitableFlowsSpec extends HealthCheckSpecification {
 
         where:
         eventName           | event
-        // https://github.com/telstra/open-kilda/issues/4865
-//        "synchronize flow"  | {FlowRequestV2 testFlow ->
-//            with(northbound.synchronizeFlow(testFlow.flowId)) {!it.rerouted}}
+        /* https://github.com/telstra/open-kilda/issues/4865
+        "synchronize flow"  | {FlowRequestV2 testFlow ->
+            with(northbound.synchronizeFlow(testFlow.flowId)) {!it.rerouted}} */
         "update flow"       | {FlowRequestV2 testFlow ->
             flowHelperV2.updateFlow(testFlow.flowId,
                     testFlow.tap { it.description = it.description + " updated" })}
+    }
+
+    @Tidy
+    @Tags([LOW_PRIORITY])
+    def "Single-switch flow rules are not updated when try to reroute the flow"() {
+        given: "An active switch"
+        def sw = topology.activeSwitches.find { it.features.contains(SwitchFeature.MULTI_TABLE) }
+        SwitchPropertiesDto initSwProps = switchHelper.getCachedSwProps(sw.dpId)
+
+        and: "Multi table mode is disabled on it"
+        if (initSwProps.multiTable) {
+            northbound.updateSwitchProperties(sw.dpId, changeSwitchPropsMultiTableValue(initSwProps, false))
+        }
+        checkDefaultRulesOnSwitch(sw)
+
+        when: "Create a single-switch flow"
+        def flow = flowHelperV2.singleSwitchFlow(sw)
+        flowHelperV2.addFlow(flow)
+
+        then: "Flow rules are created in single table mode"
+        assert "flow rules are in tables"(sw, flow) == [SINGLE_TABLE_ID]
+
+        and: "Flow is valid"
+        northbound.validateFlow(flow.flowId).each { direction -> assert direction.asExpected }
+
+        when: "Update switch properties(multi_table: true) on the switch"
+        northbound.updateSwitchProperties(sw.dpId, changeSwitchPropsMultiTableValue(initSwProps, true))
+
+        and: "reroute the flow"
+        with(northboundV2.rerouteFlow(flow.flowId)) { !it.rerouted }
+
+        then: "Flow rules stay in single table mode"
+        wait(RULES_INSTALLATION_TIME/5) {
+            "flow rules are in tables"(sw, flow) == [SINGLE_TABLE_ID]
+        }
+        cleanup:
+        flowHelperV2.deleteFlow(flow.flowId)
+        initSwProps && revertSwitchToInitState(sw, initSwProps)
     }
 
     @Tags([LOW_PRIORITY])
@@ -1021,10 +1080,10 @@ class MultitableFlowsSpec extends HealthCheckSpecification {
         northbound.validateFlow(flow.flowId).each { direction -> assert !direction.asExpected }
 
         and: "Rule info is moved into the 'missing' section on the transit and dst switches"
-        verifyAll(northbound.validateSwitch(involvedSwitches[2].dpId)) { validateInfo ->
+        verifyAll(northboundV2.validateSwitch(involvedSwitches[2].dpId)) { validateInfo ->
             validateInfo.rules.excess.empty
             validateInfo.rules.misconfigured.empty
-            validateInfo.rules.missing.sort() ==
+            validateInfo.rules.missing.cookie.sort() ==
                     [flowInfoFromDb.forwardPath.cookie.value, flowInfoFromDb.reversePath.cookie.value, sharedRuleOnDstSwitch.cookie].sort()
             validateInfo.meters.excess.empty
             validateInfo.meters.missing.empty
@@ -1079,7 +1138,7 @@ class MultitableFlowsSpec extends HealthCheckSpecification {
 
         and: "All involved switches pass switch validation"
         involvedSwitches.each { sw ->
-            with(northbound.validateSwitch(sw.dpId)) { validation ->
+            with(northboundV2.validateSwitch(sw.dpId)) { validation ->
                 validation.verifyRuleSectionsAreEmpty(["missing", "excess"])
                 validation.verifyMeterSectionsAreEmpty(["missing", "excess", "misconfigured"])
             }
@@ -1135,14 +1194,14 @@ class MultitableFlowsSpec extends HealthCheckSpecification {
 
         then: "System detects  excess rules on the src and transit switches"
         Map<SwitchId, SwitchValidationExtendedResult> validationResultsMap = involvedSwitches.collectEntries {
-            [it.dpId, northbound.validateSwitch(it.dpId)]
+            [it.dpId, northboundV2.validateSwitch(it.dpId)]
         }
         involvedSwitches[0..1].each {
             assert validationResultsMap[it.dpId].rules.missing.empty
             assert validationResultsMap[it.dpId].rules.misconfigured.empty
             assert !validationResultsMap[it.dpId].rules.excess.empty
         }
-        with(northbound.validateSwitch(involvedSwitches[2].dpId)) { validation ->
+        with(northboundV2.validateSwitch(involvedSwitches[2].dpId)) { validation ->
             validationResultsMap[involvedSwitches[2].dpId].rules.missing.empty
             validationResultsMap[involvedSwitches[2].dpId].rules.misconfigured.empty
             validationResultsMap[involvedSwitches[2].dpId].rules.excess.empty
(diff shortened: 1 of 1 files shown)


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
```diff
@@ -54,43 +57,56 @@ class MultitableFlowsSpec extends HealthCheckSpecification {
...
+        def flow = flowHelperV2.randomFlow(swPair)
+        flowHelperV2.addFlow(flow)
+        assert PathHelper.convert(northbound.getFlowPath(flow.flowId)) == desiredPath
+
+        then: "Created flow is valid"
+        northbound.validateFlow(flow.flowId).each { assert it.asExpected }
+
+        and: "Involved switches pass switch validation"
+        "validate switches"(involvedSwitches)
+        assert "is flow pingable?"(flow), FLOW_ISNT_PINGABLE_ASSERTION_MESSAGE
```
niksv, line 80: This line is not about `Involved switches pass switch validation`. I think you should move it to section `and: "Flow is pingable"` (like in next test)

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
```diff
@@ -136,17 +134,43 @@ mode with existing flows and hold flows of different table-mode types"() {
...
+        northbound.deleteLinkProps(northbound.getLinkProps(topology.isls))
+    }
+
+    @Tidy
+    @Tags([LOW_PRIORITY])
+    def "System can hold flows of different table-mode types"() {
+        given: "A potential flow on a path of 4 switches: multi -> single -> multi -> single"
+        def (involvedSwitches, swPair, desiredPath, initSwProps) =
+        "get a potential flow on a path of 4 switches: multi and single"()
+        when: "Create the prepared hybrid flow"
```
niksv, line 167: I think it's a good practice to have empty line between `geven`, `when`, `then`, `and`, 'cleanup'

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
```diff
@@ -136,17 +134,43 @@ mode with existing flows and hold flows of different table-mode types"() {
...
+        def flow = flowHelperV2.randomFlow(swPair)
+        flowHelperV2.addFlow(flow)
+        assert PathHelper.convert(northbound.getFlowPath(flow.flowId)) == desiredPath
+
+        then: "Created flow is valid"
+        northbound.validateFlow(flow.flowId).each { assert it.asExpected }
```
niksv, line 173: CRITICAL: Before creating second flow you must change mode of switches with

`when: "Update table mode for involved switches so that it becomes 'single -> multi -> single -> multi'"`

Because name of the test is `System can hold flows of different table-mode types` but you have 2 flows with same table mode types

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
```diff
@@ -218,156 +216,49 @@ mode with existing flows and hold flows of different table-mode types"() {
...
-            rules.find { it.cookie == flowInfoFromDb3.forwardPath.cookie.value }.tableId == INGRESS_RULE_MULTI_TABLE_ID
-            rules.find { it.cookie == flowInfoFromDb3.reversePath.cookie.value }.tableId == INGRESS_RULE_MULTI_TABLE_ID
-            rules.findAll { it.cookie in sharedRules*.cookie }*.tableId.unique() == [SHARED_RULE_TABLE_ID]
-        }
-
-        when: "Reroute(intentional) the flow via APIv2"
-        with(northboundV2.rerouteFlow(flow.flowId)) { !it.rerouted }
-        wait(rerouteDelay + WAIT_OFFSET) {
-            assert northboundV2.getFlowStatus(flow.flowId).status == FlowState.UP
-        }
```
niksv, line 347: Even if rerouting of one switch flow has no sense we must be sure that this operation wouldn't crash the system and that response is valid. So we should not remove this test case

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
```diff
@@ -1510,4 +1388,94 @@ mode with existing flows and hold flows of different table-mode types"() {
             it.multiTable = newValue
         }
     }
+
+    //TODO: move to PathFinder."get a potential flow on a path of 4 switches: multi and single"() or something like this
```
niksv, line 1392: PathFinder is interface, which is used for finding paths for application, not for testing. it must NOT contain any methods for finding test data.

niksv (changes requested):

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
pkazlenka: In fact, I believe that empty lines inside methods are the marks of bad design. Developer feels, that the method parts do several different things and tries to split the ones by empty line. Instead, in this case, one should split the method to several submethods with clear names.
But as long as it is convention of our project, I'll preserve the empty lines between steps.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
pkazlenka: Well, test name is obviously ambiguous. First of all, if I get it right, only switches can be in multi/single table mode, not flows. And idea behind the test is to check if there can be flow which goes through both single and multi table mode.
Second, if the deal is only about checking if the first flow rules can be placed in several tables, and the rules of the second flow can be in table '0', why do we need four switches for verification?
Do you have a guess what this test was expected to verify?

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
pkazlenka: Do we need to test this on environment for 2 minutes or this can be verified on unit tests level?

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
pkazlenka: Here and below: obviously, I'm not going to add handy test methods to the application code. The fact that we use code from application in tests is a big problem.
First of all, we need to clone all the currently used code from application to test package (so tests become independent from the tested application).
Then we can extend current models, making then something bigger than just dataclasses. In this particular case I didn't know that there is interface named PathFinder in openkilda. I just thought, what would be readable class and method name for the code that is looking for an appropriate path, it was something like `PathFinder.getWhatINeed()`. It's not org.openkilda.java-src.floodlight...Pathfinder.getWhatINeed().
Overall, I see our tests (at the specs level) as human-readable stuff without calling NB, FL or reading kafka. It should be something like 
```
SwitchPair switchPair = SwitchPairFactory.getNeighboringSwitches()
Flow flow = FlowFactory.getFlowBetween(switchPair)
flow.getToDegradedState()
assert flow.getState() == DEGRADED
```

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
```diff
@@ -1510,4 +1388,94 @@ mode with existing flows and hold flows of different table-mode types"() {
...
+            northbound.updateSwitchProperties(it.dpId, changeSwitchPropsMultiTableValue(initSwProps[it.dpId], false))
+        }
+        checkDefaultRulesOnSwitches(involvedSwitches)
+        return [involvedSwitches, swPair, desiredPath, initSwProps]
+    }
+
+    //TODO: move to wrapper class SwitchList.getGarbageRules()
+    def "validate switches"(List<Switch> switchesToValidate) {
+        switchesToValidate.each {sw ->
+            with(northbound.validateSwitch(sw.dpId)) { validation ->
```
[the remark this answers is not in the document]
pkazlenka: For now - yes. But this seems to be a big problem, as, in theory, we should run tests on both versions of API.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
```diff
@@ -1510,4 +1388,94 @@ mode with existing flows and hold flows of different table-mode types"() {
...
+                validation.verifyRuleSectionsAreEmpty(["missing", "excess", "misconfigured"])
+                validation.verifyHexRuleSectionsAreEmpty(["missingHex", "excessHex", "misconfiguredHex"])
+                validation.verifyMeterSectionsAreEmpty(["missing", "excess", "misconfigured"])
+            }
+        }
+
+    }
+
+    //TODO: move to Flow.isPingable()
```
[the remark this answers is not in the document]
pkazlenka: See the response above. Flow must be class describing data flow business entity in tests, not data model from application here.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
```diff
@@ -1510,4 +1388,94 @@ mode with existing flows and hold flows of different table-mode types"() {
...
+    def "is flow pingable?"(FlowRequestV2 flow) {
```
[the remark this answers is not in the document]
pkazlenka: Handy, isn't it?

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
niksv: Not only of out project. Spock framework itself divides blocks with empty lines 
https://spockframework.org/spock/docs/1.1/all_in_one.html 
section: Specifications as Documentation

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
niksv: Flow can be it single or multi mode too. We nad to do it because of some restrictions. To change mode of flow you need to
1. updated switch mode
2. reroute flow.

The this test case must check that two flows in different modes can live on one switch. 

How it works:
You find switches with following modes
`multi -> single -> multi -> single`

You create a flow. it has following modes on switches
`multi -> single -> multi -> single`

You change switch modes to 
`single -> multi -> single -> multi`
but first flow still has previous mode because nobody rerouted/updated it

you create second flow and get two flows with following modes

first:
`multi -> single -> multi -> single`
second:
`single -> multi -> single -> multi`

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
niksv: We can't verify it in unit tests because in positive scenario it requires 2 components: north bound, flowhs topology, in negative scenario in requires 3 components: north bound, flowhs topology, floodlight

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
niksv: If you want to add some classes only for testing - go ahead. I thought that you wanted to change application classes

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
pkazlenka: Yes, unneeded empty lines is a common bad practice, not only in groovy or spock.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
pkazlenka: Ok, I've got the picture. I think we don't need 4 switches to verify this option then. It will be enough to have two:
1. single <-> single
2. add flow
3. multi-multi
4. add 2nd flow
Because we've already verified that flows work ok on path single -> multi -> single...

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
pkazlenka: Will be done

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
niksv: No, we still need 3-4 switches. 
flow has ingress switch (src), transit switch (middle), egress switch (dst). 
Kilda install different rules for ingress/egress/transit switches so we have to check all cases.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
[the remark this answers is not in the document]
pkazlenka: Done

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
pkazlenka: Updated this test for 3-switches setup + added test for one-switch flow rerouting.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
pkazlenka: Done

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
```diff
@@ -1510,4 +1447,130 @@ mode with existing flows and hold flows of different table-mode types"() {
...
+        checkDefaultRulesOnSwitches(involvedSwitches)
+        return [involvedSwitches, swPair, desiredPath, initSwProps]
+
+    }
+    //TODO: move to wrapper class SwitchList.getGarbageRules()
+    def "validate switches"(List<Switch> switchesToValidate) {
+        switchesToValidate.each {sw ->
+            with(northboundV2.validateSwitch(sw.dpId)) { validation ->
+                validation.verifyRuleSectionsAreEmpty(["missing", "excess", "misconfigured"])
+                validation.verifyMeterSectionsAreEmpty(["missing", "excess", "misconfigured"])
```
niksv, line 1529: Note: in V2 validation we have field `asExpected`. If it's true - everything is ok on a switch. you can you this field

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
```diff
@@ -136,237 +135,188 @@ mode with existing flows and hold flows of different table-mode types"() {
...
-        def flowInfoFromDb3
+        then: "Flow rules go to multi table mode"
         wait(RULES_INSTALLATION_TIME) {
-            flowInfoFromDb3 = database.getFlow(flow.flowId)
-            verifyAll(northbound.getSwitchRules(sw.dpId).flowEntries) { rules ->
-                rules.find { it.cookie == flowInfoFromDb3.forwardPath.cookie.value }.tableId == INGRESS_RULE_MULTI_TABLE_ID
-                rules.find { it.cookie == flowInfoFromDb3.reversePath.cookie.value }.tableId == INGRESS_RULE_MULTI_TABLE_ID
-                rules.findAll { it.cookie in sharedRules*.cookie }*.tableId.unique() == [SHARED_RULE_TABLE_ID]
-            }
+            "flow rules are in tables"(sw, flow) == [INGRESS_RULE_MULTI_TABLE_ID]
```
niksv, line 269: What about section `and: "Flow is valid and UP"` after updating?
Why it was removed? Flow can be down after update, even if rules are in multi_table_mode

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
```diff
@@ -136,237 +135,188 @@ mode with existing flows and hold flows of different table-mode types"() {
...
-        northbound.getSwitchRules(sw.dpId).flowEntries.findAll {
-            def cookie = new Cookie(it.cookie)
-            cookie.type == CookieType.MULTI_TABLE_INGRESS_RULES || !cookie.serviceFlag
-        }.empty
+        and: "reroute the flow"
+        with(northboundV2.rerouteFlow(flow.flowId)) { !it.rerouted }
 
+        then: "Flow rules stay in single table mode"
+        wait(RULES_INSTALLATION_TIME/5) {
+            "flow rules are in tables"(sw, flow) == [SINGLE_TABLE_ID]
```
niksv, line 316: why `/5` ?
What if, because of bug, rules will be actually reinstalled after `RULES_INSTALLATION_TIME` but we will miss it because we waited not enough?

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
niksv: It still seems that you mean interface `org.openkilda.pce.finder.PathFinder` because the is no any other class with same name in our project. New guy will try to find a class with name `PathFinder` and will move code there to fix this TODO. 
Please name or note that you mean some test class.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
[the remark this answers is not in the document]
niksv: same with class name

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
```diff
@@ -1510,4 +1388,94 @@ mode with existing flows and hold flows of different table-mode types"() {
...
+
+    }
+
+    //TODO: move to Flow.isPingable()
+    def "is flow pingable?"(FlowRequestV2 flow) {
+        def pingResponse = northbound.pingFlow(flow.flowId, new PingInput(null))
+        return pingResponse.forward.pingSuccess && pingResponse.reverse.pingSuccess
+    }
+
+    //TODO: move to Flow.isTrafficPass()
```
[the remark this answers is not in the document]
niksv: same

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
```diff
@@ -1510,4 +1388,94 @@ mode with existing flows and hold flows of different table-mode types"() {
...
+            [flowExam.forward, flowExam.reverse].parallel.map { direction ->
+                def resources = traffExam.startExam(direction)
+                direction.setResources(resources)
+                return traffExam.waitExam(direction).hasTraffic()
+            }.collection
+        }
+        return result.every() //is true
+    }
+
+    //TODO: move to Flow.getCookies()
```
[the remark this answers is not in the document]
niksv: same

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
```diff
@@ -136,237 +135,188 @@ mode with existing flows and hold flows of different table-mode types"() {
...
-        initSwProps && revertSwitchesToInitState(involvedSwitches, initSwProps)
+        !flowsAreDeleted && flows.each { it && flowHelperV2.deleteFlow(it.flowId) }
+        initSwProps && revertSwitchesToInitState(switches, initSwProps)
         northbound.deleteLinkProps(northbound.getLinkProps(topology.isls))
     }
 
     @Tidy
     @Tags([LOW_PRIORITY])
-    def "Single-switch flow rules are (re)installed according to switch property while rerouting,syncing,updating"() {
+    def "Single-switch flow rules are updated single- to multi-table mode when #eventName"() {
```
niksv, line 240: so previous test also tested case when we changed table mode from multi table mode to single. But new test tests only single to multi. We lost test coverage

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
```diff
@@ -136,237 +135,188 @@ mode with existing flows and hold flows of different table-mode types"() {
...
+        SwitchPropertiesDto initSwProps = switchHelper.getCachedSwProps(sw.dpId)
 
-        then: "Flow rules on the switch are not recreated in single table mode because flow wasn't rerouted"
-        verifyAll(northbound.getSwitchRules(sw.dpId).flowEntries) { rules ->
-            rules.find { it.cookie == flowInfoFromDb3.forwardPath.cookie.value }.tableId == INGRESS_RULE_MULTI_TABLE_ID
-            rules.find { it.cookie == flowInfoFromDb3.reversePath.cookie.value }.tableId == INGRESS_RULE_MULTI_TABLE_ID
-            rules.findAll { it.cookie in sharedRules*.cookie }*.tableId.unique() == [SHARED_RULE_TABLE_ID]
+        and: "Multi table mode is disabled on it"
+        if (initSwProps.multiTable) {
+            northbound.updateSwitchProperties(sw.dpId, changeSwitchPropsMultiTableValue(initSwProps, false))
```
niksv, line 294: you changed test case. Prev test checked multi -> single changed. This tests single -> multi. It this particular case it ok, because reroute must not change mode in any case. But As you named PR as "refactoring" you shouldn't change case coverage

niksv (changes requested):

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
pkazlenka: Updated this method. Didn't go through all the tests though.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
pkazlenka: Returned this validation.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
pkazlenka: Originally there was `/3`, but you are right, we should wait for the whole time.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/multitable/MultitableFlowsSpec.groovy
pkazlenka: We have checks on 'multi -> single' switch transition in other tests, e.g. `"Flow rules are re(installed) according to switch props while syncing switch and rules"`. So while the coverage of this test is decreased, overall tests coverage stays effectively the same.

niksv (approved):

Issue #3849, opened 2020-11-13
Flow traffic mirroring


timofei-durakov:
# As an admin I can mirror traffic on a given flow and send it to a specific switch+port+vlan.

`Mirroring point` is a switch the flow that duplicates traffic to receiver. There could be as many mirroring points as # of switches that belong to the flow. 
`Receiver point` is a specific endpoint that listen packets from `mirroring point`.

# API proposal:
- `PUT /flows/{flow_id}/mirror` 
payload:
```
{
  `flow_id`: string,
  `points`: [
  {
      `mirror_point`: sw_id,
      `mirror_direction`: `string [FORWARD|REVERSE]`
     `receiver_point`: {
         `sw_id`: string,
         `port`: int,
         `vlan`: int
     }
 }]
}
```
# Implementation details:
if receiver is not located on the flow endpoint, new additional paths should be build to provide ability sending the traffic. There 2 options for that:
- special paths for the flow
- new service aggregate that consist of the original flow and, special flows that provide mirrors
# Poc details: 
 - receiver was not on an endpoint,
 - mirroring was done for the endpoint of the flow
 - extra flows were created to deliver traffic from flow endpoint to receiver
 - ingress/egress rules were over-written with output to group instead of port.
 - group consists of 2 buckets,
  - first outputs traffic as original flow(customer port/isl port)
  - second re-tag is with extra flow transit-vlan and send it to extra flow next hop
extra flows were used in headless mode(no ingress rules were involved)

dpoltavets:
**Estimates for the first part of this feature** (flow traffic mirroring only at the flow endpoint to another port of this endpoint switch):
- [ ] Floodlight (mirror groups and mirror rules) – 1-2w 
- [ ] Flow HS Topology (create and delete mirror points, update flow endpoint improvement) – 1-2w 
- [ ] Switch and flow validation – 1w
- [ ] Update the current API by adding information about new paths – 1w
- [ ] Nbworker (get mirror points) - 1-2d

niksv:
**Estimates for the second part of this feature** (flow traffic mirroring from flow endpoint to any other endpoint):
- [ ] Rule manager (ingress/transit/egress rules) – 1-2w 
- [ ] Flow HS Topology (create and delete mirror paths) – 1-2w 
- [ ] Switch and flow validation – 1w
- [ ] Update the current API by adding information about new paths – 1w
- [ ] Update reroute topology to reroute mirror paths 1w

PR #5004, opened 2022-11-24
Long mirrors Part 5: Added long mirrors

niksv:
Related to #3849

### DIFF
--- a/docker/db-migration/migrations/026-add-flow-mirror-reverse-path.yaml
+++ b/docker/db-migration/migrations/026-add-flow-mirror-reverse-path.yaml
@@ -1,0 +1,15 @@
+databaseChangeLog:
+  - changeSet:
+      id: tag
+      author: snikitin
+      changes:
+        - tagDatabase:
+            tag: 026-add-flow-mirror-reverse-path
+
+  - changeSet:
+      id: add_flow_mirror_reverse_property
+      author: snikitin
+      changes:
+        - sql: "CREATE PROPERTY flow_mirror.reverse_path_id IF NOT EXISTS STRING"
+      rollback:
+        - sql: "DROP PROPERTY flow_mirror.reverse_path_id IF EXISTS"
--- a/src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/service/SpeakerFlowSegmentRequestBuilder.java
+++ b/src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/service/SpeakerFlowSegmentRequestBuilder.java
@@ -33,14 +33,18 @@ import org.openkilda.floodlight.model.RulesContext;
 import org.openkilda.messaging.MessageContext;
 import org.openkilda.model.Flow;
 import org.openkilda.model.FlowEncapsulationType;
+import org.openkilda.model.FlowMirror;
+import org.openkilda.model.FlowMirrorPath;
 import org.openkilda.model.FlowMirrorPoints;
 import org.openkilda.model.FlowPath;
 import org.openkilda.model.FlowPathDirection;
 import org.openkilda.model.FlowTransitEncapsulation;
 import org.openkilda.model.IslEndpoint;
+import org.openkilda.model.MacAddress;
 import org.openkilda.model.MeterConfig;
 import org.openkilda.model.MirrorConfig;
 import org.openkilda.model.MirrorConfig.MirrorConfigData;
+import org.openkilda.model.MirrorConfig.PushVxlan;
 import org.openkilda.model.NetworkEndpoint;
 import org.openkilda.model.PathId;
 import org.openkilda.model.PathSegment;
@@ -57,6 +61,7 @@ import org.openkilda.wfm.topology.flowhs.service.FlowSegmentRequestFactoriesSequ
 
 import com.fasterxml.uuid.Generators;
 import com.fasterxml.uuid.NoArgGenerator;
+import com.google.common.collect.Lists;
 import lombok.NonNull;
 import lombok.Value;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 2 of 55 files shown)


docker/db-migration/migrations/026-add-flow-mirror-reverse-path.yaml
```diff
@@ -0,0 +1,15 @@
+databaseChangeLog:
+  - changeSet:
+      id: tag
+      author: snikitin
+      changes:
+        - tagDatabase:
+            tag: 025-add-flow-mirror-path-meter
```
dmitrii-beliakov, line 7: a copy-paste issue? I guess it must be 026-...

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/service/SpeakerFlowSegmentRequestBuilder.java
```diff
@@ -683,6 +690,52 @@ private Optional<MirrorConfig> makeMirrorConfig(@NonNull FlowPath flowPath, @Non
...
+            }
+        }
+        return result;
+    }
+    
+    private static Optional<MirrorConfigData> buildMultiSwitchMirrorConfig(
+            FlowMirrorPath mirrorPath, FlowEncapsulationType flowEncapsulationType,
+            Map<PathId, EncapsulationResources> mirrorEncapsulationMap) {
+        if (mirrorPath.getSegments() == null || mirrorPath.getSegments().isEmpty()) {
+            log.error(format("Multi switch flow mirror path %s has no segments",
```
dmitrii-beliakov, line 718: A minor thing, but still: "multi" is not a word and cannot be alone; it is either multiswitch as a noun, meaning multiswitch is the name of a type of a device, or it is multi-switch _something_, meaning multi-switch here is a compound adjective. I believe it is the second case here: it is a characteristic of a flow, right?

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/delete/actions/DeallocateResourcesAction.java
```diff
@@ -48,6 +48,7 @@ public void perform(State from, State to, Event event, FlowDeleteContext context
             resourcesManager.deallocateMirrorGroup(mirrorPathResources.getFlowPathId(),
                     mirrorPathResources.getMirrorSwitchId());
         });
+        // TODO deallocate mirrors
```
dmitrii-beliakov, line 51: Maybe to add a link to the issue? In my experience, when there is no issue, people tend to forget about TODOs

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/mirrorpoint/create/actions/RevertFlowMirrorPathAllocationAction.java
```diff
@@ -66,21 +67,24 @@ protected void perform(State from, State to, Event event, FlowMirrorPointCreateC
...
+            FlowMirrorPath forwardMirrorPath = getFlowMirrorPath(stateMachine.getForwardMirrorPathId());
+            stateMachine.getRevertCommands().addAll(ruleManager.buildMirrorPointRules(forwardMirrorPath, dataAdapter));
         } else {
             log.warn("Can't find mirror points for flow path {} and mirror switch {}. May cause excess rules.",
                     stateMachine.getFlowPathId(), stateMachine.getMirrorSwitchId());
         }
 
-        resourcesManager.deallocateCookie(stateMachine.getUnmaskedCookie());
+        resourcesManager.deallocatePathResources(stateMachine.getFlowResources());
         flowMirrorPathRepository.remove(stateMachine.getForwardMirrorPathId());
```
dmitrii-beliakov, line 81: All these removals must be a single atomic operation? What would happen if it fails on the repository removal? It is already de-allocated, but the state machine still holds the reference?

src-java/kilda-model/src/main/java/org/openkilda/model/FlowMirror.java
```diff
@@ -113,6 +116,25 @@ public boolean isOneSwitchMirror() {
...
+                .map(FlowMirrorPath::getCookie)
+                .map(FlowSegmentCookie::getFlowEffectiveId);
+        if (requestedPathCookie.isPresent()) {
+            return getMirrorPaths().stream()
+                    .filter(path -> !path.getMirrorPathId().equals(pathId))
+                    .filter(path -> path.getCookie().getFlowEffectiveId() == requestedPathCookie.get())
+                    .findAny()
+                    .map(FlowMirrorPath::getMirrorPathId);
+        } else {
+            throw new IllegalArgumentException(format("Flow mirror %s does not have mirror path %s",
```
dmitrii-beliakov, line 133: I am confused here. The else branch says there is no a single mirror path. If it is not the case, then there is at least one mirror path, then how an empty Optional could happen?
The exception message probably should be about path cookies and not about mirror path?

docker/db-migration/migrations/026-add-flow-mirror-reverse-path.yaml
niksv: yep

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/service/SpeakerFlowSegmentRequestBuilder.java
niksv: fixed

src-java/kilda-model/src/main/java/org/openkilda/model/FlowMirror.java
niksv: not really. 
```
        Optional<Long> requestedPathCookie = getPath(pathId)
                .map(FlowMirrorPath::getCookie)
                .map(FlowSegmentCookie::getFlowEffectiveId);
```
`requestedPathCookie` will be empty only if `getPath(pathId)` will return empty optional. That means there is no path with pathId `pathId`

dmitrii-beliakov: Right, but if there is no path with this `pathId ` as you described, then it is the situation described in the exception, because it is true that the `Flow mirror %s does not have the mirror path %s`, but the exception is not thrown.

niksv:
The task is on hold

PR #5006, opened 2022-12-01
Stabilize YFlow swap paths steps

pkazlenka:
* Dramatically decreased default interval to collect stats and stats waiting interval in tests


confd/templates/kafka/server.properties.tmpl
```diff
@@ -1,3 +1,6 @@
+# Generated by confd.
+# Do not change this file, all changes will be lost. Change corresponding template.
```
dmitrii-beliakov, line 2: What is the corresponding template that needs to be changed?

pkazlenka: This is kind of misleading stuff in our templates. I've filled the improvement request: https://github.com/telstra/open-kilda/issues/5033 . Long story short: we generate configuration from these templates, and, probably, the easiest way to put this warning into generated configuration files was to put the one into templates as well. All the rest of templates have the same problem.

dmitrii-beliakov: I see, thanks for the explanation and the creating the issue 👍

dmitrii-beliakov (approved):

artHalMirantis (approved):

durgakako (approved):

niksv (approved):

PR #5020, opened 2022-12-16
resuse bandwidth during Y create

### LINKED ISSUE #4965: Can't create Y-flow with bandwidth=available bandwidth
**Steps to reproduce:**
1. Find endpoints for Y flow which must have shared path. For emample
```
A ----B ---- C
      | 
      D
```
Where A is shared endpoint of Y flow, C and D - two other endpoints. A-B shared path
2. Find available bandwidth of shared path (A-B). Let in be X.
3. Create Y flow on these endpoints with bandwidth X via API `POST /v2/y-flows`

**Expected result:**
Flow was created

**Actual result:**
Flow was not created because "not enough bandwidth"

NOTE: I found variable `sharedBandwidthGroupId` in code, I guess Kilda use for shared path part to do not subtract bandwidth from ISL twice. Maybe it can help fix the bug

pablomuri:
Closes #4965

### DIFF
--- a/src-java/kilda-pce/src/main/java/org/openkilda/pce/AvailableNetworkFactory.java
+++ b/src-java/kilda-pce/src/main/java/org/openkilda/pce/AvailableNetworkFactory.java
@@ -32,7 +32,9 @@ import lombok.extern.slf4j.Slf4j;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 /**
@@ -63,20 +65,13 @@ public class AvailableNetworkFactory {
         AvailableNetwork network = new AvailableNetwork();
         try {
             // Reads all active links from the database and creates representation of the network.
-            getAvailableIsls(buildStrategy, flow)
-                    .forEach(link -> addIslAsEdge(link, network));
-
-            if (!reusePathsResources.isEmpty() && !flow.isIgnoreBandwidth()) {
-                reusePathsResources.stream()
-                        .filter(pathId -> flowPathRepository.findById(pathId)
-                                .map(path -> !path.isIgnoreBandwidth())
-                                .orElse(false))
-                        .forEach(pathId -> {
-                            // ISLs occupied by the flow (take the bandwidth already occupied by the flow into account).
-                            islRepository.findActiveByPathAndBandwidthAndEncapsulationType(
-                                    pathId, flow.getBandwidth(), flow.getEncapsulationType())
-                                    .forEach(link -> addIslAsEdge(link, network));
-                        });
+            getAvailableIsls(buildStrategy, flow).forEach(link -> addIslAsEdge(link, network));
+            if (!flow.isIgnoreBandwidth()) {
+                Set<PathId> reusePaths = new HashSet<>(reusePathsResources);
+                if (flow.getYFlowId() != null) {
+                    reusePaths.addAll(flowPathRepository.findPathIdsBySharedBandwidthGroupId(flow.getYFlowId()));
+                }
+                reuseResources(reusePaths, flow, network);
             }
         } catch (PersistenceException e) {
             throw new RecoverableException("An error from the database", e);
@@ -119,6 +114,19 @@ public class AvailableNetworkFactory {
         }
 
         return network;
+    }
+
+    private void reuseResources(Collection<PathId> reusePathsResources, Flow flow, AvailableNetwork network) {
+        reusePathsResources.stream()
+                .filter(pathId -> flowPathRepository.findById(pathId)
+                        .map(path -> !path.isIgnoreBandwidth())
+                        .orElse(false))
+                .forEach(pathId -> {
+                    // ISLs occupied by the flow (take the bandwidth already occupied by the flow into account).
+                    islRepository.findActiveByPathAndBandwidthAndEncapsulationType(
+                                    pathId, flow.getBandwidth(), flow.getEncapsulationType())
+                            .forEach(link -> addIslAsEdge(link, network));
+                });
     }
 
     private Collection<IslImmutableView> getAvailableIsls(BuildStrategy buildStrategy, Flow flow) {
--- a/src-java/kilda-pce/src/test/java/org/openkilda/pce/AvailableNetworkFactoryTest.java
+++ b/src-java/kilda-pce/src/test/java/org/openkilda/pce/AvailableNetworkFactoryTest.java
@@ -60,6 +60,9 @@ public class AvailableNetworkFactoryTest {
     public static final int DEST_PORT = 31;
     public static final PathId FORWARD_PATH_ID = new PathId("path_id_1");
     public static final PathId REVERSE_PATH_ID = new PathId("path_id_2");
+    public static final PathId PATH_ID_1 = new PathId("flow-path-id1");
+    public static final PathId PATH_ID_2 = new PathId("flow-path-id2");
+    public static final PathId PATH_ID_3 = new PathId("flow-path-id3");
     public static final Switch switchA = Switch.builder().switchId(SWITCH_ID_1).build();
     public static final Switch switchB = Switch.builder().switchId(SWITCH_ID_2).build();
     public static final Switch switchC = Switch.builder().switchId(SWITCH_ID_3).build();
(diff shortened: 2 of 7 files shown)


src-java/kilda-pce/src/main/java/org/openkilda/pce/AvailableNetworkFactory.java
```diff
@@ -65,18 +65,16 @@ public AvailableNetwork getAvailableNetwork(Flow flow, Collection<PathId> reuseP
...
-                                    pathId, flow.getBandwidth(), flow.getEncapsulationType())
-                                    .forEach(link -> addIslAsEdge(link, network));
-                        });
+            if (!flow.isIgnoreBandwidth()) {
+                if (flow.getYFlowId() != null) {
+                    Collection<PathId> flowPaths =
+                            flowPathRepository.findPathIdsBySharedBandwidthGroupId(flow.getYFlowId());
+                    if (!flowPaths.isEmpty()) {
+                        reuseResources(flowPaths, flow, network);
+                    }
```
niksv, line 74: I know that it works, but this code looks like not very correct. If we are working with Y flow we ignore `reusePathsResources` parameter. And this looks incorrect from user point of view. 
If user use `reusePathsResources` collection then he expects that this paths will be ignored.

I think we have several options here:
1. collected shared paths of Y flow in FlowHs topology and call getAvailableNetwork with list of Y flow path list
2. add some other method getAvailableNetwork only for Y flows
3. rewrite getAvailableNetwork to use repositories for finding reusePathsResources, but I think it's not possible, because sometimes we need to force the method to do not use shared resources. For example we use it in reroute operation: 
[reusePathsResources](https://github.com/telstra/open-kilda/blob/37ce148c2e974d3840c3e9a92cb238850ff07a0d/src-java/flowhs-topology/flowhs-storm-
topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/reroute/FlowRerouteFsm.java#L113)

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaFlowPathRepository.java
```diff
@@ -151,6 +151,17 @@ public Optional<FlowPath> findByCookie(FlowSegmentCookie cookie) {
...
+    public Collection<PathId> findPathIdsBySharedBandwidthGroupId(String sharedBandwidthGroupId) {
+        return framedGraph().traverse(g -> g.V()
+                        .hasLabel(FlowPathFrame.FRAME_LABEL)
+                        .has(FlowPathFrame.SHARED_BANDWIDTH_GROUP_ID_PROPERTY, sharedBandwidthGroupId)
+                        .values(FlowPathFrame.PATH_ID_PROPERTY))
+                .getRawTraversal().toStream()
+                .map(pathId -> PathIdConverter.INSTANCE.toEntityAttribute((String) pathId))
+                .collect(Collectors.toList());
+    }
+
```
niksv, line 164: Could you please add unit test for this method?

src-java/kilda-pce/src/main/java/org/openkilda/pce/AvailableNetworkFactory.java
pablomuri: Done. 
As discussed offline, I've joined both lists if it is YFlow instead of not using the reusePathResources.

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaFlowPathRepository.java
pablomuri: Done

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaFlowPathRepositoryTest.java
```diff
@@ -394,6 +394,23 @@ public void shouldFindFlowPathIdsByFlowIds() {
         assertTrue(pathIds.contains(flowB.getReversePathId()));
     }
 
+    @Test
+    public void findPathIdsBySharedBandwidthGroupId() {
+        String sharedBandwidthGroupId = "";
```
niksv, line 399: Just interesting: why groupId is empty string?

src-java/kilda-pce/src/test/java/org/openkilda/pce/AvailableNetworkFactoryTest.java
```diff
@@ -220,6 +220,167 @@ public void shouldBuildAvailableNetworkForFlowWithIgnoreBandwidthPaths() throws
         assertAvailableNetworkIsCorrect(isl, availableNetwork);
     }
 
+    @Test
+    public void shouldBuildAvailableNetworkReusingBandwidthFromYFlowAndReusePathResources() throws Exception {
+        String yFlowId = "y-flow-id";
+        Flow flow1 = Flow.builder()
+                .flowId("test-id")
+                .srcSwitch(Switch.builder().switchId(new SwitchId("1")).build())
```
niksv, line 228: we have a lot of constants in this class so you can reuse them if you want. For example 
`SWITCH_ID_1`

src-java/kilda-pce/src/test/java/org/openkilda/pce/AvailableNetworkFactoryTest.java
```diff
@@ -272,4 +433,14 @@ private static void assertAvailableNetworkIsCorrect(IslImmutableView isl, Availa
         assertNotNull(dst);
         assertEquals(1, dst.getIncomingLinks().size());
     }
+
+    private static void assertAvailableNetworkIsNotCorrect(IslImmutableView isl, AvailableNetwork availableNetwork) {
```
niksv, line 437: Wrong name of the method. Actually network is correct. Here you are checking that these is no ISL in available network

src-java/kilda-pce/src/test/java/org/openkilda/pce/AvailableNetworkFactoryTest.java
```diff
@@ -272,4 +433,14 @@ private static void assertAvailableNetworkIsCorrect(IslImmutableView isl, Availa
...
+        Node src = availableNetwork.getSwitch(isl.getSrcSwitchId());
+        assertNotNull(src);
+        assertEquals(0, src.getOutgoingLinks().size());
+
+        Node dst = availableNetwork.getSwitch(isl.getDestSwitchId());
+        assertNotNull(dst);
+        assertEquals(0, dst.getIncomingLinks().size());
```
niksv, line 444: It's not quite correct. this assert is valid only for your test case. To make it more correct you need to go throw all 
```
src.getOutgoingLinks()
src.getIncomingLinks()
dst.getOutgoingLinks()
dst.getIncomingLinks()
```

And check that all of them has no ISL `isl`

src-java/kilda-pce/src/main/java/org/openkilda/pce/AvailableNetworkFactory.java
pablomuri: Yes, much better. Even we can remove the if (!reusePaths.isEmpty()) since the reuseResources iterate over the Set, so if it is empty it doesn't iterate.

src-java/kilda-pce/src/test/java/org/openkilda/pce/AvailableNetworkFactoryTest.java
pablomuri: done

src-java/kilda-pce/src/test/java/org/openkilda/pce/AvailableNetworkFactoryTest.java
pablomuri: Done, checked it inside the test case

src-java/kilda-pce/src/test/java/org/openkilda/pce/AvailableNetworkFactoryTest.java
pablomuri: Done, checked it inside the test case

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaFlowPathRepositoryTest.java
pablomuri: just because my memory sometimes fails 😆 
fixed

niksv (approved):

pkazlenka:
Created an autotest PR, that verifies this issue: https://github.com/telstra/open-kilda/pull/5034

dmitrii-beliakov (approved):

durgakako (approved):

Issue #5015, opened 2022-12-13
Add API `GET /v2/switches/{switch-id}/flows`


niksv:
This API will be improved copy of API `/v1/switches/{switch-id}/flows`

v2 API must have 2 params
`switch_id` - mandatory
`port` - optional
(like v1 API)

but response must be different. it must be a dict with ports as keys. something like

```
{
    "flows_by_ports": 
    {
         "23": [ {<flow_1 info>}, {flow_2 info}],
         "35": [ {<flow_3 info>}, {flow_4 info}]
    {
}
```

`flow_info` can be copied from v1 API

We need new API to use it in Kilda GUI switch page. At this moment to get flows for each switch port Kilda GUI page send separate request for each port. As page send request every 30 seconds it generates 1000 request during 15 minutes per each switch page

PR #5031, opened 2023-01-02
v2 API: add get switch flows per ports

dmitrii-beliakov:


src-java/nbworker-topology/nbworker-messaging/src/main/java/org/openkilda/messaging/nbtopology/request/GetFlowsPerPortForSwitchRequest.java
```diff
@@ -0,0 +1,34 @@
+/* Copyright 2020 Telstra Open Source
```
dmitrii-beliakov, line 1: I need to correct the year

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepository.java
```diff
@@ -593,6 +599,61 @@ public Collection<Flow> findLoopedByLoopSwitchId(SwitchId switchId) {
...
+                    .as(pathSegmentAlias)
+                    .in("owns")
+                    .in("owns")
+                    .hasLabel(FlowFrame.FRAME_LABEL)
+                    .as(flowsAlias)
+                    .select(pathSegmentAlias, flowsAlias)
+            ).getRawTraversal().fill(new ArrayList<>()).stream()
+                    .map(x -> (Map<String, Vertex>) x)
+                            .collect(Collectors.toMap(m -> Integer.valueOf(m.get(pathSegmentAlias)
+                                    .property(PathSegmentFrame.SRC_PORT_PROPERTY)
```
dmitrii-beliakov, line 627: slipped out of sight when committing, there must be DST port as well.

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepository.java
```diff
@@ -622,4 +683,41 @@ protected void doRemove(FlowFrame frame) {
...
         return Flow.FlowCloner.INSTANCE.deepCopy(frame, entity);
     }
+
+    /**
+     * Find all flows with filtering by flow IDs. This method does it as a single operation for better performance.
+     * @param flowIds Flow IDs for filtering.
+     * @return an empty collection if flowIds is empty,
+     *     otherwise returns a collection of Flows with the provided IDs if they exist.
+     */
+    private Collection<Flow> findFlowsByFlowIds(Collection<String> flowIds) {
```
dmitrii-beliakov, line 693: This seems to me as a good candidate for another public method in the parent interface. But the next method is very specific to a particular task and probably not reusable elsewhere. I was wondering if I need to create a separate class or it's ok like it is now. What are opinions?

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepositoryTest.java
```diff
@@ -155,6 +199,7 @@ public void shouldUpdateFlow() {
         flow.setBandwidth(10000);
         flow.setEncapsulationType(FlowEncapsulationType.TRANSIT_VLAN);
 
+        assertTrue(flowRepository.findById(TEST_FLOW_ID).isPresent());
```
dmitrii-beliakov, line 202: I added this and the similar asserts just by the way to reduce inspection warnings about using get without checking if there is something in the Optional.

durgakako (approved):

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepository.java
```diff
@@ -622,4 +683,41 @@ protected void doRemove(FlowFrame frame) {
...
+        if (flowIds == null || flowIds.size() == 0) {
```
Fedichk, line 694: We could replace (flowIds == null || flowIds.size() == 0) with isEmpty(flowIds) method from org.apache.commons.collections.CollectionUtils or another library.

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepository.java
pablomuri: I agree to include it in the interface if you think is useful.

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepositoryTest.java
```diff
@@ -91,8 +94,49 @@ public void setUp() {
...
         switchA = createTestSwitch(TEST_SWITCH_A_ID.getId());
         switchB = createTestSwitch(TEST_SWITCH_B_ID.getId());
+        switchC = createTestSwitch(TEST_SWITCH_C_ID.getId());
 
-        assertEquals(2, switchRepository.findAll().size());
+        assertEquals(3, switchRepository.findAll().size());
+    }
+
+    @Test
+    public void findPortToFlowsMap() {
```
pablomuri, line 103: I'd change the name to something like "shouldFindSwitchFlowsByPortWithoutFilter" to keep the naming convection in this class

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepositoryTest.java
```diff
@@ -91,8 +94,49 @@ public void setUp() {
...
+        assertNotNull("The map must contain a non-null value for the test port", result.get(PORT_1));
+        assertEquals("The map must contain exactly two keys", 2, result.size());
+        assertEquals("The map must contain exactly 3 flows", 3, result.get(PORT_1).size());
+        assertTrue("The map must contain exactly 3 flows",
+                result.get(PORT_1).stream().map(Flow::getFlowId).collect(Collectors.toList())
+                .containsAll(Arrays.asList(TEST_FLOW_ID, TEST_FLOW_ID_2, TEST_FLOW_ID_3)));
+    }
+
+    @Test
+    public void findPortToFlowsMapWithFilter() {
```
pablomuri, line 122: same. I'd change the name to something like "shouldFindSwitchFlowsByPortWithFilter" to keep the naming convection in this class

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepositoryTest.java
```diff
@@ -91,8 +94,49 @@ public void setUp() {
...
+        createTestFlow(TEST_FLOW_ID_3, switchA, switchB);
+        createTestFlow(TEST_FLOW_ID_4, switchC, switchB);
+
+        Map<Integer, List<Flow>> result = flowRepository.findSwitchFlowsByPort(switchA.getSwitchId(), null);
+
+        assertNotNull(result);
+        assertTrue("The map must contain a key for the port", result.containsKey(PORT_1));
+        assertNotNull("The map must contain a non-null value for the test port", result.get(PORT_1));
+        assertEquals("The map must contain exactly two keys", 2, result.size());
+        assertEquals("The map must contain exactly 3 flows", 3, result.get(PORT_1).size());
```
pablomuri, line 115: "The map must contain exactly 3 flows" this comment is misleading. Here, you are asserting that the value 
 of PORT_1 contains 3 flows, not the map itself"

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepositoryTest.java
```diff
@@ -91,8 +94,49 @@ public void setUp() {
...
+        assertTrue("The map must contain exactly 3 flows",
```
pablomuri, line 116: Same here, but I think the line 115 is redundant

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepositoryTest.java
```diff
@@ -91,8 +94,49 @@ public void setUp() {
...
+
+        assertNotNull(result);
+        assertTrue("The map must contain a key for the port", result.containsKey(PORT_1));
+        assertFalse("The map must not contain a key for the filtered out port", result.containsKey(PORT_2));
+        assertNotNull("The map must contain a non-null value for the test port", result.get(PORT_1));
+        assertEquals("The map must contain exactly one key", 1, result.size());
+        assertEquals("The map must contain exactly 3 test flows", 3, result.get(PORT_1).size());
+        assertTrue("The map must contain exactly 3 flows for switchA",
+                result.get(PORT_1).stream().map(Flow::getFlowId).collect(Collectors.toList())
+                        .containsAll(Arrays.asList(TEST_FLOW_ID, TEST_FLOW_ID_2, TEST_FLOW_ID_3)));
```
pablomuri, line 139: same here

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/SwitchOperationsBolt.java
```diff
@@ -335,6 +339,17 @@ private SwitchConnectionsResponse getSwitchConnections(SwitchConnectionsRequest
         return switchOperationsService.getSwitchConnections(request.getSwitchId());
     }
 
+    private List<GetFlowsPerPortForSwitchResponse> getSwitchFlows(GetFlowsPerPortForSwitchRequest request) {
+        log.debug("getSwitchFlows, API call to getSwitchFlows");
```
pablomuri, line 343: I am not sure if this log could be misleading, the API call would be in [SwitchServiceImpl.java] isn't it?

pablomuri (changes requested):

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepositoryTest.java
dmitrii-beliakov: they together checks that there are exactly these 3 flows. If I remove the 115, there might 15 flows, which are not supposed to be there. It is there to make sure the query doesn't grab more than needed

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/SwitchOperationsBolt.java
dmitrii-beliakov: I think I'll remove it and other logs

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepositoryTest.java
dmitrii-beliakov: Actually, I wanted to do the opposite: remove _should_ from all test cases 😃 
Should in English means that it is up to a person to do it or not. Like as in those user stories types: must and should. It's ok to deliver a feature without "should" user stories, just nice to have, but not mandatory.
To me, _should find_ sounds like "it's ok if nothing is found" or some sort of hesitation: _maybe you'll find, I don't know, fingers crossed._

I'd vote for either not use _should_ at all (in the end it doesn't bring any new information), or add _must_ instead (this method must do what it supposed to do, otherwise it is a defect that needs to be fix). 
Another option might be given-when-then pattern, or a simplified version when-then. But we have strict rules about line length, so good descriptive names are no-go due to checkstyle 😒

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepository.java
```diff
@@ -593,6 +605,53 @@ public Collection<Flow> findLoopedByLoopSwitchId(SwitchId switchId) {
...
+                                    .map(flowIdToFlowMap::get)
+                                    .collect(Collectors.toList()),
+                            (l1, l2) ->
+                                    Stream.of(l1, l2).flatMap(Collection::stream)
+                                            .sorted()
+                                            .distinct()
+                                            .collect(Collectors.toList()),
+                                    TreeMap::new));
+        } catch (ClassCastException | NumberFormatException e) {
+            log.debug("An exception in findSwitchFlowsByPort: {}", e.getMessage());
```
timofei-durakov, line 650: Why is it debug?

timofei-durakov (changes requested):

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepository.java
dmitrii-beliakov: My reasoning is that an exception here is something not normal/recoverable; this is not going to be happening on daily basis. The scenario of usage is when something went wrong, but the error appears like "time out" error, then we want to debug to find out more details, or in opposite way: to make sure this exception is not thrown.
Examples why the code ends up here are: ferma frramework returns an object of different type or a port is not a number and cannot be parsed. I believe both are unlikely to happen.
Why not an error type? It could be, however, the exception is re-thrown, so the error is reported in a standard way. It might be tricky to find though, so I added the key word in the message to make it easier.

timofei-durakov: I,d bump the log level to error/exception tbh. The problem here is the following: this code is very core and it's not guaranteed every place it will be called will have proper logging/error handling. The exception should be rare, which makes things even more complicated. If it fire one in prod, it might be very tricky to reproduce, and not having that logs will be very painful

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/controller/v2/SwitchControllerV2.java
```diff
@@ -260,4 +261,23 @@ public CompletableFuture<SwitchValidationResultV2> validateSwitch(
...
+     * When a port ID is provided, returns flows only for the given port.
+     * Ports that don't have any associated flow are skipped, i.e. if no flows are going through this switch,
+     * returns an empty output.
+     * @param switchId a specific switch
+     * @param portIds optional. Filters the output to display this port only
+     * @return mapping of port->[flows]
+     */
+    @ApiOperation(value = "Get all flows for each port for the given switch",
+            response = SwitchFlowsPerPortResponse.class)
+    @GetMapping("/{switch_id}/flows-by-port")
```
pkazlenka, line 276: Why do you use an endpoint different to the one from #5015? It's a standard practice in this project to have V1/V2 endpoints with the same name but working slightly different. Please, get back to the initial naming.

pkazlenka (changes requested):

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/controller/v2/SwitchControllerV2.java
dmitrii-beliakov: We discussed it on demo that the pattern of plurals implies that the return type is a collection. Here we have a collection of collections, which is not consistent with other API names, such as connections, properties, and so on. This API in fact does something different that wasn't available in v1. I posted it [in slack](https://miracloud.slack.com/archives/G79HZSL0J/p1672911054415589) for visibility, there were no suggestions, so I changed it @pzakatov suggested

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/SwitchOperationsBolt.java
dmitrii-beliakov: done

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepositoryTest.java
dmitrii-beliakov: done

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepositoryTest.java
dmitrii-beliakov: done

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepositoryTest.java
dmitrii-beliakov: as we discussed on one of the past syncs, we might want to revisit the naming style. I suggest to leave it for now as-is. Once we arrive to a conclusion, I'll update the names

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepository.java
```diff
@@ -593,6 +604,56 @@ public Collection<Flow> findLoopedByLoopSwitchId(SwitchId switchId) {
...
+                                    .map(flowIdToFlowMap::get)
+                                    .collect(Collectors.toList()),
+                            (l1, l2) ->
+                                    Stream.of(l1, l2).flatMap(Collection::stream)
+                                            .sorted()
+                                            .distinct()
+                                            .collect(Collectors.toList()),
+                                    TreeMap::new));
+        } catch (ClassCastException | NumberFormatException e) {
+            log.debug(getClass().getSimpleName() + ": an exception in findSwitchFlowsByPort: " + e.getMessage(), e);
```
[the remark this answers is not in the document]
dmitrii-beliakov: done

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepository.java
```diff
@@ -593,6 +604,56 @@ public Collection<Flow> findLoopedByLoopSwitchId(SwitchId switchId) {
                 .collect(Collectors.toList());
     }
 
+    @Override
+    public Map<Integer, List<Flow>> findSwitchFlowsByPort(SwitchId switchId, Collection<Integer> ports) {
+        log.debug("{}: Starting findSwitchFlowsByPort for the switch {} for ports {}",
+                this.getClass().getSimpleName(), switchId.toString(), ports);
```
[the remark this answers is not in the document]
dmitrii-beliakov: I removed it. fixed

pablomuri:
aproved

pablomuri (approved):

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepository.java
```diff
@@ -593,6 +604,36 @@ public Collection<Flow> findLoopedByLoopSwitchId(SwitchId switchId) {
...
+        try {
+            @SuppressWarnings("unchecked")
+            Map<Integer, Collection<Flow>> portToFlowMap = collectToPortToFlowTreeMap(
+                    framedGraph().traverse(g -> g.V()
+                            .hasLabel(PathSegmentFrame.FRAME_LABEL)
+                            .or(
+                                    has(PathSegmentFrame.SRC_SWITCH_ID_PROPERTY, switchId.toString()),
+                                    has(PathSegmentFrame.DST_SWITCH_ID_PROPERTY, switchId.toString()))
+                            .as(PATH_SEGMENT_ALIAS)
+                            .in(FlowFrame.OWNS_PATHS_EDGE)
```
niksv, line 618: Should it be `FlowPathFrame.OWNS_SEGMENTS_EDGE`?

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/service/impl/SwitchServiceImpl.java
```diff
@@ -518,6 +522,25 @@ public CompletableFuture<List<FlowPayload>> getFlowsForSwitch(SwitchId switchId,
...
+        CommandMessage message = new CommandMessage(data, System.currentTimeMillis(), correlationId);
+
+        return messagingChannel.sendAndGet(nbworkerTopic, message)
+                            .thenApply(GetFlowsPerPortForSwitchResponse.class::cast)
+                            .thenApply(GetFlowsPerPortForSwitchResponse::getFlowsByPorts)
+                            .thenApply(portToFlowDto -> portToFlowDto.entrySet().stream()
+                                    .collect(Collectors.toMap(entry -> entry.getKey(), entry -> entry.getValue()
+                                            .stream()
+                                            .map(flowMapper::toFlowResponseV2).collect(Collectors.toList()))))
+                            .thenApply(SwitchFlowsPerPortResponse::new);
```
niksv, line 540: I think you can move mapping from `GetFlowsPerPortForSwitchResponse` to `SwitchFlowsPerPortResponse` into `switchMapping` module

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepository.java
dmitrii-beliakov: In fact, it is just a step in into owns edges, they are filtered later, on 620. The variable name is confusing a bit, maybe just make it `OWNS`, as it is actually: `String OWNS_PATHS_EDGE = "owns";`?

niksv: yes, it works because we have 2 edges with name "owns". But, for example, what if we will decide to rename OWNS_SEGMENTS_EDGE from 'own' to something else? In both cases (leaving FlowFrame.OWNS_PATHS_EDGE or use "owns") we will lose connection between this code and FlowPathFrame.OWNS_SEGMENTS_EDGE

dmitrii-beliakov: yeah, I see, good point. I'll add another variable

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/service/impl/SwitchServiceImpl.java
dmitrii-beliakov: I moved the mapping into the `SwitchMapper`

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepository.java
niksv: But the old value is still here

niksv (changes requested):

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/service/impl/SwitchServiceImpl.java
niksv: You didn't move it

niksv (changes requested):

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepository.java
```diff
@@ -593,6 +604,36 @@ public Collection<Flow> findLoopedByLoopSwitchId(SwitchId switchId) {
...
+            @SuppressWarnings("unchecked")
+            Map<Integer, Collection<Flow>> portToFlowMap = collectToPortToFlowTreeMap(
+                    framedGraph().traverse(g -> g.V()
+                            .hasLabel(PathSegmentFrame.FRAME_LABEL)
+                            .or(
+                                    has(PathSegmentFrame.SRC_SWITCH_ID_PROPERTY, switchId.toString()),
+                                    has(PathSegmentFrame.DST_SWITCH_ID_PROPERTY, switchId.toString()))
+                            .as(PATH_SEGMENT_ALIAS)
+                            .in(FlowFrame.OWNS_SEGMENTS_EDGE)
+                            .in(FlowFrame.OWNS_SEGMENTS_EDGE)
```
niksv, line 619: Could you pleaseu use "owns" from FlowFrame?

niksv (approved):

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaFlowRepository.java
dmitrii-beliakov: fixed. I also removed the constant from FlowFrame

PR #3859, opened 2020-11-18
Zero downtime

timofei-durakov:

PR #5035, opened 2023-01-10
#3859: Add flow mirror migration script

Fedichk:


docker/db-migration/migrations/025-move-data-from-flow-mirror-path-to-flow-mirror.yaml
```diff
@@ -0,0 +1,14 @@
...
+      changes:
+        - tagDatabase:
+            tag: 025-move-data-from-flow-mirror-path-to-flow-mirror
+
+  - changeSet:
+      id: move_data_from_flow_mirror_path_to_flow_mirror
+      author: fkantur
+      changes:
+        - customChange:
+            class: "org.openkilda.FlowMirrorMigration"
```
nikitamarchenko, line 14: nice

Fedichk: Thanks!

src-java/orientdb-migration/src/main/java/org/openkilda/FlowMirrorMigration.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @SuppressWarnings({"UnusedDeclaration", "FieldCanBeLocal"})
+    private ResourceAccessor resourceAccessor;
+
+    @Override
+    public void execute(Database database) throws CustomChangeException {
+        String script = "BEGIN;\n";
+        script += "FOREACH ($path IN (SELECT FROM flow_mirror_path)) {\n";
+        script += "LET $fm = INSERT INTO flow_mirror (flow_mirror_id, forward_path_id, mirror_switch_id, "
+                + "egress_switch_id, status, egress_port, egress_outer_vlan, egress_inner_vlan) VALUES ($path.path_id, "
```
dmitrii-beliakov, line 45: This is why we need more modern java with multiline strings. 🙂
A minor thing: you can use `StringBuilder::append` instead of `+=` and `+`

src-java/orientdb-migration/src/main/java/org/openkilda/FlowMirrorMigration.java
```diff
@@ -0,0 +1,91 @@
...
+                + "$path.path_id, $path.mirror_switch_id, $path.egress_switch_id, $path.status, $path.egress_port, "
+                + "$path.egress_outer_vlan, $path.egress_inner_vlan);\n";
+        script += "UPDATE EDGE owns SET in = $fm.@rid[0] UPSERT WHERE @rid = $path.in_owns[0];\n";
```
dmitrii-beliakov, line 48: what is the meaning of 0 here?

Fedichk: $fm.@rid will return an array of 1 element, so, it is the index of the element in an array.

src-java/orientdb-migration/src/main/java/org/openkilda/FlowMirrorMigration.java
Fedichk: Agree, StringBuilder has better performance, but from my point of view, the current implementation is a little bit more readable and focused on the SQL script itself. If you don't mind, I would prefer to keep the current implementation.

src-java/orientdb-migration/src/main/java/org/openkilda/FlowMirrorMigration.java
dmitrii-beliakov: maybe it worth to explicitly specify `return @rid` in the insert clause.

src-java/orientdb-migration/src/main/java/org/openkilda/FlowMirrorMigration.java
```diff
@@ -0,0 +1,91 @@
...
+        script += "CREATE EDGE owns FROM $fm TO $path;\n";
+        script += "};\n";
+        script += "COMMIT;\n";
```
dmitrii-beliakov, line 51: This migrate doesn't look like idempotent. If the data in DB is large it might fail in the middle and the only way to proceed is the rollback and try again. And the next attempt might fail again.

Maybe it is overengineering, but I'd think about making the script in a form of updating only not already updated edges and commit in batches. Then if it fails, only the last batch is not committed and you can restart the migrate to continue from where it stopped.

Fedichk: Sure, please provide an example of code, which implements your ideas and I will add it.

Fedichk: Also, just FYI this script runs within one pessimistic transaction.

dmitrii-beliakov: Maybe the first thing is to understand the boundaries. If the data set you are updating is very large and the operation is performed on a live DB, there could be all sorts of issues such as dead locks, slowness, connection time outs.
To address this, we can use batch operation. Most likely Orient DB supports it, but I am not an expert in this area, right now I don't have a receipt. Otherwise we can do it using plain SQL:
overall idea is that you select some subset, do operation, commit, select next subset, and so on. Semantically:
while there is data
begin
perform updates in a loop
commit
end
To select subset of input data you can use the select filtering what is already processed with limiting to some reasonable amount of data: SELECT FROM flow_mirror_path WHERE <a predicate for not processed paths*> LIMIT 1000
then you loop over that 1000 records, commit, select a new portion of 1000 records, update, commit, and so on Once there is no more data you exit the outer loop. If that scripts fails for whatever reason, you can start it again and only the last 1000 is not committed. So you'll resume the updating on the last portion.
* A predicate could be NOT EXIST (what you want to add) or a value in some column/property. In your case, I believe it is the records which don't have an edge you are adding.

Fedichk: Let's wait for Sergey. I think he knows boundaries. Then we will be able to decide if we need some changes.
I don't expect a big amount of data.

niksv:
The task is on hold

PR #5045, opened 2023-01-23
fix max latency reason not shown

pablomuri:
In the case the path is not fully calculated, we can't know the latency of the full path because there is a point during the path computation where the latency is already higher than the max_latency property.

For this case the reason was not properly added so it wasn't inlcuded in the response.

This commit fix it.

### DIFF
--- a/src-java/kilda-pce/src/test/java/org/openkilda/pce/impl/MaxLatencyPathComputationStrategyBaseTest.java
+++ b/src-java/kilda-pce/src/test/java/org/openkilda/pce/impl/MaxLatencyPathComputationStrategyBaseTest.java
@@ -16,10 +16,10 @@
 package org.openkilda.pce.impl;
 
 import static java.lang.String.format;
+import static org.hamcrest.CoreMatchers.endsWith;
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.not;
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.containsString;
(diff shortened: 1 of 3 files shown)


src-java/kilda-pce/src/test/java/org/openkilda/pce/impl/MaxLatencyPathComputationStrategyBaseTest.java
```diff
@@ -351,7 +350,7 @@ public void affinityOvercomeDiversity() throws Exception {
     }
 
     @Test
-    public void shouldNotFindPathGreaterThanMaxLatency() {
+    public void notFindPathGreaterThanMaxLatency() {
```
pkazlenka, line 353: Actually, there's an easy way to create test names: it should contain three parts: condition, action and result. E.g. <result>exception<action>InPathComputation<condition>WhenPathLatencyGreaterThanMaxLatency.
Anyway, your shorter option of the test name doesn't make the one clearer, IMHO.

pkazlenka (approved):
Mostly good, but, please, consider restoring/changing the test name from my comment.

src-java/kilda-pce/src/test/java/org/openkilda/pce/impl/MaxLatencyPathComputationStrategyBaseTest.java
pablomuri: done

pkazlenka:
Verified manually + unit test. No need to add functional test.

niksv (approved):

PR #5047, opened 2023-01-24
#5042: Move bw fail reason to the end of the log message tests

pkazlenka:
* Updated test "Unable to update to a flow with unavailable bandwidth"


pablomuri (approved):
LGTM

PR #5049, opened 2023-01-26
Update description

Fedichk:


README.md
```diff
@@ -66,11 +76,13 @@ Do that by adding the user to /etc/groups and logging out and back in again.
 
 ##### Basic installation instruction from Docker site
 
-```sudo apt-get install apt-transport-https ca-certificates curl gnupg lsb-release
-curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg
-echo "deb [arch=amd64 signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+```
+sudo apt-get install ca-certificates curl gnupg lsb-release
```
pkazlenka, line 80: What has happened to `apt-transport-https`?
Looking at your update, I have a suspicion that you are trying to insert part of instruction for ubuntu 22.04 into instruction for 20.04. 
Probably, we should either add two versions of instructions (for 20.04 and 22.04) or emphasize somewhere, that this was written/verified for ubuntu 22.04 only.
If I'm wrong, please proceed to merge.

pkazlenka (approved):

README.md
Fedichk: Regarding this page https://docs.docker.com/engine/install/ubuntu/ there is no difference between Ubuntu versions. I just added what works for me and for @artHalMirantis, as well. Maybe we needed `apt-transport-https` before, I don't know, now it is definitely redundant.

dmitrii-beliakov (approved):

Issue #5017, opened 2022-12-14
Missed `kilda.flow.packets` metric value for sub flows of Y flows


niksv:
Kilda do not collect egress stats for Y flow subflows

**Steps to reproduce on local env:**
1. Create Y flow
```
{
  "y_flow_id": "flow_1",
  "shared_endpoint": {
    "port_number": 10,
    "switch_id": "9"
  },
  "maximum_bandwidth": 10000,
  "ignore_bandwidth": true,
  "sub_flows": [
    {
      "flow_id": "sub_1",
      "endpoint": {
        "inner_vlan_id": 0,
        "port_number": 10,
        "switch_id": "2",
        "vlan_id": 123
      },
      "shared_endpoint": {
        "inner_vlan_id": 0,
        "vlan_id": 456
      }
    },
{
      "flow_id": "sub_2",
      "endpoint": {
        "inner_vlan_id": 0,
        "port_number": 10,
        "switch_id": "7",
        "vlan_id": 345
      },
      "shared_endpoint": {
        "inner_vlan_id": 0,
        "vlan_id": 789
      }
    }
  ]
}
```
2. send some traffic to shared endpoint (subflow: sub_2)
```
curl --location --request POST 'localhost:8288/api/1/traffgen/tg9/address' --header 'Content-Type: application/json' --data-raw '{  
    "address": "192.168.1.5/30",
    "vlan": 789                             
}'
```

```
curl --location --request PUT 'localhost:8288/api/1/traffgen/tg9/address/{traff_gen_id_from_prev_respose}/lldp' --header 'Content-Type: application/json' --data-raw '{
        "mac_address": "e4:1d:2d:a5:f3:ac", 
        "port_number": "e4:1d:2d:a5:f3:ac", 
        "chassis_id": "00:02:03:04:05:00", 
        "time_to_live": 100
}'
```

3. Check `kilda.flow.packets` metric in opentsdb for subflow `sub_2` in forward direction (not rate, but absolute value)

**Expected result:**
growing line

**Actual result:**
No such metric for forward direction
But there is correct metric for:
ingress: `kilda.flow.ingress.packets`
raw egress: `kilda.flow.raw.packets` 

![Screenshot from 2022-12-14 16-15-30](https://user-images.githubusercontent.com/8777565/207593063-34b1bc96-0118-484e-af65-dc5be3dcf4a2.png)

PR #5056, opened 2023-01-31
fix yflow is not created without max_bandwidth bug

### LINKED ISSUE #5039: YFlow is not created without `max_bandwidth`
**Steps to reproduce:**
make test-topology

Try to create a YFlow without max_bandwidth included in the body:
```
HTTP POST http://kilda-vm:8080/api/v2/y-flows
{
  "y_flow_id": "flow1",
  "shared_endpoint": {
    "port_number": 10,
    "switch_id": "1"
  },
  "sub_flows": [
    {
      "flow_id": "sub1",
      "endpoint": {
        "inner_vlan_id": 0,
        "port_number": 10,
        "switch_id": "2",
        "vlan_id": 123
      },
      "shared_endpoint": {
        "inner_vlan_id": 0,
        "vlan_id": 456
      }
    },
    {
      "flow_id": "sub2",
      "endpoint": {
        "inner_vlan_id": 0,
        "port_number": 10,
        "switch_id": "3",
        "vlan_id": 345
      },
      "shared_endpoint": {
        "inner_vlan_id": 0,
        "vlan_id": 789
      }
    }
  ]
}
```
Returns OK.

Get the flow `flow1`:
`HTTP GET http://kilda-vm:8080/v2/y-flows/flow1`

**Expected result:**
`Flow1`info.

**Actual result:**
"Y-flow not found"

NOTE: It has been tested in `develop` and `v1.126.3`.

pablomuri:
Closes #5039

### DIFF
--- a/src-java/rule-manager/rule-manager-implementation/src/main/java/org/openkilda/rulemanager/RuleManagerImpl.java
+++ b/src-java/rule-manager/rule-manager-implementation/src/main/java/org/openkilda/rulemanager/RuleManagerImpl.java
@@ -599,7 +599,6 @@ public class RuleManagerImpl implements RuleManager {
                             flow.getFlowId(), protectedYPointSwitchId);
                     continue;
                 }
-                requireNonNull(protectedYPointMeterId, "The y-flow protected path meterId can't be null");
 
                 boolean meterToBeAdded = externalProtectedMeterCommandUuid == null;
                 if (meterToBeAdded) {
@@ -620,7 +619,6 @@ public class RuleManagerImpl implements RuleManager {
                             flow.getFlowId(), yPointSwitchId);
                     continue;
                 }
-                requireNonNull(yPointMeterId, "The y-flow meterId can't be null");
 
                 boolean meterToBeAdded = externalMeterCommandUuid == null;
                 if (meterToBeAdded) {
@@ -646,6 +644,7 @@ public class RuleManagerImpl implements RuleManager {
                 segment.getSrcSwitchId().equals(switchId) || segment.getDestSwitchId().equals(switchId));
     }
 
+    @VisibleForTesting
     private RuleGenerator buildYPointYRuleGenerator(Switch yPointSwitch, FlowPath path,
                                                     Flow flow, FlowTransitEncapsulation encapsulation,
                                                     MeterId yPointMeterId, UUID externalMeterCommandUuid,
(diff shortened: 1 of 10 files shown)


src-java/rule-manager/rule-manager-implementation/src/main/java/org/openkilda/rulemanager/RuleManagerImpl.java
```diff
@@ -511,6 +511,11 @@ private List<SpeakerData> buildSharedEndpointYFlowCommands(List<FlowPath> flowPa
         }
         MeterId sharedMeterId = yFlow.getSharedEndpointMeterId();
 
+        if (sharedMeterId == null) {
+            log.debug("Skip commands for the y-flow {} as the sharedMeterId is null", yFlow.getYFlowId());
+            return emptyList();
```
dmitrii-beliakov, line 516: Are there tests covering these new cases? Require non-null below might throw an exception and now that iteration is skipped, so it's a different behavior

dmitrii-beliakov (approved):

src-java/rule-manager/rule-manager-implementation/src/main/java/org/openkilda/rulemanager/RuleManagerImpl.java
niksv: Line 549 contains code with generator. This generator generates OpenFlow commands. In out case it's meter and rule commands. If we do not have meterId then we should not generate meter commands, but we still must generate rules commands.

You code will successfully create Y flow withput meter, but this flow wouldn't transmit traffic. To check traffic transmission you can one of the following:
1. Create functional tests with traffic generations.
2. Generate traffic manually and look at rules counters and statistics in OpenTSDB. Example: https://github.com/telstra/open-kilda/issues/5017

niksv (changes requested):

src-java/rule-manager/rule-manager-implementation/src/main/java/org/openkilda/rulemanager/RuleManagerImpl.java
pablomuri: I was reading the code and thought it was only installing meter commands. In fact, I tried with the ping tool from the API and it works. I will try the traffic transmissions that you propose. Thanks

niksv: Ping tool has one disadvantage: during ping it doesn't use ingress (first) and egress (last) rules in flow path. It tests only transit rules because of technical restrictions. As you removed ingress rules - ping was successful

dmitrii-beliakov (approved):

niksv (approved):

artHalMirantis (approved):

PR #5070, opened 2023-02-06
Path validation

### LINKED ISSUE #4527: [Proposal] Path check
**Path check [Proposal]**
----
This API endpoint is needed to verify that a specified path complies with the chosen strategy and network availability. 

* **URL**

  /v2/network/path/check

* **Method:**
  
   `GET`

* **Header:**

  * `correlation_id` - Request’s unique identifier

* **Data Params**
   
   **Required:**
    * `nodes` - List of switches with ports to check
   
   **Optional:**
    * `encapsulation_type`: `FlowEncapsulationType`
    * `max_latency`: Maximum latency of flow path in milliseconds
    * `max_latency_tier2`: Second tier for flow path latency in milliseconds
    * `path_computation_strategy`: `PathComputationStrategy`
    * `diverse_with`: diverse with a specific flow (under discussion)
    * `bandwidth`: required bandwidth (under discussion)

   **Request Body:**
    ```json
    {
      "encapsulation_type": "TRANSIT_VLAN|VXLAN",
      "max_latency": int,
      "max_latency_tier2": int,
      "path_computation_strategy": "COST|LATENCY|MAX_LATENCY|COST_AND_AVAILABLE_BANDWIDTH",
      "nodes": [
        {
          "switch_id": "00:00:00:00:00:00:00:01",
          "output_port": int
        },
        {
          "switch_id": "00:00:00:00:00:00:00:02",
          "input_port": int,
          "output_port": int
        },
        {
          "switch_id": "00:00:00:00:00:00:00:03",
          "input_port": int
        }
      ]
    }
    ```

* **Success Response:**
  
  * **Code:** 200 <br />
    **Content:** 
      * `valid` - True in case path is valid and conform to strategy
      * `errors` - List of strings with errors in case `valid` is False
      * `correlation_id` - Correlation Id from correlation_id Header
      
      ```json
      {
        "correlation_id": "string",
        "valid": "bool",
        "errors": [
          "error0",
          "errorN"
        ]
      }
      ```
 
* **Error Response:**

 * **Code:** 4XX,5XX <br />
    ```json
    {
      "correlation_id": "string",
      "error-description": "string",
      "error-message": "string",
      "error-type": "string",
      "timestamp": 0
    }
    ```

* **Notes:**

    The API call is similar to /v1/network/paths, so I suggest using DTO from it.

dmitrii-beliakov:
This feature introduces a new API that allows to perform various validations of a given path. Which validations are executed depends on a specific input parameters combination. Validations are executed for every segment of a path in both directions separately. If there are multiple problems on a segment, all errors are returned. There are the following validations and errors:
- a link is not found.
- a switch or switch properties are not found.
- no sufficient latency.
- no sufficient latency tier 2.
- no sufficient bandwidth.
- a segment has intersections with some flow. In other words, this path cannot be protected or this path gets penalty as it is not fully diverse.
- encapsulation type mismatch.

closes #4527

### DIFF
--- a/docs/design/solutions/path-validation/path-validation.md
+++ b/docs/design/solutions/path-validation/path-validation.md
@@ -1,0 +1,97 @@
+# Path Validation
+
+## Motivation
+The main use case is to allow users to verify whether some arbitrary flow path is possible to create with the given 
+constraints without actually creating this flow. A path here is a sequence of nodes, that is a sequence of switches
+with in and out ports. Constraints are various parameters such as max latency, bandwidth, encapsulation type, 
+path computation strategy, and other.
+
+## Implementation details
+The validation of a path is done for each segment and each validation type individually. This way, the validation
+collects all errors on the path and returns them all in a single response. The response is concise and formed 
+in human-readable format.
+
+There is no locking of resources for this path and, therefore, no guarantee that it will be possible to create this flow 
+after validation in the future.
+
+## Northbound API
+
+REST URL: ```/v2/network/path/check```, method: ```GET```
+
+A user is required to provide a path represented by a list of nodes. Nodes must be ordered from start to end; 
+each next element is the next hop. A user can add optional parameters:
+- `encapsulation_type`: enum value "TRANSIT_VLAN" or "VXLAN". API returns an error for every switch in the list if it 
+  doesn't support the given encapsulation type.
+- `max_bandwidth`: bandwidth required for this path. API returns an error for each segment of the given path when the 
+  available bandwidth on the segment is less than the given value. When used in combination with `reuse_flow_resources`,
+  available bandwidth as if the given flow doesn't consume any bandwidth.
+- `max_latency`: the first tier latency value in milliseconds. API returns an error for each segment of the given path,  
+  which max latency is greater than the given value.
+- `max_latency_tier2`: the second tier latency value in milliseconds. API returns an error for each segment of the given
+  path, which max latency is greater than the given value.
+- `path_computation_strategy`: an enum value PathComputationStrategy. API will return different set of errors depending 
+  on the selected strategy. For example, when COST is selected API ignores available bandwidth and latency parameters. 
+  If none is selected, all validations are executed.  
+- `reuse_flow_resources`: a flow ID. Verify the given path as if it is created instead of the existing flow, that is as 
+  if the resources of some flow are released before validation. Returns an error if this flow doesn't exist.
+- `diverse_with_flow`: a flow ID. Verify whether the given path intersects with the given flow. API returns an error for 
+  each common segment. Returns an error if this flow doesn't exist.
+
+### Request Body
+```json
+{
+  "encapsulation_type": "TRANSIT_VLAN|VXLAN",
+  "max_bandwidth": 0,
+  "max_latency": 0,
+  "max_latency_tier2": 0,
+  "path_computation_strategy": "COST|LATENCY|MAX_LATENCY|COST_AND_AVAILABLE_BANDWIDTH",
+  "reuse_flow_resources": "flow_id",
+  "diverse_with_flow": "diverse_flow_id",
+  "nodes": [
+    {
+      "switch_id": "00:00:00:00:00:00:00:01",
+      "output_port": 0
+    },
+    {
+      "switch_id": "00:00:00:00:00:00:00:02",
+      "input_port": 0,
+      "output_port": 1
+    },
+    {
+      "switch_id": "00:00:00:00:00:00:00:03",
+      "input_port": 0
+    }
+  ]
+}
+```
+
+### Responses
+
+Code: 200
+
+Content:
+- is_valid: True if the path is valid and conform to the strategy, false otherwise
+- errors: List of strings with error descriptions if any
+- correlation_id: Correlation ID from correlation_id Header
+```json
+{
+  "correlation_id": "string",
+  "is_valid": "bool",
+  "errors": [
+    "error0",
+    "errorN"
+  ]
+}
+```
+
+Codes: 4XX,5XX
+```json
+{
+  "correlation_id": "string",
+  "error-description": "string",
+  "error-message": "string",
+  "error-type": "string",
+  "timestamp": 0
+}
+```
+
--- a/src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/info/network/PathValidationResult.java
+++ b/src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/info/network/PathValidationResult.java
@@ -1,0 +1,35 @@
+/* Copyright 2023 Telstra Open Source
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+package org.openkilda.messaging.info.network;
+
+import org.openkilda.messaging.info.InfoData;
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import lombok.Builder;
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+
+import java.util.List;
+
+@Value
+@Builder
+@EqualsAndHashCode(callSuper = false)
+@JsonNaming(value = PropertyNamingStrategy.SnakeCaseStrategy.class)
+public class PathValidationResult extends InfoData {
+    Boolean isValid;
+    List<String> errors;
+}
--- a/src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/payload/network/PathValidationPayload.java
+++ b/src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/payload/network/PathValidationPayload.java
@@ -1,0 +1,59 @@
+/* Copyright 2023 Telstra Open Source
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+package org.openkilda.messaging.payload.network;
+
+import org.openkilda.messaging.payload.flow.FlowEncapsulationType;
+import org.openkilda.messaging.payload.flow.PathNodePayload;
+import org.openkilda.model.PathComputationStrategy;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Builder;
+import lombok.Value;
+
+import java.util.List;
+import javax.validation.constraints.PositiveOrZero;
+
+@Value
+@Builder
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class PathValidationPayload {
+    @JsonProperty("bandwidth")
+    Long bandwidth;
+
+    @JsonProperty("max_latency")
+    @PositiveOrZero(message = "max_latency cannot be negative")
+    Long latencyMs;
+
+    @JsonProperty("max_latency_tier2")
+    @PositiveOrZero(message = "max_latency_tier2 cannot be negative")
+    Long latencyTier2ms;
+
+    @JsonProperty("nodes")
+    List<PathNodePayload> nodes;
+
+    @JsonProperty("diverse_with_flow")
+    String diverseWithFlow;
+
+    @JsonProperty("reuse_flow_resources")
+    String reuseFlowResources;
+
+    @JsonProperty("flow_encapsulation_type")
+    FlowEncapsulationType flowEncapsulationType;
+
+    @JsonProperty("path_computation_strategy")
+    PathComputationStrategy pathComputationStrategy;
+}
--- a/src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/FlowEncapsulationTypeMapper.java
+++ b/src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/FlowEncapsulationTypeMapper.java
@@ -1,0 +1,31 @@
+/* Copyright 2023 Telstra Open Source
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+package org.openkilda.wfm.share.mappers;
+
+import org.openkilda.model.FlowEncapsulationType;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.factory.Mappers;
+
+@Mapper
+public interface FlowEncapsulationTypeMapper {
+
+    FlowEncapsulationTypeMapper INSTANCE = Mappers.getMapper(FlowEncapsulationTypeMapper.class);
+
+    FlowEncapsulationType toOpenKildaModel(
+            org.openkilda.messaging.payload.flow.FlowEncapsulationType flowEncapsulationType);
+
+}
--- a/src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/PathValidationDataMapper.java
+++ b/src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/PathValidationDataMapper.java
@@ -1,0 +1,67 @@
+/* Copyright 2023 Telstra Open Source
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+package org.openkilda.wfm.share.mappers;
+
+import org.openkilda.messaging.payload.network.PathValidationPayload;
+import org.openkilda.model.PathValidationData;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.factory.Mappers;
+
+import java.time.Duration;
+import java.util.LinkedList;
+import java.util.List;
+
+@Mapper
+public abstract class PathValidationDataMapper {
+
+    public static PathValidationDataMapper INSTANCE = Mappers.getMapper(PathValidationDataMapper.class);
+
+    /**
+     * Converts NB PathValidationDto to messaging PathValidationData.
+     * @param pathValidationPayload NB representation of a path validation data
+     * @return the messaging representation of a path validation data
+     */
+    public PathValidationData toPathValidationData(PathValidationPayload pathValidationPayload) {
+        List<PathValidationData.PathSegmentValidationData> segments = new LinkedList<>();
+
+        for (int i = 0; i < pathValidationPayload.getNodes().size() - 1; i++) {
+            segments.add(PathValidationData.PathSegmentValidationData.builder()
+                    .srcSwitchId(pathValidationPayload.getNodes().get(i).getSwitchId())
+                    .srcPort(pathValidationPayload.getNodes().get(i).getOutputPort())
+                    .destSwitchId(pathValidationPayload.getNodes().get(i + 1).getSwitchId())
+                    .destPort(pathValidationPayload.getNodes().get(i + 1).getInputPort())
+                    .build());
+        }
+
+        return PathValidationData.builder()
+                .srcSwitchId(pathValidationPayload.getNodes().get(0).getSwitchId())
+                .destSwitchId(
+                        pathValidationPayload.getNodes().get(pathValidationPayload.getNodes().size() - 1).getSwitchId())
+                .bandwidth(pathValidationPayload.getBandwidth())
+                .latency(pathValidationPayload.getLatencyMs() == null ? null :
+                        Duration.ofMillis(pathValidationPayload.getLatencyMs()))
+                .latencyTier2(pathValidationPayload.getLatencyTier2ms() == null ? null :
+                        Duration.ofMillis(pathValidationPayload.getLatencyTier2ms()))
+                .diverseWithFlow(pathValidationPayload.getDiverseWithFlow())
+                .reuseFlowResources(pathValidationPayload.getReuseFlowResources())
+                .flowEncapsulationType(FlowEncapsulationTypeMapper.INSTANCE.toOpenKildaModel(
+                        pathValidationPayload.getFlowEncapsulationType()))
+                .pathComputationStrategy(pathValidationPayload.getPathComputationStrategy())
+                .pathSegments(segments)
+                .build();
+    }
+}
--- a/src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/PathsBolt.java
+++ b/src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/PathsBolt.java
@@ -15,6 +15,7 @@
 
 package org.openkilda.wfm.topology.nbworker.bolts;
 
+import org.openkilda.messaging.command.flow.PathValidateRequest;
 import org.openkilda.messaging.error.ErrorType;
 import org.openkilda.messaging.error.MessageException;
 import org.openkilda.messaging.info.InfoData;
@@ -55,6 +56,8 @@ public class PathsBolt extends PersistenceOperationsBolt {
         List<? extends InfoData> result = null;
         if (request instanceof GetPathsRequest) {
             result = getPaths((GetPathsRequest) request);
+        } else if (request instanceof PathValidateRequest) {
+            result = pathService.validatePath((PathValidateRequest) request);
         } else {
             unhandledInput(tuple);
         }
--- a/src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/RouterBolt.java
+++ b/src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/RouterBolt.java
@@ -20,6 +20,7 @@ import static org.openkilda.wfm.topology.utils.KafkaRecordTranslator.FIELD_ID_PA
 import org.openkilda.messaging.Message;
 import org.openkilda.messaging.command.CommandData;
 import org.openkilda.messaging.command.CommandMessage;
+import org.openkilda.messaging.command.flow.PathValidateRequest;
 import org.openkilda.messaging.error.ErrorMessage;
 import org.openkilda.messaging.info.InfoMessage;
 import org.openkilda.messaging.nbtopology.request.BaseRequest;
@@ -88,7 +89,7 @@ public class RouterBolt extends AbstractBolt {
             emitWithContext(StreamType.FEATURE_TOGGLES.toString(), input, new Values(request));
         } else if (request instanceof KildaConfigurationBaseRequest) {
             emitWithContext(StreamType.KILDA_CONFIG.toString(), input, new Values(request));
-        } else if (request instanceof GetPathsRequest) {
+        } else if (request instanceof GetPathsRequest || request instanceof PathValidateRequest) {
             emitWithContext(StreamType.PATHS.toString(), input, new Values(request));
         } else if (request instanceof HistoryRequest) {
             emitWithContext(StreamType.HISTORY.toString(), input, new Values(request));
--- a/src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
+++ b/src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
@@ -1,0 +1,600 @@
+/* Copyright 2023 Telstra Open Source
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+package org.openkilda.wfm.topology.nbworker.validators;
+
+import org.openkilda.messaging.info.network.PathValidationResult;
+import org.openkilda.model.Flow;
+import org.openkilda.model.FlowEncapsulationType;
+import org.openkilda.model.FlowPath;
+import org.openkilda.model.Isl;
+import org.openkilda.model.IslStatus;
+import org.openkilda.model.KildaConfiguration;
+import org.openkilda.model.PathComputationStrategy;
+import org.openkilda.model.PathSegment;
+import org.openkilda.model.PathValidationData;
+import org.openkilda.model.PathValidationData.PathSegmentValidationData;
+import org.openkilda.model.Switch;
+import org.openkilda.model.SwitchId;
+import org.openkilda.model.SwitchProperties;
+import org.openkilda.persistence.repositories.FlowRepository;
+import org.openkilda.persistence.repositories.IslRepository;
+import org.openkilda.persistence.repositories.IslRepository.IslEndpoints;
+import org.openkilda.persistence.repositories.SwitchPropertiesRepository;
+import org.openkilda.persistence.repositories.SwitchRepository;
+
+import com.google.common.collect.Sets;
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import org.apache.commons.collections4.map.LazyMap;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.storm.shade.com.google.common.collect.Maps;
+
+import java.time.Duration;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+public class PathValidator {
+
+    private static final String LATENCY = "latency";
+    private static final String LATENCY_TIER_2 = "latency tier 2";
+    private final IslRepository islRepository;
+    private final FlowRepository flowRepository;
+    private final SwitchPropertiesRepository switchPropertiesRepository;
+    private final SwitchRepository switchRepository;
+    private final KildaConfiguration kildaConfiguration;
+    private Map<IslEndpoints, Optional<Isl>> islCache;
+    private Map<SwitchId, Optional<Switch>> switchCache;
+
+    public PathValidator(IslRepository islRepository,
+                         FlowRepository flowRepository,
+                         SwitchPropertiesRepository switchPropertiesRepository,
+                         SwitchRepository switchRepository, KildaConfiguration kildaConfiguration) {
+        this.islRepository = islRepository;
+        this.flowRepository = flowRepository;
+        this.switchPropertiesRepository = switchPropertiesRepository;
+        this.switchRepository = switchRepository;
+        this.kildaConfiguration = kildaConfiguration;
+    }
+
+    /**
+     * Validates whether it is possible to create a path with the given parameters. When there are obstacles, this
+     * validator returns all errors found on each segment. For example, when there is a path 1-2-3-4 and there is
+     * no link between 1 and 2, no sufficient latency between 2 and 3, and not enough bandwidth between 3-4, then
+     * this validator returns all 3 errors.
+     *
+     * @param pathValidationData path parameters to validate.
+     * @return a response object containing the validation result and errors if any
+     */
+    public PathValidationResult validatePath(PathValidationData pathValidationData) {
+        Set<String> result = pathValidationData.getPathSegments().stream()
+                .map(segment -> executeValidations(
+                        InputData.of(pathValidationData, segment),
+                        getPerSegmentValidations(pathValidationData)))
+                .flatMap(Set::stream)
+                .collect(Collectors.toSet());
+
+        result.addAll(executeValidations(InputData.of(pathValidationData),
+                getPerPathValidations(pathValidationData)));
+
+        return PathValidationResult.builder()
+                .isValid(result.isEmpty())
+                .errors(new LinkedList<>(result))
+                .build();
+    }
+
+    private Optional<Isl> findIslByEndpoints(SwitchId srcSwitchId, int srcPort, SwitchId destSwitchId, int destPort) {
+        if (islCache == null) {
+            islCache = LazyMap.lazyMap(Maps.newHashMap(), endpoints -> islRepository.findByEndpoints(
+                    new SwitchId(endpoints.getSrcSwitch()), endpoints.getSrcPort(),
+                    new SwitchId(endpoints.getDestSwitch()), endpoints.getDestPort()));
+        }
+        return islCache.get(new IslEndpoints(srcSwitchId.toString(), srcPort, destSwitchId.toString(), destPort));
+    }
+
+    private Map<SwitchId, Switch> findSwitchesByIds(Set<SwitchId> switchIds) {
+        if (switchCache == null) {
+            switchCache = LazyMap.lazyMap(Maps.newHashMap(), switchRepository::findById);
+        }
+
+        return switchIds.stream()
+                .map(switchCache::get)
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .collect(Collectors.toMap(Switch::getSwitchId, Function.identity()));
+    }
+
+    private Set<String> executeValidations(InputData inputData,
+                                           List<Function<InputData, Set<String>>> validations) {
+
+        return validations.stream()
+                .map(f -> f.apply(inputData))
+                .flatMap(Set::stream)
+                .collect(Collectors.toSet());
+    }
+
+    private List<Function<InputData, Set<String>>> getPerPathValidations(
+            PathValidationData pathValidationData) {
+        List<Function<InputData, Set<String>>> validationFunctions = new LinkedList<>();
+
+        if (isLatencyValidationRequired(pathValidationData)) {
+            validationFunctions.add(this::validateLatency);
+        }
+
+        if (isLatencyTier2ValidationRequired(pathValidationData)) {
+            validationFunctions.add(this::validateLatencyTier2);
+        }
+
+        return validationFunctions;
+    }
+
+    private List<Function<InputData, Set<String>>> getPerSegmentValidations(
+            PathValidationData pathValidationData) {
+        List<Function<InputData, Set<String>>> validationFunctions = new LinkedList<>();
+
+        validationFunctions.add(this::validateForwardAndReverseLinks);
+
+        if (isBandwidthValidationRequired(pathValidationData)) {
+            validationFunctions.add(this::validateBandwidth);
+        }
+
+        if (isDiverseWithFlowValidationRequired(pathValidationData)) {
+            validationFunctions.add(this::validateDiverseWithFlow);
+        }
+
+        if (isEncapsulationTypeValidationRequired(pathValidationData)) {
+            validationFunctions.add(this::validateEncapsulationType);
+        }
+
+        return validationFunctions;
+    }
+
+    private boolean isEncapsulationTypeValidationRequired(PathValidationData pathValidationData) {
+        return getOrDefaultFlowEncapsulationType(pathValidationData) != null;
+    }
+
+    private boolean isDiverseWithFlowValidationRequired(PathValidationData pathValidationData) {
+        return StringUtils.isNotBlank(pathValidationData.getDiverseWithFlow());
+    }
+
+    private boolean isBandwidthValidationRequired(PathValidationData pathValidationData) {
+        return pathValidationData.getBandwidth() != null
+                && pathValidationData.getBandwidth() != 0;
+    }
+
+    private boolean isLatencyTier2ValidationRequired(PathValidationData pathValidationData) {
+        return pathValidationData.getLatencyTier2() != null
+                && !pathValidationData.getLatencyTier2().isZero()
+                && (getOrDefaultPathComputationStrategy(pathValidationData) == PathComputationStrategy.LATENCY
+                || getOrDefaultPathComputationStrategy(pathValidationData) == PathComputationStrategy.MAX_LATENCY);
+    }
+
+    private boolean isLatencyValidationRequired(PathValidationData pathValidationData) {
+        return pathValidationData.getLatency() != null
+                && !pathValidationData.getLatency().isZero()
+                && (getOrDefaultPathComputationStrategy(pathValidationData) == PathComputationStrategy.LATENCY
+                || getOrDefaultPathComputationStrategy(pathValidationData) == PathComputationStrategy.MAX_LATENCY);
+    }
+
+    private Set<String> validateForwardAndReverseLinks(InputData inputData) {
+        Map<SwitchId, Switch> switchMap = findSwitchesByIds(
+                Sets.newHashSet(inputData.getSegment().getSrcSwitchId(), inputData.getSegment().getDestSwitchId()));
+        Set<String> errors = Sets.newHashSet();
+        if (!switchMap.containsKey(inputData.getSegment().getSrcSwitchId())) {
+            errors.add(getSrcSwitchNotFoundError(inputData));
+        }
+        if (!switchMap.containsKey(inputData.getSegment().getDestSwitchId())) {
+            errors.add(getDestSwitchNotFoundError(inputData));
+        }
+        if (!errors.isEmpty()) {
+            return errors;
+        }
+
+        errors.addAll(validateForwardLink(inputData));
+        errors.addAll(validateReverseLink(inputData));
+
+        return errors;
+    }
+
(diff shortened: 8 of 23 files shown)


docs/design/solutions/path-validation/path-validation.md
```diff
@@ -0,0 +1,80 @@
...
+A user is required to provide a path represented by a list of nodes. Nodes must be ordered from start to end, 
+each next element is the next hop. A user can add optional parameters.
+
+### Request Body
+```json
+{
+  "encapsulation_type": "TRANSIT_VLAN|VXLAN",
+  "max_latency": 0,
+  "max_latency_tier2": 0,
+  "path_computation_strategy": "COST|LATENCY|MAX_LATENCY|COST_AND_AVAILABLE_BANDWIDTH",
```
niksv, line 32: I would suggest to add couple fields here:
1. `max_bandwidth`
2. `reuse_flow_resources: {flow_id}` (you can change this name)
Last one is needed for following case:
Topology:
```
A-------B-------C
        |
        D 
```
We have existing flow `flow1` with path ABC
and we want to check if this flow can be on path ABD, but ISL AB was already occupied by flow1. So it wouldn't enough bandwidth on it to check path ABD. So if user specified `reuse_flow_resources` the API must ignore resources of this flow. (like we do in case of flow Update. TO find new path we ignore current path of a flow)

docs/design/solutions/path-validation/path-validation.md
```diff
@@ -0,0 +1,80 @@
+# Path Validation
+
+## Motivation
+The main use case is to allow users to validate if some arbitrary flow path is possible to create with the given 
+constraints without actually creating this flow. A path here is a sequence of nodes, that is a sequence of switches
+with in and out ports. Constraints are various parameters such as max latency, bandwidth, encapsulation type, 
+path computation strategy, and other.
+
+## Implementation proposal
+Reuse PCE components (specifically extend AvailableNetwork) for the path validation in the same manner as it is used for
```
pablomuri, line 10: I think you're going to extend not only AvailableNetwork but also the finders. I mean, the only way to check that a path accomplish max_latency, or cost_and_available_bandwidth is to find the actual path, and compare it with the given one. And the most of times it will be only one available path.

BTW, take also in care that the AvailableNetwork (right now) is going to give you the network isls that already have enough bandwidth, so you are not going to be able to see the complete network.

pablomuri (commented):
Just a couple of observations but LGTM

docs/design/solutions/path-validation/path-validation.md
dmitrii-beliakov: I didn't know that about bandwidth, thanks! From one perspective, available bandwidth is one of the parameters, if there's not enough bandwidth, this path is invalid. But it then brings the question whether this is a mandatory parameter. Also, in case of not sufficient bandwidth, there will be no way to validate other parameters via AvailableNetwork.

niksv: max_bandwidth is not mandatory. At least user always can set it to 0

dmitrii-beliakov:
closes #4527

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,456 @@
...
+        if (getReverseBandwidthWithReusableResources(reverse.get(), inputData) < inputData.getPath().getBandwidth()) {
+            errors.add(getReverseBandwidthErrorMessage(inputData, reverse.get().getAvailableBandwidth()));
+        }
+
+        return errors;
+    }
+
+    private long getForwardBandwidthWithReusableResources(Isl isl, InputData inputData) {
+        return getBandwidthWithReusableResources(inputData, isl,
+                pathSegment -> inputData.getSegment().getSrcSwitchId().equals(pathSegment.getSrcSwitchId())
```
pkazlenka, line 225: Four lines lambda is too hard to read. Please, move to the method with self-explaining name, if possible.
(same for below methods)

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,456 @@
...
+                    .findAny());
+
+            return segment.map(s -> isl.getAvailableBandwidth() + s.getBandwidth())
+                    .orElseGet(isl::getAvailableBandwidth);
+        }
+
+        return isl.getAvailableBandwidth();
+    }
+
+    private Set<String> validateEncapsulationType(InputData inputData, RepositoryData repositoryData) {
```
pkazlenka, line 259: Shouldn't you also validate encapsulation type on the intermediate switches?

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/controller/v2/NetworkControllerV2.java
```diff
@@ -0,0 +1,67 @@
...
+     * @param pathValidationDto a payload with a path and additional flow parameters provided by a user
+     * @return either a successful response or the list of errors
+     */
+    @GetMapping(path = "/path/check")
+    @ApiOperation(value = "Validates that a given path complies with the chosen strategy and the network availability")
+    @ResponseStatus(HttpStatus.OK)
+    public CompletableFuture<PathValidateResponse> validateCustomFlowPath(
+            @RequestBody PathValidationDto pathValidationDto) {
+
+        if (pathValidationDto == null
```
pkazlenka, line 58: Shouldn't this stuff to be in some Validator class?

pkazlenka (changes requested):
Reviewed. Please, look for the solutions where it makes sense.

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/PathValidationDataMapper.java
```diff
@@ -0,0 +1,67 @@
...
+                    .destSwitchId(pathValidationDto.getNodes().get(i + 1).getSwitchId())
+                    .destPort(pathValidationDto.getNodes().get(i + 1).getInputPort())
+                    .build());
+        }
+
+        return PathValidationData.builder()
+                .srcSwitchId(pathValidationDto.getNodes().get(0).getSwitchId())
+                .destSwitchId(pathValidationDto.getNodes().get(pathValidationDto.getNodes().size() - 1).getSwitchId())
+                .bandwidth(pathValidationDto.getBandwidth())
+                .latency(pathValidationDto.getLatencyMs() == null ? null :
```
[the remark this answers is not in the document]
dmitrii-beliakov: deserialization needs to support Optional, in jackson it is `jackson-datatype-jdk8` as far as I know, and we currently don't have this dependency. I'll investigate what could be done.

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,456 @@
...
+     * this validator returns all 3 errors.
+     *
+     * @param pathValidationData path parameters to validate.
+     * @return a response object containing the validation result and errors if any
+     */
+    public PathValidationResult validatePath(PathValidationData pathValidationData) {
+        Set<String> result = pathValidationData.getPathSegments().stream()
+                .map(segment -> executeValidations(
+                        new InputData(pathValidationData, segment),
+                        new RepositoryData(islRepository, flowRepository, switchPropertiesRepository, switchRepository),
```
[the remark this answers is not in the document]
dmitrii-beliakov: Actually, this class is just a holder for several references. This sort of Java objects are cheap. I can move the contents of that class to a member variables, there will be no RepositoryData at all. But I created it like this, so that all validation methods could be parts of a stream and I chose BiFunction with encapsulated repos, because I didn't know which repos I will require here. Also, I can unwrap the 4 repos into a new functional interface.

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,456 @@
...
+                        getValidations(pathValidationData)))
+                .flatMap(Set::stream)
+                .collect(Collectors.toSet());
```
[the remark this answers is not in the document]
dmitrii-beliakov: There could be duplicates, but removing them is expected. The errors contain the segment information, so there won't be removal of error from different segments. There could be exactly the same errors because they are given by different validators. For example, if there is no link on a segment, then you'll find it when validating bandwidth on this segment as well as latency on this segment.

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,456 @@
...
+
+        return Collections.emptySet();
+    }
+
+    private Set<String> validateBandwidth(InputData inputData, RepositoryData repositoryData) {
+        Optional<Isl> forward = repositoryData.getIslRepository().findByEndpoints(
+                inputData.getSegment().getSrcSwitchId(), inputData.getSegment().getSrcPort(),
+                inputData.getSegment().getDestSwitchId(), inputData.getSegment().getDestPort());
+        if (!forward.isPresent()) {
+            return Collections.singleton(getNoForwardIslError(inputData));
```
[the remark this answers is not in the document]
dmitrii-beliakov: good catch! 👍 As far as I know, currently, it is not possible to have only a reverse path on some segment, but this code should not have this assumption.

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
dmitrii-beliakov: The validations are executed against each segment on a path. So it validates the encapsulation type on all switches on a path. Or am I missing something? what do you by intermediate switches?

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/controller/v2/NetworkControllerV2.java
dmitrii-beliakov: I think it is quite common to sanitize the user's input or put some defaults in a controller before passing to a service. Like they do a bean validation in hibernate right in the method signature:
```
@GetMapping(path = "/someUrl")
public ResponseEntity<List<Object>> method(@RequestParam(value = "list", required = false) List<@Size(min = 1, max = 2) String> list)
```
Is it worth it to create a separate class for these 3 conditions?

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/PathValidationDataMapper.java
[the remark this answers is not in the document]
pkazlenka: Ok. I don't think it's necessary to add new library for such a little stuff, but if it can simplify our code significantly - probably we can create an enhancement/refactoring PR.

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/controller/v2/NetworkControllerV2.java
pkazlenka: I believe we should stay unified in all our decisions. If we leave validation to the separate classes in other parts, we should create a new one here. 
If we allow simple validations in other controllers, we may leave the one here as well. Please, check the other code.

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,456 @@
...
+                data.getSegment().getSrcSwitchId(), data.getSegment().getSrcPort(),
+                data.getSegment().getDestSwitchId(), data.getSegment().getDestPort());
+    }
+
+    private String getReverseIslNotActiveError(InputData data) {
+        return String.format(
+                "The ISL is not in ACTIVE state between end points: switch %s port %d and switch %s port %d",
+                data.getSegment().getDestSwitchId(), data.getSegment().getDestPort(),
+                data.getSegment().getSrcSwitchId(), data.getSegment().getSrcPort());
+    }
```
pablomuri, line 409: These functions are not being tested

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,456 @@
...
+                inputData.getSegment().getSrcSwitchId(),
+                inputData.getSegment().getSrcPort(),
+                inputData.getSegment().getDestSwitchId(),
+                inputData.getSegment().getDestPort());
+
+        if (!forward.isPresent()) {
+            return Collections.singleton(getNoForwardIslError(inputData));
+        }
+        if (forward.get().getStatus() != IslStatus.ACTIVE) {
+            return Collections.singleton(getForwardIslNotActiveError(inputData));
```
pablomuri, line 167: this line is not being tested

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,456 @@
...
+
+        return Collections.emptySet();
+    }
+
+    private Set<String> validateReverseLink(InputData inputData, RepositoryData repositoryData) {
+        if (!repositoryData.getSwitchRepository().findById(inputData.getSegment().getDestSwitchId()).isPresent()) {
+            return Collections.singleton(getDestSwitchNotFoundError(inputData));
+        }
+        if (!repositoryData.getSwitchRepository().findById(inputData.getSegment().getSrcSwitchId()).isPresent()) {
+            return Collections.singleton(getSrcSwitchNotFoundError(inputData));
```
pablomuri, line 178: this line is not being tested

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,456 @@
...
+                inputData.getSegment().getDestSwitchId(),
+                inputData.getSegment().getDestPort(),
+                inputData.getSegment().getSrcSwitchId(),
+                inputData.getSegment().getSrcPort());
+
+        if (!reverse.isPresent()) {
+            return Collections.singleton(getNoReverseIslError(inputData));
+        }
+        if (reverse.get().getStatus() != IslStatus.ACTIVE) {
+            return Collections.singleton(getReverseIslNotActiveError(inputData));
```
pablomuri, line 191: this line is not being tested

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,456 @@
...
+        Set<String> errors = new HashSet<>();
+        if (getForwardBandwidthWithReusableResources(forward.get(), inputData) < inputData.getPath().getBandwidth()) {
+            errors.add(getForwardBandwidthErrorMessage(inputData, forward.get().getAvailableBandwidth()));
+        }
+
+        Optional<Isl> reverse = repositoryData.getIslRepository().findByEndpoints(
+                inputData.getSegment().getDestSwitchId(), inputData.getSegment().getDestPort(),
+                inputData.getSegment().getSrcSwitchId(), inputData.getSegment().getSrcPort());
+        if (!reverse.isPresent()) {
+            return Collections.singleton(getNoForwardIslError(inputData));
```
pablomuri, line 214: this line is not being tested

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,456 @@
...
+
+        return isl.getAvailableBandwidth();
+    }
+
+    private Set<String> validateEncapsulationType(InputData inputData, RepositoryData repositoryData) {
+        Set<String> errors = new HashSet<>();
+        Optional<SwitchProperties> srcSwitchProperties = switchPropertiesRepository.findBySwitchId(
+                inputData.getPath().getSrcSwitchId());
+        if (!srcSwitchProperties.isPresent()) {
+            errors.add(getSrcSwitchNotFoundError(inputData));
```
pablomuri, line 264: this line is not being tested

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,456 @@
...
+        });
+        return errors;
+    }
+
+    private Set<String> validateDiverseWithFlow(InputData inputData, RepositoryData repositoryData) {
+        if (!repositoryData.getIslRepository().findByEndpoints(inputData.getSegment().getSrcSwitchId(),
+                inputData.getSegment().getSrcPort(),
+                inputData.getSegment().getDestSwitchId(),
+                inputData.getSegment().getDestPort()).isPresent()) {
+            return Collections.singleton(getNoForwardIslError(inputData));
```
pablomuri, line 294: this line is not being tested

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,456 @@
...
+
+        return Collections.emptySet();
+    }
+
+    private Set<String> validateLatencyTier2(InputData inputData, RepositoryData repositoryData) {
+        Optional<Isl> isl = repositoryData.getIslRepository().findByEndpoints(
+                inputData.getSegment().getSrcSwitchId(), inputData.getSegment().getSrcPort(),
+                inputData.getSegment().getDestSwitchId(), inputData.getSegment().getDestPort());
+        if (!isl.isPresent()) {
+            return Collections.singleton(getNoForwardIslError(inputData));
```
pablomuri, line 336: this line is not being tested

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,456 @@
...
+        }
+
+        Duration actualLatency = Duration.ofNanos(isl.get().getLatency());
+        if (actualLatency.compareTo(inputData.getPath().getLatencyTier2()) > 0) {
+            return Collections.singleton(getLatencyTier2ErrorMessage(inputData, isl.get().getLatency()));
+        }
+
+        return Collections.emptySet();
```
pablomuri, line 344: this line is not being tested

pablomuri (changes requested):
Thanks for your effort, this is so valuable. Thanks also for fixing a lot of typos and so on.
Just a few minor comments. 
The lines not being tested have been checked with the IntelliJ coverage.

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/FlowEncapsulationTypeMapper.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.mapstruct.Mapper;
+import org.mapstruct.factory.Mappers;
+
+@Mapper
+public interface FlowEncapsulationTypeMapper {
+
+    FlowEncapsulationTypeMapper INSTANCE = Mappers.getMapper(FlowEncapsulationTypeMapper.class);
+
+    FlowEncapsulationType toOpenKildaModel(
```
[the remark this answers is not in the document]
dmitrii-beliakov: There are two enums with the same name in different modules/packages (or there are two families of classes in general). One in kilda-model, those model classes representing flows, paths, encapsulation types are used in NB API. Another one in base-messaging, those model class representing flows, paths, encapsulation types are used in Storm bolts. So on the path from user to storm and back everything is transformed two times.
This mapstruct mapper annotation on a class generates a converter from this interface. By default, it maps class' fields with the same name, or enums with same enum values. So this method takes, let's say, `FlowEncapsulationType.VXLAN` from messagins and returns `FlowEncapsulationType.VXLAN` from kilda-model.

(I hope I didn't messed up in the explanation, it's confusing 😃)

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/info/network/PathValidationResult.java
```diff
@@ -0,0 +1,46 @@
+/* Copyright 2018 Telstra Open Source
```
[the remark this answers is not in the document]
dmitrii-beliakov: fixed

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/info/network/PathValidationResult.java
```diff
@@ -0,0 +1,46 @@
...
+import lombok.Value;
+
+import java.util.List;
+
+@Value
+@Builder
+@EqualsAndHashCode(callSuper = false)
+@JsonNaming(value = PropertyNamingStrategy.SnakeCaseStrategy.class)
+public class PathValidationResult extends InfoData {
+    @JsonProperty("is_valid")
```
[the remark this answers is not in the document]
dmitrii-beliakov: I removed the redundant annotations

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/RouterBolt.java
```diff
@@ -90,6 +91,8 @@ private void processRequest(Tuple input, String key, BaseRequest request) {
             emitWithContext(StreamType.KILDA_CONFIG.toString(), input, new Values(request));
         } else if (request instanceof GetPathsRequest) {
             emitWithContext(StreamType.PATHS.toString(), input, new Values(request));
+        } else if (request instanceof PathValidateRequest) {
```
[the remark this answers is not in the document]
dmitrii-beliakov: done

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,456 @@
...
+        }
+
+        if (pathValidationData.getBandwidth() != null
+                && pathValidationData.getBandwidth() != 0
+                && (pathValidationData.getPathComputationStrategy() == null
+                || pathValidationData.getPathComputationStrategy() == COST_AND_AVAILABLE_BANDWIDTH)) {
+            validationFunctions.add(this::validateBandwidth);
+        }
+
+        if (pathValidationData.getDiverseWithFlow() != null && !pathValidationData.getDiverseWithFlow().isEmpty()) {
```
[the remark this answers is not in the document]
dmitrii-beliakov: I changed it to use StringUtils

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,456 @@
...
+        errors.addAll(validateReverseLink(inputData, repositoryData));
+
+        return errors;
+    }
+
+    private Set<String> validateForwardLink(InputData inputData, RepositoryData repositoryData) {
+        if (!repositoryData.getSwitchRepository().findById(inputData.getSegment().getSrcSwitchId()).isPresent()) {
+            return Collections.singleton(getSrcSwitchNotFoundError(inputData));
+        }
+        if (!repositoryData.getSwitchRepository().findById(inputData.getSegment().getDestSwitchId()).isPresent()) {
```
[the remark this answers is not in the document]
dmitrii-beliakov: I refactored it in multiple places

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,456 @@
...
+            return Collections.singleton(getNoForwardIslError(inputData));
+        }
+        if (forward.get().getStatus() != IslStatus.ACTIVE) {
+            return Collections.singleton(getForwardIslNotActiveError(inputData));
+        }
+
+        return Collections.emptySet();
+    }
+
+    private Set<String> validateReverseLink(InputData inputData, RepositoryData repositoryData) {
```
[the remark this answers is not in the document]
dmitrii-beliakov: Not really, at least not for free. There are two kind of segments: a segment as a Kilda model and a segment as a part of user's input, which is just a holder of several fields. The later is not a fully-fledge segment and doesn't have any reverse functionality. 
However, I refactored it to have less duplicated code.

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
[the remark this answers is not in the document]
dmitrii-beliakov: I changed the implementation in multiple places to execute validations independently for each direction

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,456 @@
...
+    private Set<String> validateBandwidth(InputData inputData, RepositoryData repositoryData) {
+        Optional<Isl> forward = repositoryData.getIslRepository().findByEndpoints(
+                inputData.getSegment().getSrcSwitchId(), inputData.getSegment().getSrcPort(),
+                inputData.getSegment().getDestSwitchId(), inputData.getSegment().getDestPort());
+        if (!forward.isPresent()) {
+            return Collections.singleton(getNoForwardIslError(inputData));
+        }
+
+        Set<String> errors = new HashSet<>();
+        if (getForwardBandwidthWithReusableResources(forward.get(), inputData) < inputData.getPath().getBandwidth()) {
```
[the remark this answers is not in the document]
dmitrii-beliakov: done

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
dmitrii-beliakov: fixed

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/controller/v2/NetworkControllerV2.java
dmitrii-beliakov: There are some examples of validations or some simple data transformation in other controllers. I'd leave it like that. It is really just an input sanity check: provide at least two points in a path is mandatory. Just a more verbose version of `required = true`

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/payload/network/PathValidationDto.java
```diff
@@ -0,0 +1,79 @@
+/* Copyright 2019 Telstra Open Source
```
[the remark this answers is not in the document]
dmitrii-beliakov: The class is renamed to `PathValidationPayload`. I fixed the year there. 
A copy-paste sin. 😔

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
dmitrii-beliakov: The code has changed and some methods are moved. But I added new test cases for the gaps in tests you mentioned.

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/FlowEncapsulationTypeMapper.java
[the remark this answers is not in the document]
pablomuri: Thank you for the explanation, now I understand the functionality. Maybe the name is not very descriptive, but I really can't think of a better idea. If you don't think of anything better I think the current name is enough.

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/info/network/PathValidationResult.java
```diff
@@ -0,0 +1,46 @@
...
+    Boolean isValid;
```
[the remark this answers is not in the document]
dmitrii-beliakov: In the end they are private. Lombok preprocessor sees the Value annotation and changes this class to be an immutable class with getters for all fields. It also makes all fields private. Yeah, not sure I like it as well, maybe it's better to be explicit and disable the corresponding inspection in idea (it will mark private as redundant).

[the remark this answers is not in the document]
Fedichk: I see. Thank you for the clarification!

pkazlenka (approved):

Fedichk (approved):

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/PathValidationDataMapper.java
```diff
@@ -0,0 +1,67 @@
...
+                        Duration.ofMillis(pathValidationDto.getLatencyTier2ms()))
+                .diverseWithFlow(pathValidationDto.getDiverseWithFlow())
+                .reuseFlowResources(pathValidationDto.getReuseFlowResources())
+                .flowEncapsulationType(FlowEncapsulationTypeMapper.INSTANCE.toOpenKildaModel(
+                        pathValidationDto.getFlowEncapsulationType()))
+                .pathComputationStrategy(pathValidationDto.getPathComputationStrategy())
+                .pathSegments(segments)
+                .build();
+    }
+
```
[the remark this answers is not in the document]
dmitrii-beliakov: I removed redundant spaces in multiple places.

docs/design/solutions/path-validation/path-validation.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Motivation
+The main use case is to allow users to verify whether some arbitrary flow path is possible to create with the given 
+constraints without actually creating this flow. A path here is a sequence of nodes, that is a sequence of switches
+with in and out ports. Constraints are various parameters such as max latency, bandwidth, encapsulation type, 
+path computation strategy, and other.
+
+## Implementation details
+The validation of a path is done for each segment and each validation type individually. This way, the validation
+collects all errors on the path and returns them all in a single response. The response is concise amd formed
```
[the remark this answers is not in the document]
dmitrii-beliakov: fixed

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,513 @@
...
+                && !pathValidationData.getLatencyTier2().isZero()
+                && (pathValidationData.getPathComputationStrategy() == null
+                || pathValidationData.getPathComputationStrategy() == PathComputationStrategy.LATENCY
+                || pathValidationData.getPathComputationStrategy() == MAX_LATENCY);
+    }
+
+    private boolean isLatencyValidationRequired(PathValidationData pathValidationData) {
+        return pathValidationData.getLatency() != null
+                && !pathValidationData.getLatency().isZero()
+                && (pathValidationData.getPathComputationStrategy() == null
```
niksv, line 166: same

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,513 @@
...
+                || pathValidationData.getPathComputationStrategy() == PathComputationStrategy.LATENCY
+                || pathValidationData.getPathComputationStrategy() == MAX_LATENCY);
+    }
+
+    private Set<String> validateForwardAndReverseLinks(InputData inputData, RepositoryData repositoryData) {
+        Map<SwitchId, Switch> switchMap = repositoryData.getSwitchRepository().findByIds(
```
niksv, line 172: why can't we use switch repository from `this`?

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,513 @@
...
+        if (isl.isPresent() && isl.get().getStatus() != IslStatus.ACTIVE) {
+            errors.add(linkNotActiveErrorProducer.apply(inputData));
+        }
+
+        return errors;
+    }
+
+    private Optional<Isl> getIslByEndPoints(SwitchId srcSwitchId, int srcPort, SwitchId destSwitchId, int destPort,
+                                            RepositoryData repositoryData) {
+        return repositoryData.getIslRepository().findByEndpoints(srcSwitchId, srcPort, destSwitchId, destPort);
```
niksv, line 229: why not to use repository from `this`?

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,513 @@
...
+    }
+
+    private Optional<Isl> getForwardIslOfSegmentData(PathSegmentValidationData data, RepositoryData repositoryData) {
+        return repositoryData.getIslRepository().findByEndpoints(
+                data.getSrcSwitchId(), data.getSrcPort(),
+                data.getDestSwitchId(), data.getDestPort());
+    }
+
+    private Optional<Isl> getReverseIslOfSegmentData(PathSegmentValidationData data, RepositoryData repositoryData) {
+        return repositoryData.getIslRepository().findByEndpoints(
```
niksv, line 239: same

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,513 @@
...
+
+        return errors;
+    }
+
+    private Set<String> validateDiverseWithFlow(InputData inputData, RepositoryData repositoryData) {
+        if (!repositoryData.getIslRepository().findByEndpoints(inputData.getSegment().getSrcSwitchId(),
+                inputData.getSegment().getSrcPort(),
+                inputData.getSegment().getDestSwitchId(),
+                inputData.getSegment().getDestPort()).isPresent()) {
+            return Collections.singleton(getNoForwardIslError(inputData));
```
niksv, line 344: Why are we checking ISL existence here? didn't we check it in validateForwardAndReverseLinks?

niksv (changes requested):

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/payload/network/PathValidationPayload.java
```diff
@@ -0,0 +1,80 @@
...
+    @JsonProperty("flow_encapsulation_type")
+    FlowEncapsulationType flowEncapsulationType;
+
+    @JsonProperty("path_computation_strategy")
+    PathComputationStrategy pathComputationStrategy;
+
+    @JsonCreator
+    public PathValidationPayload(
+            @JsonProperty("bandwidth") Long bandwidth,
+            @JsonProperty("latency_ms") Long latencyMs,
```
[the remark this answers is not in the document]
dmitrii-beliakov: I copied it from another dto where they have 3 latency: ms, ns, and max. I eventually removed 2, but this artefact remained. As a matter of fact, it works when passing both labels. I suspect that the serializer uses lombok's builder and if not possible this constructor. I am investigating if I can remove the constructor at all.

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/payload/network/PathValidationPayload.java
```diff
@@ -0,0 +1,80 @@
...
+            @JsonProperty("max_latency_tier2") Long latencyTier2ms,
```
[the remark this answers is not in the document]
dmitrii-beliakov: The JSON property is what a user enters and there are other places where it is without any suffixes (in flow, in get paths). I made it the same to be consistent with other APIs, but for internal usages I keep it with ms, to know the meaning of that long. Probably, this long really should be Duration or some similar class

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/PathsBolt.java
```diff
@@ -46,7 +47,8 @@ public PathsBolt(PersistenceManager persistenceManager, PathComputerConfig pathC
     public void init() {
         super.init();
 
-        pathService = new PathsService(repositoryFactory, pathComputerConfig);
+        pathService = new PathsService(repositoryFactory, pathComputerConfig, repositoryFactory.createIslRepository(),
+                repositoryFactory.createFlowRepository());
```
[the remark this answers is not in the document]
dmitrii-beliakov: Good point, I guess nothing deter from creating the repos inside the service.

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,513 @@
...
+
+    private boolean isDiverseWithFlowValidationRequired(PathValidationData pathValidationData) {
+        return StringUtils.isNotBlank(pathValidationData.getDiverseWithFlow());
+    }
+
+    private boolean isBandwidthValidationRequired(PathValidationData pathValidationData) {
+        return pathValidationData.getBandwidth() != null
+                && pathValidationData.getBandwidth() != 0
+                && (pathValidationData.getPathComputationStrategy() == null
+                || pathValidationData.getPathComputationStrategy() == COST_AND_AVAILABLE_BANDWIDTH);
```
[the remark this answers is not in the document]
dmitrii-beliakov: hm, could you tell what this strategy means?

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,513 @@
...
+    }
+
+    private boolean isLatencyTier2ValidationRequired(PathValidationData pathValidationData) {
+        return pathValidationData.getLatencyTier2() != null
+                && !pathValidationData.getLatencyTier2().isZero()
+                && (pathValidationData.getPathComputationStrategy() == null
```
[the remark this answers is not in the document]
dmitrii-beliakov: I made it so that we can execute all validations if we don't know what strategy to use. With defaulting it's trickier to interpret results because it is not clear which validations have been executed, or why not: there is no strategy in the response. 
Do you want to add this strategy defaulting requirement to this task?

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
dmitrii-beliakov: We could. In the beginning, I made these methods without dependencies, so that we can move them easily to other utility classes if this class grows too big and I wasn't sure if we want to have them static and if I want to use repos all the time or use some caching, or hide where the actual data came from (make methods like findByIds right in repositoryData and hide where it came from). So I thought that from this point it is easier to go in any direction.
What's your recommendation here?

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
dmitrii-beliakov: I thought it's better to validate it than fail with some exception and don't need to make an assumption that some other validation has been executed before this one.

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/payload/network/PathValidationPayload.java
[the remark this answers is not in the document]
niksv: I'm talked about Json properties. You are added ms for layency, but not for tier2.
```
@JsonProperty("latency_ms")
@JsonProperty("max_latency_tier2")
```            

But, anyway, you already fixed it

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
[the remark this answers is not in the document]
niksv: It means that we are comparing ISLs by cost first. If cost is equal we are comparing them be available bandwidth. 
It's fighting with unbalanced bandwidth on ISLs

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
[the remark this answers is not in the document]
niksv: But our PCE can't work without strategy. If you think that it's trickier to interpret results with defaulting we can mark `strategy` param as mandatory. 

But mostly I'm looking at current PCE. For example. An operator want to check some path and send a requrst without strategy, but wit latency. PCE says that latany is too small and the path is bad. But when the operator send same request to flow create API - flow will be created because default strategy is `cost` and it doesn't care about latency

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
niksv: I would recommend to use this.repository just to do not make current code more complicated. If we will need to move it to some other class - we will path repos inside

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
niksv: okay

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/payload/network/PathValidationPayload.java
```diff
@@ -0,0 +1,59 @@
...
+
+@Value
+@Builder
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class PathValidationPayload {
+    @JsonProperty("bandwidth")
+    Long bandwidth;
+
+    @JsonProperty("max_latency")
+    @PositiveOrZero(message = "max_latency cannot be negative")
```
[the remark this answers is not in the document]
dmitrii-beliakov: I am using it in other places, but here JSON keys differ from variable names.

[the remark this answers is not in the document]
niksv: okay

niksv:
@dmitrii-beliakov Hi, some time ago I talked with @pkazlenka  about tests for this PR and we found out that real path computation able to intersect with diversity flow, but this validation will throw an error. I'm wondering that we can miss something else. Also we can miss something in future when some one update PCE, but not this validation. Maybe we can try to reuse PCE methods? We allready talked about it and you said that it will be hard to implement. I created a POC of it. Could you please take a look and tell me if you will find some problems in this POC.

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,513 @@
...
+
+        Optional<Flow> diverseFlow = flowRepository.findById(inputData.getPath().getDiverseWithFlow());
+        if (!diverseFlow.isPresent()) {
+            return Collections.singleton(getNoDiverseFlowFoundError(inputData));
+        }
+
+        if (diverseFlow.get().getData().getPaths().stream()
+                .map(FlowPath::getSegments)
+                .flatMap(List::stream)
+                .anyMatch(pathSegment -> inputData.getSegment().getSrcSwitchId().equals(pathSegment.getSrcSwitchId())
```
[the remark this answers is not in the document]
pkazlenka: Yes. I've added test where we compare potential flow with path identical to existing one. It's easier to calculate expected path check errors in this case.

dmitrii-beliakov:
@niksv regarding the reusing available network class and the PCE methods. I tried the same approach as in POC, but then I realized that some parts of the validation is located in the building of network while others are hidden in PCE. I then brought it to the discussion that it is going to be like a boolean: you either can create this path or there is one error out of many unknown. This approach still requires maintenance (for example, imagine there is a new feature that creates flows with trying to bring up inactive links, so we need to go to path check and update it, it's not going to be a maintenance-free). Secondly, if PCE builds the available network differently, then path check still might give different result even though it uses exactly the same method of PCE. So there will be need to keep it consistent as well, which probably is not so easy to do. There were some other details, I already forgot, but these were the main reasons to switch to another approach.
The validation approach is different: it tells facts about the provided path. PCE implementation could be one or another, but the fact that there is an intersection between two path is the fact. What you are saying about the intersect with a diversity flow doesn't seem like a problem.
I think what could be done is to add this requirement of getting diagnostics on trying to create a path separately after all validations are done (and if it makes sense to continue)

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,580 @@
...
+    }
+
+    private Optional<Isl> findIslByEndpoints(SwitchId srcSwitchId, int srcPort, SwitchId destSwitchId, int destPort) {
+        IslEndpoints endpoints = new IslEndpoints(srcSwitchId.toString(), srcPort, destSwitchId.toString(), destPort);
+        if (!islCache.containsKey(endpoints)) {
+            islCache.put(endpoints,
+                    islRepository.findByEndpoints(srcSwitchId, srcPort, destSwitchId, destPort));
+        }
+        return islCache.get(endpoints);
+    }
```
niksv, line 118: hint: You can use lazy map 
https://github.com/telstra/open-kilda/blob/36a896231e8fb9c4c70c66a6ef20e54f4d4e8a1d/src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/service/FlowPathBuilder.java#L214-L220

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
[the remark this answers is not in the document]
niksv: You must remove strategy from here because the only reason of not to check bandwidth is empty getBandwidth()

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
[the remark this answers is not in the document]
niksv: I think you should remove line `pathValidationData.getPathComputationStrategy() == null`

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -0,0 +1,580 @@
...
+        return String.format(
+                "There is not enough bandwidth between end points: switch %s port %d and switch %s port %d"
+                        + " (reverse path). Requested bandwidth %d, but the link supports %d.",
+                data.getSegment().getDestSwitchId(), data.getSegment().getDestPort(),
+                data.getSegment().getSrcSwitchId(), data.getSegment().getSrcPort(),
+                data.getPath().getBandwidth(), actualBandwidth);
+    }
+
+    private String getNoLinkOnPath() {
+        return "Path latency cannot be calculated because there is no link at least at one path segment.";
```
niksv, line 505: I think you should specify which exactly link is missed. It will look like "Something went wrong" otherwise.

niksv (changes requested):
Are we checking that segments in API body connected to each other?

Can we have a positive response for a request with not connected segments?

niksv:
Regarding the POC of moving checks into PCE: we decided to do it as the second part of this feature

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
dmitrii-beliakov: I split the validation into per-segment and per-path. Per-path doesn't (shouldn't) look at the segments level and this is why there is no a specific segment in the message. It only says that per-path validation cannot be performed. 
The diagnostic information about specific segments is there, but on per-segment level. In case there are links missing, per-segment validation would report "there is no ISL here and here", while per-path validations report that it is not possible to execute them.

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
dmitrii-beliakov: done

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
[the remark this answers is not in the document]
dmitrii-beliakov: done

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
dmitrii-beliakov: done

niksv (approved):

PR #5079, opened 2023-02-08
fix flow_id in path from flow update request ignored

### LINKED ISSUE #5075: flow_id in the request line is ignored while flow update
Steps
case 1
1) create flow (e.g. with flow_id = flow1) - POST [/v2/flows]
2) make **wrong** put request to update it  - - PUT [/v2/flows/flow_not_match]
where body of the request contains flow_id = flow1
note: system shouldn't contain any flow with "flow_not_match" id

actual result: flow is updated successfully
expected result: 404 with the corresponding error message like "flow not found"

case 2
1) create flow (e.g. with flow_id = flow1) - POST [/v2/flows]
2) make  put request to update it  - - PUT [/v2/flows/flow1]
where body of the request **doesn't contains** flow_id = flow1
e.g. body should contain flow_id = flow_not_match
note: system shouldn't contain any flow with "flow_not_match" id

actual result: 404 with the corresponding error message like "flow not found"
expected result: flow should be updated, flow_id from the body should be ignored

actually we are taking flow_id from body, but it should be taken from the request line
also flow_id update is not supported, so it can be set as nullable in the body

pablomuri:
During the flow update request from v1 and v2 API the flow_id in the path was ignored and kilda was only updating the flow with the flow_id from the request body.

This commit fixes this by always replacing the flow_id in the request body with the one in the path at the service level in both, v1 and v2 APIs.

Closes #5075


niksv (approved):

pablomuri:
Fixed some unit test that was failing

artHalMirantis (approved):

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/controller/v1/FlowController.java
```diff
@@ -137,7 +137,7 @@ public CompletableFuture<FlowResponsePayload> deleteFlow(@PathVariable(name = "f
     @ResponseStatus(HttpStatus.OK)
     public CompletableFuture<FlowResponsePayload> updateFlow(@PathVariable(name = "flow-id") String flowId,
                                                              @RequestBody FlowUpdatePayload flow) {
-        return flowService.updateFlow(flow);
+        return flowService.updateFlow(flowId, flow);
```
dmitrii-beliakov, line 140: Flow Id is also present in the payload and it could differ from the path variable? I am wondering why don't we put the path variable inside the payload? Possibly raise an error if they are not the same

pablomuri: This was discussed in the issue https://github.com/telstra/open-kilda/issues/5075#issuecomment-1422517669

pablomuri:
As discussed offline, we decide to return a bad request error if the flow_id in the path and the one in the body are different.

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/service/impl/FlowServiceImpl.java
```diff
@@ -263,8 +264,8 @@ public CompletableFuture<FlowResponsePayload> updateFlow(final FlowUpdatePayload
     }
 
     @Override
-    public CompletableFuture<FlowResponseV2> updateFlow(FlowRequestV2 request) {
-        log.info("API request: Processing flow update: {}", request);
+    public CompletableFuture<FlowResponseV2> updateFlow(final String flowId, FlowRequestV2 request) {
+        log.info("API request: Update flow request for flow{}: {}", flowId, request);
```
niksv, line 268: Missed space `flow{}`

niksv (approved):

src-java/northbound-service/northbound/src/main/java/org/openkilda/northbound/service/impl/FlowServiceImpl.java
pablomuri: done

Issue #5061, opened 2023-02-02
Added ability two create HA-flows


niksv:
HA-flows looks like Y-flows but there is one difference. 
HA-flow has 3 endpoints: A-end, B-end, Z-end (shared).

All traffic from Z-end must be delivered to A-end and B-end.
Traffic from A-end and B-end must be joined and delivered to Z-end.

PR #5081, opened 2023-02-10
HA flows Part 2: Added DB objects

niksv:
Related to #5061


src-java/kilda-model/src/main/java/org/openkilda/model/HaFlow.java
```diff
@@ -0,0 +1,593 @@
...
+            }
+            setProtectedReversePathId(reversePath.getHaPathId());
+        } else {
+            setProtectedReversePathId(null);
+        }
+    }
+
+    private void validateEndpoints(HaFlowPath path) {
+        if (!path.getSharedSwitchId().equals(getSharedEndpoint().switchId)) {
+            throw new IllegalArgumentException(format("HA-path %s has shared endpoint switch id %s, but %s expected",
```
dmitrii-beliakov, line 217: ID is short for `identification`; it is an abbreviation, usually spelled upper case. 
An article is missing: "... has **_the_** shared ...". Passive voice in the second part: " ... but %s **_is_** expected"

src-java/kilda-model/src/main/java/org/openkilda/model/HaFlow.java
```diff
@@ -0,0 +1,593 @@
...
+                    path.getHaPathId(), path.getSharedSwitchId(), getSharedEndpoint().switchId));
+        }
+        Set<SwitchId> subFlowSwitchIds = getSubFlows().stream()
+                .map(HaSubFlow::getEndpointSwitchId)
+                .collect(Collectors.toSet());
+        Set<SwitchId> pathSwitchIds = path.getSubFlowSwitchIds();
+
+        if (!subFlowSwitchIds.equals(pathSwitchIds)) {
+            throw new IllegalArgumentException(format("HA-path %s must have %s switch endpoints, but has following "
```
dmitrii-beliakov, line 226: a common collocation: **_the_** following

src-java/kilda-model/src/main/java/org/openkilda/model/HaFlow.java
```diff
@@ -0,0 +1,593 @@
...
+                            + "endpoints: %s",
+                    path.getHaPathId(), subFlowSwitchIds, pathSwitchIds));
+        }
+    }
+
+    private void validateForwardPath(HaFlowPath haPath) {
+        validateEndpoints(haPath);
+        if (!haPath.isForward()) {
+            throw new IllegalArgumentException(format("Cookie %s of HA-path %s must have forward direction",
```
dmitrii-beliakov, line 235: To me, in this case, "direction" is a countable noun (meaning a route or a path, right?), so, _**a**_ forward direction

src-java/kilda-model/src/main/java/org/openkilda/model/HaFlow.java
```diff
@@ -0,0 +1,593 @@
...
+    private void validateReversePath(HaFlowPath haPath) {
+        validateEndpoints(haPath);
+        if (!haPath.isReverse()) {
+            throw new IllegalArgumentException(format("Cookie %s of HA-path %s must have reverse direction",
+                    haPath.getCookie(), haPath.getHaPathId()));
+        }
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
dmitrii-beliakov, line 249: out of curiosity: why not to use the annotations to generate equals and hashCode?

src-java/kilda-model/src/main/java/org/openkilda/model/HaFlow.java
```diff
@@ -0,0 +1,593 @@
...
+        FlowStatus haFlowStatus = null;
+        for (HaSubFlow subFlow : getSubFlows()) {
+            FlowStatus subFlowStatus = subFlow.getStatus();
+            if (subFlowStatus == FlowStatus.IN_PROGRESS) {
+                haFlowStatus = FlowStatus.IN_PROGRESS;
+                break;
+            }
+            if (haFlowStatus == null) {
+                haFlowStatus = subFlowStatus;
+            } else if (haFlowStatus == FlowStatus.DOWN && subFlowStatus != FlowStatus.DOWN
```
dmitrii-beliakov, line 305: What about the case when both are `DOWN`?

src-java/kilda-model/src/main/java/org/openkilda/model/HaFlow.java
```diff
@@ -0,0 +1,593 @@
...
+        @Override
+        public Set<HaSubFlow> getSubFlows() {
+            return Collections.unmodifiableSet(subFlows);
+        }
+
+        @Override
+        public void setSubFlows(Set<HaSubFlow> subFlows) {
+            for (HaSubFlow subFlow : this.subFlows) {
+                boolean keepSubFlow = subFlows.stream()
+                        .anyMatch(n -> n == subFlow || n.getHaSubFlowId().equals(subFlow.getHaSubFlowId()));
```
dmitrii-beliakov, line 482: What's the meaning of the first condition? `==` is to check whether it is the same object? But then the next condition is true, except when subFlowId is null. I mean, is the first one really needed here?

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/frames/HaFlowFrame.java
```diff
@@ -0,0 +1,373 @@
...
+    @Override
+    @Property(PROTECTED_REVERSE_PATH_ID_PROPERTY)
+    @Convert(PathIdConverter.class)
+    public abstract void setProtectedReversePathId(PathId protectedReversePathId);
+
+    @Override
+    public Optional<HaSubFlow> getSubFlow(String subFlowId) {
+        if (subFlows == null) {
+            // init the cache map with sub flows.
+            getSubFlows();
```
dmitrii-beliakov, line 259: no concurrency issues here? When two threads start initializing the map, then what?

dmitrii-beliakov (approved):

src-java/kilda-model/src/main/java/org/openkilda/model/HaFlow.java
dmitrii-beliakov: I imagine this sort of "direction" as an arrow. There one direction: forward, to the right, there is another direction, reverse, to the left. So, two directions in total. I can count them -> countable.
Direction could be uncountable when it means an abstract idea. It's something like when we speak about sea turtles and their ability to know where they are and where to go without a compass, or they have a sense of direction (no specific arrow here, here it means an ability to know where to go and don't get lost on the way). 
For switches, I think it is the first case.

src-java/kilda-model/src/main/java/org/openkilda/model/HaFlow.java
dmitrii-beliakov: Yeah, I see, thanks for the explanation.

src-java/kilda-model/src/main/java/org/openkilda/model/HaFlow.java
niksv: Db objects are tricky. Each object has inner field `data` which contains actual data fields. It we will use @EqualsAndHashCode it will compare whole `data` object. But we need to compare a set of data fields. Also `data` is an interface which has 2 implementations: POJO and object, connected with DB. Both implementation must have equal `equals()` so to do not duplicate equal and hashcode for each implementation we can add `equal()` here manually.

pkazlenka:
Devs internal. Considered tested as long as autotests pass.

src-java/kilda-model/src/main/java/org/openkilda/model/HaFlow.java
niksv: fixed

PR #5082, opened 2023-02-10
HA flows Part 3: Added DB repositories

niksv:
Related to #5061


src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaHaFlowPathRepository.java
```diff
@@ -0,0 +1,115 @@
...
+                .toListExplicit(HaFlowPathFrame.class).stream()
+                .map(HaFlowPath::new)
+                .collect(Collectors.toList());
+    }
+
+    @Override
+    public Optional<HaFlowPath> remove(PathId haFlowPathId) {
+        TransactionManager transactionManager = getTransactionManager();
+        if (transactionManager.isTxOpen()) {
+            // This implementation removes dependant entities (segments, haSubFlowEdges) in a separate transactions,
```
dmitrii-beliakov, line 73: Either "in separate transactions," or "in a separate transaction,"

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaHaFlowRepository.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    @Override
+    protected HaFlowFrame doAdd(HaFlowData data) {
+        HaFlowFrame frame = KildaBaseVertexFrame.addNewFramedVertex(
+                framedGraph(), HaFlowFrame.FRAME_LABEL, HaFlowFrame.class);
+        HaFlow.HaFlowCloner.INSTANCE.copyWithoutSubFlowsAndPaths(data, frame);
+        frame.setSubFlows(data.getSubFlows().stream()
+                .peek(subFlow -> {
+                    if (!(subFlow.getData() instanceof HaSubFlowFrame)) {
```
dmitrii-beliakov, line 106: Why do you use the condition via ~not? It allows String and Integer to pass the condition.
Also, if an object in the stream is of type HaSubFlowFrame, it is not added to subFlor repo, but collected to the set later, is that intended? 
I'd create a set and then use it for add to repo and the setter.

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaHaFlowPathRepositoryTest.java
```diff
@@ -0,0 +1,381 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public class FermaHaFlowPathRepositoryTest extends InMemoryGraphBasedTest {
+    public static final FlowSegmentCookie COOKIE_1 = FlowSegmentCookie.builder().flowEffectiveId(1).build();
```
dmitrii-beliakov, line 58: probably, all fields here could be private

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaHaFlowPathRepositoryTest.java
```diff
@@ -0,0 +1,381 @@
...
+            PathId pathId, long bandwidth, FlowSegmentCookie cookie, String sharedGroupId, MeterId sharedMeterId,
+            MeterId yPointMeterId, Switch sharedSwitch, SwitchId yPointSwitchId, GroupId yPointGroupId) {
+        HaFlowPath path = buildHaFlowPath(
+                pathId, bandwidth, cookie, sharedGroupId, sharedMeterId, yPointMeterId, sharedSwitch,
+                yPointSwitchId, yPointGroupId);
+        haFlowPathRepository.add(path);
+        return path;
+    }
+
+    static HaFlowPath buildHaFlowPath(
```
dmitrii-beliakov, line 348: could be private

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaHaFlowRepositoryTest.java
```diff
@@ -0,0 +1,301 @@
...
+        assertHaFlow(HA_FLOW_ID_2, SWITCH_ID_2, PORT_2, VLAN_2, INNER_VLAN_2, LATENCY_3, LATENCY_4, BANDWIDTH_2,
+                FlowEncapsulationType.VXLAN, PRIORITY_2, DESCRIPTION_2, PathComputationStrategy.LATENCY,
+                FlowStatus.IN_PROGRESS, false, false, false, false, false, foundFlow2.get());
+        assertEquals(0, foundFlow2.get().getSubFlows().size());
+        assertEquals(0, foundFlow2.get().getPaths().size());
+    }
+
+    @Test
+    public void findByIdNoExistentFlowTest() {
+        assertFalse(haFlowRepository.findById("123").isPresent());
```
dmitrii-beliakov, line 194: A bit tricky question what this test case proves. 
I suggest you combine it with the create or remove. You can assert that something was in the repo and it is not there after the removal. Or something was not in the repo and it is there after the creation. Or both.

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaHaFlowRepositoryTest.java
```diff
@@ -0,0 +1,301 @@
...
+        assertTrue(haFlowRepository.findById(haFlow2.getHaFlowId()).isPresent());
+        assertEquals(1, haFlowRepository.findAll().size());
+        assertEquals(0, haFlowPathRepository.findAll().size());
+        assertEquals(0, haSubFlowRepository.findAll().size());
+        assertEquals(0, pathSegmentRepository.findByPathId(path1.getHaPathId()).size());
+        assertEquals(0, pathSegmentRepository.findByPathId(path2.getHaPathId()).size());
+    }
+
+    private void assertPathsFlows(Collection<HaFlowPath> actualPaths, HaFlowPath... expectedPaths) {
+        Map<PathId, HaFlowPath> actualMap = haFlowPathsToMap(actualPaths);
```
dmitrii-beliakov, line 235: This assertion doesn't take into consideration redundant paths. You assert the size in `createHaFlowTest`, but don't assert in `findByIdHaFlowTest`. I think you can simplify it by converting the input collection to a set, as well as put the expected vars into set, then use assert equals on sets. 
A separate question is about uniqueness: actualPaths can have dupes, but they disappear once it is converted to the map.

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaHaSubFlowRepositoryTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+        Optional<HaSubFlow> subFlow3 = haSubFlowRepository.findById(SUB_FLOW_ID_3);
+        assertTrue(subFlow3.isPresent());
+        assertSubFlow(SUB_FLOW_ID_3, SWITCH_ID_3, PORT_3, VLAN_3, INNER_VLAN_3, null, subFlow3.get());
+
+        assertFalse(haSubFlowRepository.findById(SUB_FLOW_ID_4).isPresent());
+    }
+
+    @Test
+    public void haSubExistTest() {
```
dmitrii-beliakov, line 128: haSub**Flows**ExistTest

Fedichk (approved):

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaHaFlowPathRepositoryTest.java
dmitrii-beliakov: Maybe it's not a big a deal and just a matter of habits and style. Nevertheless, in my experience, dependencies between tests brings more harm than benefits (btw the same for inheritance in tests). I see it like dependency between impl modules. By its nature, those methods are some helpers or utility methods and they don't belong to the class responsible for testing. Those utility methods have better reusability when they are separated from tests. It could be a static method in a helper class or a service class via composition if there is a need to set up some other objects like other repositories.
I see it in the similar way as we all can write methods here and there like check whether the value is null and then set a default one, or we can extract it to some utility class called ObjectUtils and use it everywhere. Yeah, another point is when somebody wants to extract it, they might find it's already been done before.

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaHaFlowRepositoryTest.java
dmitrii-beliakov: thanks! in fact, equals for sets implies they have the same size, otherwise at least one element is not present in other set. So with the sets approach line 280 is redundant.

src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaHaFlowPathRepositoryTest.java
niksv: Good point. I moved builders into separate class

src-java/kilda-persistence-tinkerpop/src/main/java/org/openkilda/persistence/ferma/repositories/FermaHaFlowRepository.java
dmitrii-beliakov: Thanks for explaining! 👍
I guess this is `instance of` who drew my attention,  it is sometimes tricky to get the meaning. I'd hide this checking under the super interface to get if this object is transient or not. Just for readability, something like `if(subFlow.isTransientObject()) then persist it`. But I believe this current idiom is used for all other frames? so probably not worth to touch it in just one place.

dmitrii-beliakov: About using not, I meant the theoretical/technical possibility. If the type is known, then, as an option, you can revert the if-clause: `if ((subFlow.getData() instanceof HaSubFlowDataImpl))`. With `not` it's like `String` is not an instance of `HaSubFlowFrame`, an absurd example, but technically true.

niksv: Yes, We are using same instanceof in all Frame repos, so we need to change all of them for consistency

dmitrii-beliakov (approved):

artHalMirantis:
tested by https://github.com/telstra/open-kilda/pull/5089

PR #5086, opened 2023-02-15
Add max depth to the error while path computation

### LINKED ISSUE #5085: Add max depth in reason when path can't be found
During the path computation if the path has more hops than the maxAllowedDepth property (35 by default) the PCE will raise an exception with a description saying "Allowed depth exceeded" which will be returned as an error by the northbound. 

We need to include in the message error also the max allowed depth. We can't include the actual depth of the path since following the algorithm of the path computation because it will fail when the depth >= max depth.

https://github.com/telstra/open-kilda/blob/97c4998939f7c99be379a7efc5586cb107a3f0bf/src-java/kilda-pce/src/main/java/org/openkilda/pce/finder/FailReasonType.java#L20

https://github.com/telstra/open-kilda/blob/97c4998939f7c99be379a7efc5586cb107a3f0bf/src-java/kilda-pce/src/main/java/org/openkilda/pce/PathComputerConfig.java#L31-L33

pablomuri:
During the path computation, if the path has more hops than the max depth property, it raises an error. This commit extends this error to include the maximun depth allowed.

Closes #5085

### DIFF
--- a/src-java/kilda-pce/src/main/java/org/openkilda/pce/finder/BestWeightAndShortestPathFinder.java
+++ b/src-java/kilda-pce/src/main/java/org/openkilda/pce/finder/BestWeightAndShortestPathFinder.java
@@ -425,7 +425,8 @@ public class BestWeightAndShortestPathFinder implements PathFinder {
                 }
                 if (current.allowedDepth <= 0) {
                     reasons.put(FailReasonType.ALLOWED_DEPTH_EXCEEDED,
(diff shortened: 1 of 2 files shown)


dmitrii-beliakov (approved):

niksv (approved):

durgakako (approved):

pablomuri:
This change can't be functionally tested in our current pipeline

PR #5090, opened 2023-02-20
Bugfix/2660 empty rule none drop fix

### LINKED ISSUE #2660: GET Switch Rules is not showing proper rule information in the 'instructiion -> none' field
**Step to reproduce:**

1. get switch rules via NB `GET /v1/switches/{switch-id}/rules`
2. get switch rules via FL http://127.0.0.1:8081/wm/core/switch/00:00:00:00:00:00:00:02/flow/json
3. compare rules from item 1 against item 2;

**Actual result**
FL:
```
instructions: {
none: "drop"
}
```
NB:
```
"instructions": {
   "apply_actions": null,
   "none": null,
   "instruction_goto_meter": null,
   "instruction_goto_table": null
},
```

**Expected result:**
The `none` field should be the same

IvanChupin:
Closes #2660

Sets none = "drop" value for empty FlowInstructions
[/v1/switches/{switch-id}/rules]

### DIFF
--- a/confd/vars/docker-compose.yaml
+++ b/confd/vars/docker-compose.yaml
@@ -1,5 +1,5 @@
 ---
-#no_gui: true
+no_gui: true
 #no_kibana: true
 #no_floodlight_2: true
 #no_floodlight_stats: true
@@ -8,10 +8,10 @@
 #no_opentsdb: true
 #no_elasticsearch: true
 #no_logstash: true
-#no_grpc_stub: true
+no_grpc_stub: true
 no_server42_server: true
-#no_server42_control: true
-#no_server42_stats: true
-#no_server42_server_stub: true
-#no_server42_storm_stub: true
+no_server42_control: true
+no_server42_stats: true
+no_server42_server_stub: true
+no_server42_storm_stub: true
 single_orientdb: true
--- a/src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/converter/OfFlowStatsMapper.java
+++ b/src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/converter/OfFlowStatsMapper.java
@@ -17,7 +17,6 @@ package org.openkilda.floodlight.converter;
 
 import static java.util.stream.Collectors.toList;
 
-import org.apache.commons.collections4.CollectionUtils;
 import org.openkilda.messaging.info.rule.FlowApplyActions;
 import org.openkilda.messaging.info.rule.FlowApplyActions.FlowApplyActionsBuilder;
 import org.openkilda.messaging.info.rule.FlowCopyFieldAction;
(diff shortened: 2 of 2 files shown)


niksv (commented):
looks good but could you please add some unit tests?

IvanChupin:
> looks good but could you please add some unit tests?

Thanks, fixed, unit tests added

niksv:
@IvanChupin unit tests are red now

src-java/floodlight-service/floodlight-modules/src/test/java/org/openkilda/floodlight/converter/OfFlowStatsMapperTest.java
```diff
@@ -135,29 +136,29 @@ public void testFlowEntry() {
...
+                = OfFlowStatsMapper.INSTANCE.toFlowInstructions(buildInstruction());
+
+        assertEquals(expectedNoneEmptyFlowInstructions, actualNoneEmptyFlowInstructions);
+
+        FlowInstructions expectedEmptyFlowInstructions = FlowInstructions.builder().none(NONE_DROP).build();
+        List<OFInstruction> emptyInstructions = Collections.emptyList();
+        FlowInstructions actualEmptyInstructions = OfFlowStatsMapper.INSTANCE.toFlowInstructions(emptyInstructions);
+
+        assertEquals(expectedEmptyFlowInstructions, actualEmptyInstructions);
+    }
```
niksv, line 160: Here we have 2 test data for 2 test cases. Please split this test into 2 tests - one with some flow instructions and one with empty instructions

README.md
```diff
@@ -233,6 +233,49 @@ To check how debugging works we need to:
...
+```
+docker exec -it floodlight_1(2) bash
+```
+
+the entrypoint.sh should contain this argument
+```
+"-agentlib:jdwp=transport=dt_socket,address=50506,suspend=n,server=y” argument.
+```
+
+### Debug for a couple components
```
niksv, line 278: Please move floodlight debug into separate PR. Why do we need it? If something will go wrong with your PR - it can be reverted. If it will be reverted - this Readme will be reverted too. But it is not related to the bug and we can keep it.

niksv (changes requested):

README.md
nikitamarchenko: Probably in separate readme

src-java/floodlight-service/floodlight-modules/src/test/java/org/openkilda/floodlight/converter/OfFlowStatsMapperTest.java
IvanChupin: thanks, fixed

README.md
IvanChupin: @nikitamarchenko  thanks, should I put this under the floodlight readme.md file?

niksv (approved):

dmitrii-beliakov (approved):

pkazlenka:
Bugfix was verified manually. No need to add functional test, unit test is enough in this case.

PR #5108, opened 2023-03-01
Improve performance of SwitchConnectedDevices storing

### LINKED ISSUE #5051: Improve performance of SwitchConnectedDevices storing
SwitchConnectedDevices DB model has several fields `switch_id`, `port_id`, 'vlan' etc and `unique_index` based on these fields.
Also this model has indexes for each of base fields. So while storing new object we must update 5-7 indexes instead of one unique index.

We have 4 device read operations:
1. by switch_id
2. by flow_id
3. by arp unique index
4. by lldp unique index

it means that following indexes are useless
```
switch_connected_device.chassis_id
switch_connected_device.ip_address 
switch_connected_device.mac_address
switch_connected_device.port_id
switch_connected_device.port_number
switch_connected_device.type
switch_connected_device.vlan
```

So the task is:
1. update methods `findLldpByUniqueFieldCombination` `findArpByUniqueFieldCombination`. They must use unique index for searching.
2. remove useless indexes
3. check that all changes above improve performance. (Main performance mark is Orient DB CPU load, optional performance mark is capacity of PacketBolt in ConnectedDevicesTopology)

Fedichk:
Closes #5051
In this PR we:
- removes redundant indexes in DB
- updates Oversize property for switch_connected_device class
- add cache and update logic for SwitchConnectedDevice entity to reduce the number of transactions  

For measure, we used 100 connected devices with ≈100 update messages per device (10000 messages).
Initially, the processing time was ≈24-25 seconds after the changes ≈12.5 seconds.

### DIFF
--- a/src-java/connecteddevices-topology/connecteddevices-storm-topology/src/main/java/org/openkilda/wfm/topology/connecteddevices/service/PacketService.java
+++ b/src-java/connecteddevices-topology/connecteddevices-storm-topology/src/main/java/org/openkilda/wfm/topology/connecteddevices/service/PacketService.java
@@ -18,8 +18,8 @@ package org.openkilda.wfm.topology.connecteddevices.service;
 import static java.lang.String.format;
 import static org.openkilda.model.ConnectedDeviceType.ARP;
 import static org.openkilda.model.ConnectedDeviceType.LLDP;
-import static org.openkilda.model.SwitchConnectedDevice.createUniqueArpIndex;
-import static org.openkilda.model.SwitchConnectedDevice.createUniqueLldpIndex;
+import static org.openkilda.model.SwitchConnectedDevice.buildUniqueArpIndex;
+import static org.openkilda.model.SwitchConnectedDevice.buildUniqueLldpIndex;
 import static org.openkilda.model.cookie.Cookie.ARP_INGRESS_COOKIE;
 import static org.openkilda.model.cookie.Cookie.ARP_INPUT_PRE_DROP_COOKIE;
 import static org.openkilda.model.cookie.Cookie.ARP_POST_INGRESS_COOKIE;
@@ -87,7 +87,7 @@ public class PacketService {
             return;
         }
 
-        String uniqueIndex = createLldpUniqueIndex(data, flowRelatedData.getOriginalVlan());
+        String uniqueIndex = buildLldpUniqueIndex(data, flowRelatedData.getOriginalVlan());
         SwitchConnectedDevice device = Optional.ofNullable(switchConnectedDeviceCache.get(uniqueIndex))
                 .orElseGet(() -> switchConnectedDeviceRepository.findLldpByUniqueIndex(uniqueIndex).orElse(null));
 
@@ -117,7 +117,7 @@ public class PacketService {
             return;
         }
 
-        String uniqueIndex = createArpUniqueIndex(data, flowRelatedData.getOriginalVlan());
+        String uniqueIndex = buildArpUniqueIndex(data, flowRelatedData.getOriginalVlan());
         SwitchConnectedDevice device = Optional.ofNullable(switchConnectedDeviceCache.get(uniqueIndex))
                 .orElseGet(() -> switchConnectedDeviceRepository.findArpByUniqueIndex(uniqueIndex).orElse(null));
 
@@ -298,14 +298,14 @@ public class PacketService {
     }
 
     @VisibleForTesting
-    String createLldpUniqueIndex(LldpInfoData data, int vlan) {
-        return createUniqueLldpIndex(data.getSwitchId(), data.getPortNumber(), vlan, data.getMacAddress(),
+    String buildLldpUniqueIndex(LldpInfoData data, int vlan) {
+        return buildUniqueLldpIndex(data.getSwitchId(), data.getPortNumber(), vlan, data.getMacAddress(),
                 data.getChassisId(), data.getPortId());
     }
 
     @VisibleForTesting
-    String createArpUniqueIndex(ArpInfoData data, int vlan) {
-        return createUniqueArpIndex(data.getSwitchId(), data.getPortNumber(), vlan, data.getMacAddress(),
+    String buildArpUniqueIndex(ArpInfoData data, int vlan) {
+        return buildUniqueArpIndex(data.getSwitchId(), data.getPortNumber(), vlan, data.getMacAddress(),
                 data.getIpAddress());
     }
 
@@ -418,7 +418,7 @@ public class PacketService {
                 .source(flowRelatedData.getSource())
                 .build();
         switchConnectedDeviceRepository.add(connectedDevice);
-        switchConnectedDeviceCache.put(createLldpUniqueIndex(data, vlan), connectedDevice);
+        switchConnectedDeviceCache.put(buildLldpUniqueIndex(data, vlan), connectedDevice);
     }
 
     private void createArpDevice(ArpInfoData data, FlowRelatedData flowRelatedData) {
@@ -445,7 +445,7 @@ public class PacketService {
                 .source(flowRelatedData.getSource())
                 .build();
         switchConnectedDeviceRepository.add(connectedDevice);
-        switchConnectedDeviceCache.put(createArpUniqueIndex(data, vlan), connectedDevice);
+        switchConnectedDeviceCache.put(buildArpUniqueIndex(data, vlan), connectedDevice);
     }
 
     private String getPacketName(ConnectedDevicePacketBase data) {
--- a/src-java/connecteddevices-topology/connecteddevices-storm-topology/src/main/java/org/openkilda/wfm/topology/connecteddevices/service/PacketService.java
+++ b/src-java/connecteddevices-topology/connecteddevices-storm-topology/src/main/java/org/openkilda/wfm/topology/connecteddevices/service/PacketService.java
@@ -1,4 +1,4 @@
-/* Copyright 2019 Telstra Open Source
+/* Copyright 2023 Telstra Open Source
  *
  *   Licensed under the Apache License, Version 2.0 (the "License");
  *   you may not use this file except in compliance with the License.
@@ -15,6 +15,7 @@
 
 package org.openkilda.wfm.topology.connecteddevices.service;
 
+import static java.lang.String.format;
 import static org.openkilda.model.ConnectedDeviceType.ARP;
 import static org.openkilda.model.ConnectedDeviceType.LLDP;
 import static org.openkilda.model.cookie.Cookie.ARP_INGRESS_COOKIE;
@@ -50,6 +51,8 @@ import lombok.Value;
 import lombok.extern.slf4j.Slf4j;
 
 import java.time.Instant;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Optional;
 
 @Slf4j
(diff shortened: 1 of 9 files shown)


src-java/connecteddevices-topology/connecteddevices-storm-topology/src/main/java/org/openkilda/wfm/topology/connecteddevices/service/PacketService.java
```diff
@@ -75,52 +80,71 @@ public PacketService(PersistenceManager persistenceManager) {
...
+        SwitchConnectedDevice device;
 
-            if (device == null) {
-                return;
-            }
+        if (cachedDevice != null) {
+            device = cachedDevice;
+        } else {
+            Optional<SwitchConnectedDevice> optionalDevice = switchConnectedDeviceRepository
+                    .findArpByUniqueIndex(createUniqueArpIndex(data, flowRelatedData.originalVlan));
```
niksv, line 135: minor: I think you can save createUniqueArpIndex into some variable just to do not call it two times

src-java/connecteddevices-topology/connecteddevices-storm-topology/src/test/java/org/openkilda/wfm/topology/connecteddevices/service/PacketServiceTest.java
```diff
@@ -510,4 +510,14 @@ private ArpInfoData createArpInfoData() {
...
+
+    private String createUniqueLldpIndex(LldpInfoData data, int vlan) {
+        return String.format("%s_%s_%s_%s_%s_%s_%s", data.getSwitchId(), data.getPortNumber(), LLDP,
+                vlan, data.getMacAddress(), data.getChassisId(), data.getPortId());
+    }
+
+    private String createUniqueArpIndex(ArpInfoData data, int vlan) {
+        return String.format("%s_%s_%s_%s_%s_%s", data.getSwitchId(), data.getPortNumber(), ARP,
+                vlan, data.getMacAddress(), data.getIpAddress());
+    }
```
niksv, line 522: We have 3 places which creates such index. Maybe we should move it to some common place? For example SwitchConnectedDevice?

Fedichk: Actually, in one of these 3 places, the method has a different signature and src-java/kilda-persistence-tinkerpop/src/test/java/org/openkilda/persistence/ferma/repositories/FermaSwitchConnectedDevicesRepositoryTest.java doesn't have a dependency on ArpInfoData/LldpInfoData. So, I updated the method usage similarly to `findFlowRelatedDataForOneSwitchFlow` method.

pablomuri:
Could you please explain a bit more about what are you doing on each commit? I mean, extend a bit more each commit-msg.
Thanks!

src-java/connecteddevices-topology/connecteddevices-storm-topology/src/test/java/org/openkilda/wfm/topology/connecteddevices/service/PacketServiceTest.java
niksv: Okay, What we have now:
1. `createUniqueLldpIndex` in `PacketService`
2. `createUniqueLldpIndex` in `FermaSwitchConnectedDevicesRepositoryTest`
3. `recalculateUniqueIndex` in `SwitchConnectedDeviceFrame`

each method does the same. If we will change one - we must change all. Also you a user will need to read devices in some separate - he will must to create another one method `createUniqueLldpIndex` in this separate module. 

What I'm proposing: 
1. Create a static method in SwitchConnectedDevice. something like `createUniqueLldpIndex(switcchid, port, .....`
2. Use this method everywhere

src-java/connecteddevices-topology/connecteddevices-storm-topology/src/main/java/org/openkilda/wfm/topology/connecteddevices/service/PacketService.java
```diff
@@ -75,52 +80,70 @@ public PacketService(PersistenceManager persistenceManager) {
...
-            if (device == null) {
-                return;
-            }
+        if (cachedDevice != null) {
+            device = cachedDevice;
+        } else {
+            Optional<SwitchConnectedDevice> optionalDevice = switchConnectedDeviceRepository
+                    .findLldpByUniqueIndex(uniqueIndex);
+            device = optionalDevice.orElse(null);
+        }
```
dmitrii-beliakov, line 98: too many nulls. I'd make these lines 89-98 shorter, something like: 
```
SwitchConnectedDevice device = Optional.ofNullable(switchConnectedDeviceCache.get(uniqueIndex))
                .orElseGet(() -> switchConnectedDeviceRepository.findLldpByUniqueIndex(uniqueIndex).orElse(null));
```
the same for similar usage. (Could be even combined with the next if-else, but I don't know how to do that nicely using java 8)

src-java/connecteddevices-topology/connecteddevices-storm-topology/src/main/java/org/openkilda/wfm/topology/connecteddevices/service/PacketService.java
```diff
@@ -366,55 +401,49 @@ private Flow getFlowBySwitchIdInPortAndOutVlan(SwitchId switchId, int inPort, in
...
-            return device.get();
-        }
-
-        Optional<Switch> sw = switchRepository.findById(data.getSwitchId());
+    private void createLldpDevice(LldpInfoData data, FlowRelatedData flowRelatedData) {
+        Optional<Switch> optionalSwitch = switchRepository.findById(data.getSwitchId());
+        int vlan = flowRelatedData.getOriginalVlan();
 
-        if (!sw.isPresent()) {
+        if (!optionalSwitch.isPresent()) {
```
dmitrii-beliakov, line 408: I'd pull this up so that `create...` method has only one responsibility: to create. And if it's not possible to create a device, then don't call create method. But maybe this is too much for this PR.

src-java/connecteddevices-topology/connecteddevices-storm-topology/src/main/java/org/openkilda/wfm/topology/connecteddevices/service/PacketService.java
Fedichk: It is a great idea. Will update. Thank you!

src-java/connecteddevices-topology/connecteddevices-storm-topology/src/main/java/org/openkilda/wfm/topology/connecteddevices/service/PacketService.java
```diff
@@ -75,52 +80,70 @@ public PacketService(PersistenceManager persistenceManager) {
...
-            if (flowRelatedData == null) {
-                return;
-            }
+        FlowRelatedData flowRelatedData = findFlowRelatedData(data);
+        if (flowRelatedData == null) {
+            return;
+        }
 
-            SwitchConnectedDevice device = getOrCreateLldpDevice(data, flowRelatedData.originalVlan);
+        String uniqueIndex = createUniqueLldpIndex(data, flowRelatedData.getOriginalVlan());
```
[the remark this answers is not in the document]
Fedichk: Would something like createUniqueLldpIndexString / createUniqueLldpIndexValue / your option be better?

src-java/connecteddevices-topology/connecteddevices-storm-topology/src/main/java/org/openkilda/wfm/topology/connecteddevices/service/PacketService.java
Fedichk: You are right about responsibilities. But I`ve tried to make minimum logic changes and yes such refactoring could cause many changes. So, I would prefer to leave it in its current state.

src-java/connecteddevices-topology/connecteddevices-storm-topology/src/main/java/org/openkilda/wfm/topology/connecteddevices/service/PacketService.java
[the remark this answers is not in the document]
dmitrii-beliakov: One thought is that there are methods in this class called `createMessageKey` so maybe stick to this pattern? The methods are similar. 
Otherwise, what came to my mind is `name`. It is an index name, right? Maybe also replace create with get because `create unique index` is a well-known SQL command. To me, `getUniqueLlldIndexName` and `getUniqueArpIndexName` look better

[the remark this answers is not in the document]
Fedichk: In DB we have the property: `switch_connected_device.unique_index`. So, `UniqueIndex` is like the name of the field and the method builds the content of the field. Maybe `buildUniqueLlldIndex`, but it is again similar to the SQL command? @niksv , @pablomuri , @durgakako @IvanChupin @pkazlenka , @artHalMirantis What do you think?

src-java/connecteddevices-topology/connecteddevices-storm-topology/src/test/java/org/openkilda/wfm/topology/connecteddevices/service/PacketServiceTest.java
Fedichk: Fixed. Thanks!

src-java/connecteddevices-topology/connecteddevices-storm-topology/src/main/java/org/openkilda/wfm/topology/connecteddevices/service/PacketService.java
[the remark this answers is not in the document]
IvanChupin: For me getSomething() means that there is certainly no create operation.
createSomething() obviously create something, but I am not sure, should I expect that this method create and store and get, or just only create and get..
I would use generateSomeIndex(), at least for me it would be clear that we certainly generating some index without storing it. But usually for me, generate is somehow related to uuid format.. idk should we use generate with not uuid format, on the other hand, we use word generate to create some public key for ssh, so maybe this word could be used..:)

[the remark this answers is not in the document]
Fedichk: If there are no other opinions will update to `buildUniqueLlldIndex`, because from my point of view creating is too generic, and generating is like we make something from nothing but building, is like making something from smaller parts (exactly our case).
Thank you for your thoughts!

dmitrii-beliakov (approved):

src-java/connecteddevices-topology/connecteddevices-storm-topology/src/main/java/org/openkilda/wfm/topology/connecteddevices/service/PacketService.java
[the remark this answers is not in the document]
pablomuri: I agree with @niksv @IvanChupin. I prefer "generate" but "build" is also ok.

niksv (approved):

pkazlenka:
Verified using existing autotests on virtual and hw environment. No need to add new tests as nothing has changed from user perspective.

PR #5110, opened 2023-03-02
Bugfix/4574 switch validation fail for grpc timeout

IvanChupin:
Fix for SwitchManagerTopology:
set coordinator bolt to listen SpeakerWorkerBolt, now it should listen call back registration tasks and register this "subtasks" from SwitchValidationFsm. 
As result it fixes the problem with switch validation fail during grpc downtime.. 
Fixed error field, now error field for logical ports entity should be populated correctly in case of any errors from grpc(e.g. timeout)

Added unit test for LogicalPortsValidationEntryV2Test

### DIFF
--- a/src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/info/switches/v2/LogicalPortsValidationEntryV2.java
+++ b/src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/info/switches/v2/LogicalPortsValidationEntryV2.java
@@ -75,6 +75,7 @@ public class LogicalPortsValidationEntryV2 implements Serializable {
                     .excess(entry.getExcess())
                     .proper(entry.getProper())
                     .misconfigured(entry.getMisconfigured())
+                    .error(error)
                     .build());
         }
         return result;
--- a/src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/info/switches/v2/LogicalPortsValidationEntryV2.java
+++ b/src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/info/switches/v2/LogicalPortsValidationEntryV2.java
@@ -23,6 +23,7 @@ import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import lombok.Builder;
 import lombok.Data;
+import org.apache.commons.lang3.StringUtils;
 
 import java.io.Serializable;
 import java.util.ArrayList;
@@ -59,6 +60,8 @@ public class LogicalPortsValidationEntryV2 implements Serializable {
         builder.proper(joinLists(nonNullEntries.stream().map(LogicalPortsValidationEntryV2::getProper)));
         builder.missing(joinLists(nonNullEntries.stream().map(LogicalPortsValidationEntryV2::getMissing)));
         builder.misconfigured(joinLists(nonNullEntries.stream().map(LogicalPortsValidationEntryV2::getMisconfigured)));
+        nonNullEntries.stream().filter(e -> StringUtils.isNoneBlank(e.error)).findFirst()
+                .map(LogicalPortsValidationEntryV2::getError).ifPresent(builder::error);
         return builder.build();
     }
 
--- a/src-java/base-topology/base-messaging/src/test/java/org/openkilda/messaging/info/switches/v2/LogicalPortsValidationEntryV2Test.java
+++ b/src-java/base-topology/base-messaging/src/test/java/org/openkilda/messaging/info/switches/v2/LogicalPortsValidationEntryV2Test.java
@@ -43,6 +43,22 @@ public class LogicalPortsValidationEntryV2Test {
     }
 
     @Test
+    public void splitAndUniteEmptyLogicalPortEntryWithErrorTest() {
+        LogicalPortsValidationEntryV2 entry = LogicalPortsValidationEntryV2.builder()
+                .asExpected(false)
+                .error("Timeout for waiting response on DumpLogicalPortsRequest() Details: Error in SpeakerWorkerService")
+                .missing(new ArrayList<>())
+                .misconfigured(new ArrayList<>())
+                .excess(new ArrayList<>())
+                .proper(new ArrayList<>())
+                .build();
+        List<LogicalPortsValidationEntryV2> list = entry.split(4, 4);
+        assertEquals(1, list.size());
+        LogicalPortsValidationEntryV2 united = LogicalPortsValidationEntryV2.join(list);
+        assertEquals(entry, united);
+    }
+
+    @Test
     public void splitAndUniteNullLogicalPortEntryTest() {
         LogicalPortsValidationEntryV2 entry = LogicalPortsValidationEntryV2.builder()
                 .asExpected(true)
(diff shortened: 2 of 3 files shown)


src-java/base-topology/base-messaging/src/test/java/org/openkilda/messaging/info/switches/v2/LogicalPortsValidationEntryV2Test.java
```diff
@@ -42,6 +42,22 @@ public void splitAndUniteEmptyLogicalPortEntryTest() {
...
+                .misconfigured(new ArrayList<>())
+                .excess(new ArrayList<>())
+                .proper(new ArrayList<>())
+                .build();
+        List<LogicalPortsValidationEntryV2> list = entry.split(4, 4);
+        assertEquals(1, list.size());
+        LogicalPortsValidationEntryV2 united = LogicalPortsValidationEntryV2.join(list);
+        assertEquals(entry, united);
+    }
+
```
niksv, line 60: I think also you should test for a case when we have list size > 1.

IvanChupin: thanks, added appropriate test.

niksv (approved):

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/info/switches/v2/LogicalPortsValidationEntryV2.java
```diff
@@ -59,6 +60,8 @@ static LogicalPortsValidationEntryV2 join(List<LogicalPortsValidationEntryV2> en
         builder.proper(joinLists(nonNullEntries.stream().map(LogicalPortsValidationEntryV2::getProper)));
         builder.missing(joinLists(nonNullEntries.stream().map(LogicalPortsValidationEntryV2::getMissing)));
         builder.misconfigured(joinLists(nonNullEntries.stream().map(LogicalPortsValidationEntryV2::getMisconfigured)));
+        nonNullEntries.stream().filter(e -> StringUtils.isNoneBlank(e.error)).findFirst()
```
dmitrii-beliakov, line 63: Why `findFirst`? What if there more than one element in the steam?

src-java/base-topology/base-messaging/src/test/java/org/openkilda/messaging/info/switches/v2/LogicalPortsValidationEntryV2Test.java
```diff
@@ -147,6 +164,22 @@ public void splitAndUniteManyEntriesLogicalPortEntryTest() {
...
+        LogicalPortsValidationEntryV2 entry = LogicalPortsValidationEntryV2.builder()
+                .asExpected(true)
+                .error("Some error message")
+                .missing(buildLogicalPortsInfo(1, 500))
+                .excess(buildLogicalPortsInfo(1000, 600))
+                .proper(buildLogicalPortsInfo(2000, 700))
+                .misconfigured(buildMisconfiguredLogicalPortsInfo(3000, 800))
+                .build();
+        List<LogicalPortsValidationEntryV2> list = entry.split(100, 200);
+        assertEquals(14, list.size());
```
dmitrii-beliakov, line 178: Could add a message to the assertion explaining what `14` means?

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/info/switches/v2/LogicalPortsValidationEntryV2.java
IvanChupin: Thanks, I thought about it, but actually if there will be multiple elements, then they have to contain the same error status(all the same, in case of null, all should contain null), due to the way how this collection filled initially. Otherwise it should be error probably, because if this field will be different among multiple components it means that some problems during this collection creation. 
We have to think about this field that spread across the whole collection as a single value. (whether this api design mistake or not, I am not sure.)
Could you please suggest the best way to deal with this kind of situation?
.. Or we could leave it as it is?

src-java/base-topology/base-messaging/src/test/java/org/openkilda/messaging/info/switches/v2/LogicalPortsValidationEntryV2Test.java
IvanChupin: thank you, fixed

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/info/switches/v2/LogicalPortsValidationEntryV2.java
dmitrii-beliakov: Seems like there are a few assumptions (which might not be true) and that makes it more difficult to understand the code. The behavior doesn't have to be deterministic (execute the same steps twice, but receive different errors). So, I wouldn't leave it as it is. 
At least you can log the case: "Otherwise it should be error probably". Another option is to remove dupes from the stream of errors (for example, Stream::distinct) and then use Collectors::joining to keep all errors. If there is really only the duplicate errors, you get the same result as it is now, nothing to lose. But if, for whatever reason, it is not, you don't swallow errors here. To me that makes sense, I'd suggest this approach. The whole thing is a join method, so you join the errors as well. Yet another option is to change String to a collection of strings, but this is tricky because it might be a breaking change and probably is out of scope for this task.

IvanChupin: Thank you, fixed, please check.
@dmitrii-beliakov

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/info/switches/v2/LogicalPortsValidationEntryV2.java
```diff
@@ -60,8 +60,12 @@ static LogicalPortsValidationEntryV2 join(List<LogicalPortsValidationEntryV2> en
         builder.proper(joinLists(nonNullEntries.stream().map(LogicalPortsValidationEntryV2::getProper)));
         builder.missing(joinLists(nonNullEntries.stream().map(LogicalPortsValidationEntryV2::getMissing)));
         builder.misconfigured(joinLists(nonNullEntries.stream().map(LogicalPortsValidationEntryV2::getMisconfigured)));
-        nonNullEntries.stream().filter(e -> StringUtils.isNoneBlank(e.error)).findFirst()
-                .map(LogicalPortsValidationEntryV2::getError).ifPresent(builder::error);
+        if (nonNullEntries.stream().map(e -> e.error).distinct().count() > 1) {
+            builder.error(nonNullEntries.stream().map(e -> e.error)
+                    .distinct().collect(Collectors.joining(",")));
+        } else {
+            nonNullEntries.stream().map(e -> e.error).filter(Objects::nonNull).findFirst().ifPresent(builder::error);
```
dmitrii-beliakov, line 66: This if-else seems redundant now. joining on a stream of one element returns you simply that element. What you probably wanted to handle here is an empty stream after the filtering to not add an empty string as an error? As an option, you can simplify it by collecting 64-65 into a variable and pass it to the builder if it is not blank (you don't need to collect the stream multiple times). Or use a different reduce operation, if you prefer one-liners: 
```
nonNullEntries.stream().map(e -> e.error).filter(Objects::nonNull).reduce(String::concat).ifPresent(builder::error);
```

IvanChupin: Thank you, what I actually want to handle in the else statement is the case where only one element exist or none elements at all. 

 `joining on a stream of one element returns you simply that element.` ->
 I might be wrong, but I do not really understand where I am trying to join a stream of one element.. 


I understand the point in your suggested line of code
`nonNullEntries.stream().map(e -> e.error).filter(Objects::nonNull).reduce(String::concat).ifPresent(builder::error);`. 
But here is the trick: If you somehow receive null element and none-null element in the same stream, then according to the filter predicate you will just filter null elements which is not the behaviour I want to receive.  
 What I really want is to keep null element if there is another none-null element exist. And keep this "null" word in the error field at the end. Because this should mean that something wrong with the code that generated this elements..

dmitrii-beliakov (approved):

PR #5120, opened 2023-03-11
HA flows Part 7: Add H&S create operation

niksv:
API create stub was replaced with HaFlowCreateFSM.
This FSM allocates resources and pathes for ha-flows but doesn't install rules yet

Related To #5061


src-java/flowhs-topology/flowhs-storm-topology/src/test/java/org/openkilda/wfm/topology/flowhs/validation/HaFlowValidatorTest.java
```diff
@@ -0,0 +1,255 @@
...
+                        buildSubFlow(SUB_FLOW_ID_2, SWITCH_ID_3, PORT_3, VLAN_3, INNER_VLAN_3)))
+                .build();
+        haFlowValidator.validate(request);
+    }
+
+
+    @Test(expected = InvalidFlowException.class)
+    public void failIfMaxLatencyTier2HigherThanMaxLatencyTest()
+            throws InvalidFlowException, UnavailableFlowEndpointException {
+        HaFlowRequest request = buildHaFlowRequestWithMaxLatency((long) 1000, (long) 500);
```
dmitrii-beliakov, line 171: Why don't you use literals here as on 192? Same for 178

src-java/kilda-model/src/main/java/org/openkilda/model/HaFlow.java
```diff
@@ -255,6 +257,46 @@ private void validateReversePath(HaFlowPath haPath) {
...
+        } else if (pathId.equals(getProtectedReversePathId()) && getProtectedForwardPathId() != null) {
+            return Optional.of(getProtectedForwardPathId());
+        } else {
+            // Handling the case of non-active paths.
+            Optional<Long> requestedPathCookie = getPath(pathId)
+                    .map(HaFlowPath::getCookie)
+                    .map(FlowSegmentCookie::getFlowEffectiveId);
+            if (requestedPathCookie.isPresent()) {
+                return getPaths().stream()
+                        .filter(path -> !path.getHaPathId().equals(pathId))
```
dmitrii-beliakov, line 290: shouldn't you filter out the protected path as well?

src-java/kilda-pce/src/main/java/org/openkilda/pce/impl/InMemoryPathComputer.java
```diff
@@ -114,57 +122,144 @@ private GetPathsResult getPath(AvailableNetwork network, Flow flow,
...
+
+    private static int findFirstReverseSharedSegmentId(List<Segment> firstSegments, List<Segment> secondSegments) {
+        int i = firstSegments.size() - 1;
+        int j = secondSegments.size() - 1;
+
+        while (i >= 0 && j >= 0 && firstSegments.get(i).areEndpointsEqual(secondSegments.get(j))) {
+            i--;
+            j--;
+        }
+        return i + 1;
```
dmitrii-beliakov, line 226: If none segments are equal these methods still return a value and it's gonna be out of range. It's not what the method name suggests and you are encoding some special meaning in -1 and size(). I'd suggest to use some different approach, like Optional

src-java/kilda-pce/src/main/java/org/openkilda/pce/impl/InMemoryPathComputer.java
```diff
@@ -114,57 +122,144 @@ private GetPathsResult getPath(AvailableNetwork network, Flow flow,
...
+        haPath.isBackupPath(firstPath.isBackupPath() || secondPath.isBackupPath());
+        haPath.minAvailableBandwidth(Math.min(
+                firstPath.getMinAvailableBandwidth(), secondPath.getMinAvailableBandwidth()));
+        return haPath.build();
+    }
+
+    private static SwitchId findYPointForForwardPaths(Path firstPath, Path secondPath) {
+        int id = findLastForwardSharedSegmentId(firstPath.getSegments(), secondPath.getSegments());
+        if (id == -1) {
+            return firstPath.getSrcSwitchId();
```
dmitrii-beliakov, line 194: In case there are no equal end points, this method is still say that a Y-point exists.

dmitrii-beliakov (commented):
Great job! 👍
I reviewed 57/65. I'll give it another pass later.

src-java/kilda-model/src/main/java/org/openkilda/model/HaFlow.java
niksv: No. this method must return opposite protected path too. 
Main and protected path have different cookies.

src-java/kilda-pce/src/main/java/org/openkilda/pce/impl/InMemoryPathComputer.java
niksv: good point. Fixed

src-java/kilda-pce/src/main/java/org/openkilda/pce/impl/InMemoryPathComputer.java
niksv: fixed

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/validation/FlowValidator.java
```diff
@@ -93,6 +100,27 @@ public FlowValidator(PersistenceManager persistenceManager) {
         this.physicalPortRepository = persistenceManager.getRepositoryFactory().createPhysicalPortRepository();
     }
 
+    /**
+     * Validates the specified flow id.
+     *
+     * @param id - ID of a flow, a y-flow, a ha-flow or a ha-subFlow.
+     * @throws InvalidFlowException is thrown if a flow, a y-flow or ha-flow with specified ID exists.
+     */
+    public void validateFlowIdUniqueness(String id) throws InvalidFlowException {
```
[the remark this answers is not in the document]
dmitrii-beliakov: Yeah, probably it is a separate issue (maybe better to create one). I think if DB doesn't enforce that uniqueness between different indexes and the concurrent environment allows to create multiple flows at the same time, then when application scales the probability of this problem increases. So, it's better not to forget about it and fix it.
It is going to be difficult to diagnose and fix on a running system (could be some flaky test due to very fast env as well)

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/flow/resources/FlowResourcesManager.java
```diff
@@ -147,8 +172,58 @@ FlowResources allocateResources(Flow flow, PathId forwardPathId, PathId reverseP
...
+    HaFlowResources allocateResources(
+            HaFlow haFlow, PathId forwardPathId, PathId reversePathId, SwitchId yPointSwitchId)
+            throws ResourceAllocationException {
+        HaPathResources.HaPathResourcesBuilder forward = HaPathResources.builder()
+                .pathId(forwardPathId);
+        HaPathResources.HaPathResourcesBuilder reverse = HaPathResources.builder()
+                .pathId(reversePathId);
+        Map<String, PathId> reverseSubPathIds = new HashMap<>();
+        for (int i = 0; i < haFlow.getSubFlows().size(); i++) {
+            char suffix = (char) (HA_SUB_PATH_BASE_SUFFIX + i);
```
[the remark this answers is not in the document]
dmitrii-beliakov: yeah, I understand, but I was pointing at possible high value of i. Even though 30 subflows sound unrealistic, I thought it is better to throw an exception in this case rather than having the prefix `DEL` symbol

src-java/kilda-model/src/main/java/org/openkilda/model/HaFlow.java
dmitrii-beliakov: alright, shouldn't you then include that different cookie? 🙂
I mean in the section above it returns the opposite of main or the opposite of protected, but else handles only the main. The method doesn't need to return the opposite of a protected non-active path?

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/flow/resources/FlowResourcesManager.java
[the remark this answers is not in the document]
niksv: Added check

src-java/kilda-model/src/main/java/org/openkilda/model/HaFlow.java
niksv: Why do you think that `else` handles only the main?
Else section tries to find a path by cookie. main and protected paths have different cookies. So else section handles both of them

src-java/floodlightrouter-topology/floodlightrouter-storm-topology/replay_pid76393.log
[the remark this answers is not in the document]
niksv: yep. thank you. deleted :)

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/haflow/create/HaFlowCreateFsm.java
```diff
@@ -0,0 +1,432 @@
...
+                    errorMessage);
+            Message message = new ErrorMessage(error, getCommandContext().getCreateTime(),
+                    getCommandContext().getCorrelationId());
+            sendNorthboundResponse(message);
+        }
+
+        fireError(errorMessage);
+    }
+
+    private void resetState() {
```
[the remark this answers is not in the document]
dmitrii-beliakov: Right, so it's like there are two states, or groups of states. One kind of state is related to resources and another kind of state related to retries. I understand it as we have two fsms here. One handles retries and another all those states defined below. 
I mean reset state doesn't fully reset the _state_ since it remembers the number of retries. This reset state is more like "prepare for an Entry event". That was confusing for me. But changes could be even more confusing. Maybe add some comment here or at declaration that it is intended and they are not part of "state" in that sense of state, or that they are not supposed to be changed in reset state.

dmitrii-beliakov (approved):

artHalMirantis:
RP can be merged, autochecks are enough (including autotests introduced during development)

Issue #2072, opened 2019-02-21
Ability to add single switch flow into diverse group


IlyaMoiseev:
Now it is forbidden for technical reasons
Related #1231

PR #5129, opened 2023-03-17
#2072: Add single-switch flow into diversity group tests

pkazlenka:
* Re-enabled tests for the feature
* Added y-flow test
* Removed tests with two single-switch flows (moved to unit level)
* Grouped API calls to asynchronous pools where possible


src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/PathHelper.groovy
```diff
@@ -269,4 +273,24 @@ class PathHelper {
...
+                                                        List<List<PathNodePayload>> comparedFlows) {
+        def baseFlowSwitches = getInvolvedSwitchesV2(baseFlow)
+        def comparedFlowSwitches = comparedFlows.collect {getInvolvedSwitchesV2(it)}.flatten() as Set
+        def baseFlowIsls = baseFlow.size() > 1 ? getInvolvedIsls(convert(baseFlow)) : []
+        def comparedFlowIsls = comparedFlows.collect {getInvolvedIsls(convert(it))}.flatten() as Set
+        def intersectingSwitchSize = baseFlowSwitches.intersect(comparedFlowSwitches).size()
+        def intersectingIslSize = baseFlowIsls.intersect(comparedFlowIsls).size()
+        return new OverlappingSegmentsStats(intersectingIslSize,
+                intersectingSwitchSize,
+                intersectingIslSize ? intersectingIslSize / baseFlowIsls.size() * 100 as int : 0,
```
dmitrii-beliakov, line 293: maybe it is a controversial topic, but nevertheless I think this is a tautological test. It repeats the algorithm it is going to test. If there is a bug in the algorithm itself, this test cannot catch it. Changes are going to be symmetric to prod and test, but it doesn't say anything about correctness.
A better approach is "compare to what an expert says is correct", i.e. write down the expected values without this sort of calculations. Otherwise, there must be a test testing that this calculation is correct, and there should not be such a thing as a test for test.

pkazlenka: The approach you are describing is DDD - data driven development. When we exactly know what would be input. It works well on unit testing level, where you can mock (and thus control) almost everything. Also, if we change just one detail in env, e.g., add ISL to have necessary topology for other tests, you'll have to fix data manually for all the rest of the tests.
Another approach is to independently implement tested feature and compare results calculated by test implementation. This approach is less robust, but much more easier to maintain.

dmitrii-beliakov: But is it the case that you need these stats in many tests? 
There might be a test for some fixed topology where the expected result is known. I still think there should not be calculations like this. 
To me, if you really need it in the dynamic manner, then let's extract that logic into a separate place in PROD. Write unit tests with very-well known expected result and you will reuse the same PROD logic in tests as a service.

pkazlenka: Yes, at the moment there are 4 cases where we verify stats: three usual flows, one usual + 2 one-switch, three usual y-flows and y-flow + 2 one-switch y-flows.
Overall, 21 calls of this method x 4 elements each (i.e. 48 stats)

Now, let's imagine, that API v3 would have logic changes. Or we'll find a mistake in our current logic and decide to change the one. Or, even better, 2 new fields of OverlappingStats object will be introduced. With current approach, tester would have to just add several lines in this method, and won't touch the rest of the code.
With your approach, you'll need to sit and re-calculate 21*6 numbers (and I assure you, it's not so easy at it looks like, I tried). And each new test would multiply this number. 

And I just remind, that this object is quite simple to calculate and has just 4 fields. Imagine your approach with such complicated matters like path calculations.

Long story short: in ideal world where we would have 10 persons writing tests DDD would be more predictable and robust approach. But in real world, we need to use computers to help us calculate expected results.

dmitrii-beliakov: I am not convinced that it is a good approach. In my experience, tautological tests almost always have very little value, but require maintenance. Simply, we have a program X, we don't trust it as-is and want to verify its correctness with a test. A test is a program Y, we trust it as-is and use its output to verify the program X. But in reality programs X and Y are the same program. Why we don't apply the same logic to Y, how comes we trust it without a test? If we can somehow achieve trust in Y, then why can't we simply apply it to X?
In the example you provided with a modification, the tautological approach is not so useful as well. Suppose we want to change the formulae and divide not by `baseFlowIsls.size`, but by `foo.size`. Dev changes PROD to use foo.size, QA changes the test use foo.size. But how do we know that this is correct? Right! Somebody will need `to sit and re-calculate ...`. 
If after re-calculation we start trust the code Y, which verifies the correctness of the code X, then actually, this verification says that the code is correct the 2nd time, because X and Y is the same thing. We already know that X is correct before executing Y. Then, we can remove step with the code Y and save on maintenance and engineers' time. Right?
Even without the example with changes, right now, how the reviewers can see that the code producing the expected results is the same as the results you got when doing your calculations? So, I don't think we can get rid of manual calculations, or not using the "expert" approach.

Surely, there is some value in this test, such as the test is executed in multiple places making sure the value is propagated to all places and so on. But it could be by different means, which would be easier, with less code and less maintenance. For example, we can formulate the testing of this feature as a question "are you confident that stats are working correctly in these 10 places?". We can answer "let me see, we have a unit test that covers all required (possible) cases, so the service that calculating stats is ok, we have a unit test showing that if you put a value from a service to the response it arrives to a user without modification (immutable types, mappers unit tests). Therefore, if we see the value in the response, it went from the service without modification, we know that the service output is correct, hence we are confident that the value in the response is correct".
It is possible to achieve `10 persons writing tests` by splitting the job. For example QAs analyze which combination of tests proves or gives high confidence that something is working and their output could be a text with explanation and several tasks for devs to write certain unit tests proving specific things.

I mean, I don't want to argue here, do whatever you believe is appropriate to test and verify it. My input is that you might want to reconsider the approach for testing similar cases on QA Council (or whatever you call it 😀). I feel like the testing and verification could be achieved with less effort.

dmitrii-beliakov: My problem is that after executing a robust test in various environments I only know that for a specific feature the test implementation is the same as the prod implementation (and I knew it even before executing the test). What I still don't know is that this implementation is correct.

pkazlenka: Why do you say that the implementations are the same? Testing and prod codes are written independently, based on specifications given.

The fact that we use application classes in testing framework is another topic, and we are looking for the volunteer to refactor the tests to not import application classes in tests.

dmitrii-beliakov: Seems like are going in circles. How do you know that the testing code producing the expected result is correct?

dmitrii-beliakov: I disagree, this is not how auto tests work. At least not all of them, maybe what you described is just one approach (to me, not the best one).
A better approach is in unit tests make sure that service classes do their job correctly. Pick some number of cases and the rest goes by induction, inter- or extrapolation. Most of test cases goes here. Specifically, if the service calculates something differently for one- and multi-switch flows, then there are these different cases. In integration tests you test that different modules passes that resulting calculation correctly. That is to test that some mapper doesn't tamper with the resulting data. In E2E you pick just some of the cases to show that all pieces wired together. In exploratory testing you analyze priorities and detect what is missing or better be covered. 
In E2E tests you know beforehand what the desired behavior is; the test is to compare this expected behavior to the actual one. 
In the boundary-value analysis, we extract the boundary values from the specification and use them to build constraints, we do not rewrite the algorithm from the specification. In the example on wiki *they don't write another program to produce min and max values*, as well as they *don't write another program that adds numbers and compare the sum it produces to the original add method*. I don't see how you use boundary-value analysis here. Am I missing something?

Yeah, I asked and answered myself this and many other questions. Do you want me to write my answers here? I can, but it is not so different conceptually from the actual real case. I believe we don't need to implement another serializer or router firmware to make sure the original serializer and firmware are working as expected.
It reminds me the centuries old argument. How to make sure that a clock on a ship showing the correct time? Take another clock? This is proven to be an unreliable solution. 

Again, I don't say there is no value at all. It is a lot of effort to reach some little confidence. We spend time on implementing a feature two times and still don't have an evidence that this feature works correctly. Moreover, you spent some time on tedious calculations, but in the end that job is discarded (at least I don't see it in this PR), the result of those calculations is not in the code base and is not used by the test automation. 
I still believe that the approach you are using is not right. If you want for some reason to use some program to test other program, at least you need to write a test proving that the first program works correctly. Specifically here, you need a test that shows that this testing framework calculates stats correctly (probably the same thing applies to other places as well).

niksv: Thoughts:
1. From my point of view, the best way to make sure that everything works as it should is to have hardcoded expected results.
2. At the same time, sometimes it's really easier to write a method that will collect this expected result. But @dmitrii-beliakov  is right that we are not sure if it will work. 

I am not an expert in testing theory. In this particular case the method that calculates the expected result is quite simple. I understand that this sounds like a phrase from a joke but maybe we should "write tests for tests? If we write a unit test for this method, it will:

1. Take a small amount of time
2. Ensure us that program Y works correctly
3. Save us from having to change the expected result in 40 places.

pablomuri: I agree with @dmitrii-beliakov on this.

> A better approach is in unit tests make sure that service classes do their job correctly. Pick some number of cases and the rest goes by induction, inter- or extrapolation. Most of test cases goes here. Specifically, if the service calculates something differently for one- and multi-switch flows, then there are these different cases. In integration tests you test that different modules passes that resulting calculation correctly. That is to test that some mapper doesn't tamper with the resulting data. In E2E you pick just some of the cases to show that all pieces wired together. In exploratory testing you analyze priorities and detect what is missing or better be covered.

Of course, having all possible combinations tested in FT have value. But sometimes it can have a lot of work if we follow DDD as you have said. At the end I agree with the @niksv's conclusions.

pkazlenka: Okay guys, thank you for your opinions. There are always a bunch of approach to testing, so, depending on product and resources, the different ones can be accepted. Unfortunately:

1. We can not add unit tests for functional tests. We neither don't have resources to do that (we have hundreds of helper functions and covering them with unit tests is not possible with our resources neither in near nor in observable future). We don't have pipeline prepared for it and our devopses don't have resources to help us with that.
2. We can not move to DDD on such a huge and unpredictable environments as local or staging kildas. We don't have resources to rewrite and maintain the tests to take hardcoded switches/isls/paths in pre-tests and recalculate manually all the expected values. Also, there will be significant efforts to add mocked endpoints to kilda to eliminate unpredictable behavior.

Also, I remain you, that dramatic change of testing strategy is not a part of this particular PR. As you can see, expected values were calculated in these particular tests for years, and this didn't have an impact on final quality of the product. I didn't change approach in this PR, I've just collected the repeating code into human-readable (I hope) methods to simplify review, further modification and avoid misprints.

So, please, review this PR and approve/reject with particular comments. Global changes into our development approaches should be discussed somewhere else.

niksv (approved):

artHalMirantis (approved):

PR #5131, opened 2023-03-22
Added ability to turn off storm topologies run

niksv:
With new config options you will be able to skip running of a set of topologies. It can save env preparation time if you will turn of some optional topologies

This is restoring of PR #5093. That PR was reverted because image build job desn't support it yet


niksv:
Jenkins unable to build Kilda with new parameters so I'll abandon this PR

PR #5139, opened 2023-03-23
Add HA-flow related parallelism options into FlowHs topology

niksv:
This PR will help us to change parallelism of HA bolts. It is needed because be default they take parallelism from topology option and it is overkill, because these bolts is not highloaded. To save CPU and memory resources we should set low parallelism value for them.  

Related To #5061


dmitrii-beliakov (approved):
It's not clear from the related issue why do you need this change. Could you please add some short description to this PR on why we need this change?

niksv:
@dmitrii-beliakov done

pkazlenka:
Passed regression is enough to validate this change.

PR #5143, opened 2023-03-24
Decrease stats pull timeout in Stats topology test

niksv:
Pull timeout was decreased from 1 second to 100 ms.
It must be enough and it saves ~1m for running the test (3m43s vs 2m48s).


artHalMirantis (approved):

artHalMirantis:
no need to be tested more as all-in-one

dmitrii-beliakov (approved):

PR #5168, opened 2023-04-19
[TEST] Improve post-test cleanups

pkazlenka:
* Fixed several logical/syntax mistakes in tests which caused overall environment instability
* Moved 'resilience' test package to the end of test execution to minimize possible env damage.

### DIFF
--- a/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/yflows/YFlowCreateSpec.groovy
+++ b/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/yflows/YFlowCreateSpec.groovy
@@ -395,6 +395,11 @@ source: switchId="${flow.sharedEndpoint.switchId}" port=${flow.sharedEndpoint.po
 
         cleanup:
         yFlowResponse && !exc && yFlowHelper.deleteYFlow(yFlowResponse.YFlowId)
+        Wrappers.wait(WAIT_OFFSET) {
+            /*Sometimes test is too fast, so one of subflows stays in 'In Progress' at this stage.
+            Let's wait for it to be removed */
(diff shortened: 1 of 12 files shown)


niksv (approved):

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/yflows/YFlowCreateSpec.groovy
```diff
@@ -395,6 +395,11 @@ source: switchId="${flow.sharedEndpoint.switchId}" port=${flow.sharedEndpoint.po
 
         cleanup:
         yFlowResponse && !exc && yFlowHelper.deleteYFlow(yFlowResponse.YFlowId)
+        Wrappers.wait(WAIT_OFFSET) {
```
artHalMirantis, line 398: there is no need in this wrapper, cause yFlowHelper.deleteYFlow contains assertion on the flow history, which is filing the delete action (the flow history logs it once it verifies the subflows are removed)

pkazlenka: This is workaround for another problem. One of subflows hangs 'In Progress' for a while after test is over, and then it is also removed, but this can break the next test (e.g. 'consume' some bandwidth).

artHalMirantis: deleteYFlow cant return you true until the both subflows will be deleted (with particular status assigned), and there no need to make it once again

pkazlenka: deleteYFlow should be called in this test only if test is failed and exception was not raised on flow creation.

PR #5188, opened 2023-05-05
Set the python requests library version for lab

pkazlenka:


Fedichk (approved):

pablomuri:
Could you please explain in the commit-msg why we use this version. Explain the problem or at least reference it to keep traceability

pablomuri (changes requested):
Could you please explain in the commit-msg why we use this version. Explain the problem or at least reference it to keep traceability

pkazlenka:
> Could you please explain in the commit-msg why we use this version. Explain the problem or at least reference it to keep traceability

Done

niksv (approved):

PR #5194, opened 2023-05-08
HA flows Part 12: Add H&S partial update operation

niksv:
Part of #5061

### DIFF
--- a/src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/service/haflow/HaFlowUpdateService.java
+++ b/src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/service/haflow/HaFlowUpdateService.java
@@ -18,31 +18,68 @@ package org.openkilda.wfm.topology.flowhs.service.haflow;
 import static java.lang.String.format;
 
 import org.openkilda.floodlight.api.response.rulemanager.SpeakerCommandResponse;
+import org.openkilda.messaging.Message;
+import org.openkilda.messaging.command.haflow.HaFlowPartialUpdateRequest;
 import org.openkilda.messaging.command.haflow.HaFlowRequest;
+import org.openkilda.messaging.command.haflow.HaFlowRequest.Type;
+import org.openkilda.messaging.command.haflow.HaFlowResponse;
+import org.openkilda.messaging.command.haflow.HaSubFlowDto;
+import org.openkilda.messaging.command.haflow.HaSubFlowPartialUpdateDto;
+import org.openkilda.messaging.command.yflow.FlowPartialUpdateEndpoint;
 import org.openkilda.messaging.error.ErrorType;
+import org.openkilda.messaging.error.InvalidFlowException;
+import org.openkilda.messaging.info.InfoMessage;
+import org.openkilda.model.Flow;
+import org.openkilda.model.FlowEndpoint;
+import org.openkilda.model.HaFlow;
+import org.openkilda.model.HaSubFlow;
+import org.openkilda.model.PathComputationStrategy;
+import org.openkilda.model.SwitchId;
 import org.openkilda.pce.PathComputer;
 import org.openkilda.persistence.PersistenceManager;
+import org.openkilda.persistence.repositories.KildaFeatureTogglesRepository;
+import org.openkilda.persistence.tx.TransactionManager;
 import org.openkilda.rulemanager.RuleManager;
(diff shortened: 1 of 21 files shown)


artHalMirantis (approved):
from the testing prospective  - its ready to merge
the tests are enough to cover basic needs/cases

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/service/haflow/HaFlowUpdateService.java
```diff
@@ -96,6 +136,230 @@ public void handleUpdateRequest(
...
+                || isFieldUpdated(request.getEncapsulationType(), haFlow.getEncapsulationType())
+                || isUpdateRequiredByDiverseFlowIdField(request, haFlow);
+    }
+
+    private static boolean isUpdateRequiredByPathComputationStrategy(
+            HaFlowPartialUpdateRequest request, HaFlow haFlow) {
+        PathComputationStrategy strategy = Optional.ofNullable(request.getPathComputationStrategy())
+                .orElse(haFlow.getPathComputationStrategy());
+        boolean strategyIsLatencyBased = LATENCY_BASED_STRATEGIES.contains(strategy);
+        boolean changedMaxLatency = isFieldUpdated(request.getMaxLatency(), haFlow.getMaxLatency());
```
dmitrii-beliakov, line 293: minor, non-blocking: the value on 293 is only needed if 292 is true and 295 is false, but it is calculated every time even when we know that the result will be discarded. I'd suggest to split/extract it in separate methods and call them only when needed (for example, as an operand in `||`)

src-java/flowhs-topology/flowhs-storm-topology/src/test/java/org/openkilda/wfm/topology/flowhs/service/haflow/HaFlowUpdateServiceTest.java
```diff
@@ -0,0 +1,456 @@
...
+                            .endpoint(new FlowPartialUpdateEndpoint(SWITCH_ID_3, PORT_3, VLAN_3, INNER_VLAN_3))
+                            .description(DESCRIPTION_2)
+                            .build()))
+            .diverseFlowId(HA_FLOW_ID_2)
+            .build();
+
+
+    @Mock
+    private FlowGenericCarrier carrier;
+    HaFlowUpdateService service;
```
dmitrii-beliakov, line 94: could be private. I'd separate fields in different blocks: all mocks, all expected results, and so on.

src-java/flowhs-topology/flowhs-storm-topology/src/test/java/org/openkilda/wfm/topology/flowhs/service/haflow/HaFlowUpdateServiceTest.java
```diff
@@ -0,0 +1,456 @@
...
+
+    @Test
+    public void isFullUpdateRequiredHaPathId() {
+        assertFalse(service.isFullUpdateRequired(patchBuilder().haFlowId(HA_FLOW_ID_1).build(), HA_FLOW));
+        assertFalse(service.isFullUpdateRequired(patchBuilder().haFlowId(HA_FLOW_ID_2).build(), HA_FLOW));
+    }
+
+    @Test
+    public void isFullUpdateRequiredSharedSwitch() {
+        assertTrue(service.isFullUpdateRequired(patchBuilder().sharedEndpoint(
```
dmitrii-beliakov, line 116: the expected and actual are swapped here? Actually not super clear which is the expected. Variable names could really help here.

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/service/haflow/HaFlowUpdateService.java
niksv: good point

src-java/flowhs-topology/flowhs-storm-topology/src/test/java/org/openkilda/wfm/topology/flowhs/service/haflow/HaFlowUpdateServiceTest.java
niksv: will fix in next PR

Issue #5198, opened 2023-05-15
HaFlowDiversitySpec."Able to create diverse ha-flows" test is unstable


niksv:
PR #5173 added shuffle into test HaFlowDiversitySpec."Able to create diverse ha-flows"
After that test became unstable. Need to change the way of picking switch triplets in the test.

Shuffle picking was temporary removed to stabilize test #5199

PR #5199, opened 2023-05-15
HA-flow: improvements and refactoring

### LINKED ISSUE #5197: HA-flow validation can't validate HA-flow with protected path
**Steps to reproduce:**
1. Create HA-flow with protected path via API `POST /v2/ha-flows`
```json
{
  "ha_flow_id": "haflow1",
  "shared_endpoint": {
    "inner_vlan_id": 0,
    "port_number": 10,
    "switch_id": "2",
    "vlan_id": 12
  },
  "maximum_bandwidth": 1000,
  "allocate_protected_path": "true",
  "sub_flows": [
    {
      "endpoint": {
        "inner_vlan_id": 0,
        "port_number": 10,
        "switch_id": "9",
        "vlan_id": 123
      },
      "description": "some1"
    },
    {
      "endpoint": {
        "inner_vlan_id": 0,
        "port_number": 10,
        "switch_id": "7",
        "vlan_id": 123
      },
      "description": "some2"
    }
  ]
}
```
2. Validate HA-flow via API `POST /v2/ha-flows/{ha_flow_id}/validate`

**Expected result:**
Validation is ok

**Actual result:**
Validation Fail. Y point group contains 4 buckets, but only 2 expected.

niksv:
* make several classes base for several FSMs
* add workaround for bug #5198
* improved func tests


src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/HaFlowRuleManagerProcessingAction.java
```diff
@@ -86,4 +91,15 @@ protected HaFlow copyHaFlowWithPathIds(
         }
         return haFlow;
     }
+
+    protected Set<PathId> getOverlappingPathIds(HaFlow haFlow) {
```
dmitrii-beliakov, line 95: I don't get it. Let's say inside for loop I added path IDs [1,2,3,4], then I remove all that are present in subpathIds collection, which is say [1,2]. So I return [3,4], but why they are called overlapping? Seems the opposite: it gives non-overlapping, but only from one side.
I understand overlap as intersection of the sets, I'd expect to add to a set all elements of one collection that are at the same time present in the other using contains.

niksv: It think it is bat naming again. Let me explain. We have 2 types of ingress rules.
1. own rule of the flow - 1 rule for each primary path of each flow
2. shared rules - these rules are shared by several flows (when flows share one in port for example)

When we removing/adding rules for some flow to the switch we must not remove shared rules which are used by other flows and we should not re install shared rules which are used by other flows

This method finds all paths (except own haFlow paths and protected paths) which have save source endpoint as haFlow endpoints. This pathIds will be used in RuleManager to check are they use shared rules or not.

dmitrii-beliakov: thanks for the explanation. So, a better name with the get semantics could be: unshared, independent, solo, or something similar. Or it could be rewritten into something like excludeOverlappingPaths

dmitrii-beliakov (changes requested):
please change the name for the method for overlapping paths. Otherwise looks good! 👍

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/HaFlowRuleManagerProcessingAction.java
niksv: thank you for review! fixed

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/HaFlowRuleManagerProcessingAction.java
```diff
@@ -86,4 +91,19 @@ protected HaFlow copyHaFlowWithPathIds(
         }
         return haFlow;
     }
+
+    /**
+     * Finds path IDs of paths which can use same shared rules as paths of HA-flow use on endpoint switches.
+     * Excludes paths IDs of HA-fLow suf paths from the result.
```
dmitrii-beliakov, line 97: a typo: sub

src-java/rule-manager/rule-manager-implementation/src/main/java/org/openkilda/rulemanager/RuleManagerImpl.java
```diff
@@ -604,6 +601,28 @@ private List<SpeakerData> buildForwardHaRules(
         return rules;
     }
 
+    private boolean generateReverseIngressMeter(
```
IvanChupin, line 604: maybe consider a slightly different method name?
To make it more obvious that it is not actually generating reverseIngressMeter.
I understand that if you read this method with interrogative intonation then you will get the right meaning.. but who reads the methods in the interrogative intonation first?:)


```suggestion
    private boolean shouldGenerateReverseIngressMeter(
```

src-java/rule-manager/rule-manager-implementation/src/main/java/org/openkilda/rulemanager/RuleManagerImpl.java
```diff
@@ -612,19 +631,24 @@ private List<SpeakerData> buildReverseHaRules(
...
             // reverse ingress rule
             if (ingress) {
                 Set<FlowSideAdapter> overlappingAdapters = new HashSet<>();
                 if (filterOutUsedSharedRules) {
-                    overlappingAdapters.addAll(getOverlappingMultiTableIngressAdapters(haFlow, subPath, adapter));
+                    overlappingAdapters.addAll(getOverlappingMultiTableIngressAdapters(
+                            haFlow, subPath, subPathIds, adapter));
                 }
                 if (subPath.getSrcSwitchId().equals(haPath.getYPointSwitchId())) {
+                    boolean generateMeterCommand = generateReverseIngressMeter(
```
IvanChupin, line 645: If the previous comment regarding method name is counted, then imho it is better to change this variable as well. 

```suggestion
                    boolean shouldGenerateMeterCommand = generateReverseIngressMeter(
```

src-java/kilda-model/src/main/java/org/openkilda/model/HaFlowPath.java
```diff
@@ -178,6 +179,24 @@ public Set<SwitchId> getEndpointSwitchIds() {
...
     }
 
+    public Set<PathId> getSubPathIds() {
+        return getSubPaths().stream().map(FlowPath::getPathId).collect(Collectors.toSet());
+    }
+
+    /**
+     * Get all switches used by the path (including segment switches).
+     */
+    public Set<SwitchId> getAllInvolvedSwitches() {
```
IvanChupin, line 189: Also we could use stream like fashion:
   
```
   public Set<SwitchId> getAllInvolvedSwitches() {
        return getSubPaths().stream().flatMap(e -> e.getSegments().stream())
                .flatMap(e -> Stream.of(e.getSrcSwitchId(), e.getDestSwitchId()))
                .collect(Collectors.toSet());
    }
```

IvanChupin (approved):

src-java/rule-manager/rule-manager-implementation/src/main/java/org/openkilda/rulemanager/RuleManagerImpl.java
niksv: fixed

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/OnReceivedInstallResponseAction.java
```diff
@@ -13,32 +13,30 @@
...
 
 import lombok.extern.slf4j.Slf4j;
 
 import java.util.UUID;
 
 @Slf4j
-public class OnReceivedInstallResponseAction extends BaseReceivedResponseAction {
+public class OnReceivedInstallResponseAction<T extends HaFlowPathSwappingFsm<T, S, E, C, ?, ?>, S, E,
+        C extends SpeakerResponseContext> extends BaseReceivedResponseAction<T, S, E, C> {
     public static final String ACTION_NAME = "install";
```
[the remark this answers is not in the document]
niksv: fixed

src-java/rule-manager/rule-manager-implementation/src/main/java/org/openkilda/rulemanager/RuleManagerImpl.java
niksv: fixed

src-java/rule-manager/rule-manager-implementation/src/main/java/org/openkilda/rulemanager/RuleManagerImpl.java
dmitrii-beliakov: please don't use word should in this context.
Currently, the method name means that we hope that the Meter will be generated, but it might not happen. Better words could be: required, ought, expected.

PR #5205, opened 2023-05-19
Updated README.md - corrected steps for build process on Ubuntu

zdunektomasz:
Updated document regarding Ubuntu steps to build Open-Kilda.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,65 +51,77 @@ We do not recommend upgrading pip and install docker-compose using the methods d
 
 For running a virtual environment (that is a Docker instance with the Open vSwitch service) it is required to have Linux kernel 4.18+ for OVS meters support.
 The following commands will install necessary dependencies on Ubuntu 18.04:
-  Add Python's PPA repository
-  ```shell
-  sudo add-apt-repository -y ppa:deadsnakes/ppa
-  ```
-  Install required packages:
-  ```shell
-  sudo apt update && sudo apt-get install -y \
-    maven \
-    openjdk-8-jdk \
-    python \
-    python3.8 \
-    python3-pip \
-    virtualenv \
-    make \
-    tox \
-    rsync \
-    openvswitch-switch \
-    linux-generic-hwe-18.04
-  ```
-    Install required components for Python 3.6:
-    ```shell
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,66 +47,97 @@ The following packages are required for building OpenKilda controller:
...
+    python3.8 \
+    python3-pip \
+    virtualenv \
+    make \
+    tox \
+    rsync \
+    openvswitch-switch \
+    linux-generic-hwe-18.04
+  ```
+    Install required components for Python 3.6:
```
dmitrii-beliakov, line 73: Some problem with the formatting here. The text falls into the code block. And it renders it as a inline code instead of a code block. It might be because of indentation.

README.md
```diff
@@ -46,66 +47,97 @@ The following packages are required for building OpenKilda controller:
...
+    ```shell
+    sudo pip3 install setuptools-rust==1.1.2 setuptools==46.4.0 --upgrade
+    ```
+
+  To avoid version conflict you can install python3-pip with the official script. To do it, you need to download script & run script:
```
dmitrii-beliakov, line 78: I'd recommend to avoid using so direct approach: you **need**. It's better to assume that a reader in the future knows the area better. This guide needs to distinguish whether something is mandatory or not. `You need` feels like if you don't do it, you are wrong. But the user might not have any conflict in the first place. Or maybe tomorrow there will be a better way to resolve this problem. It's better to specify whether running the script is mandatory or it is just one of many ways to resolve the version conflict if they have it.
In these cases, ~~you need~~ you could simply use `could`: "if you have a python version conflict error, you could...". _Or you could use other alternatives._
Also, please use `and` instead of the ampersand.

dmitrii-beliakov (changes requested):
I can fix the formatting and style issues. If you want, you can assign this issue to me, I'll take it over.

dmitrii-beliakov (changes requested):
I can fix the formatting and style issues. If you want, you can assign this issue to me, I'll take it over.

zdunektomasz:
thanks for update, I will fix it.

niksv (approved):

dmitrii-beliakov (approved):

PR #5214, opened 2023-05-24
[TEST] Add initial mechanism to unify work with paths

pkazlenka:
* Introduced new classes to work with paths (lists of nodes) in the same way over all the tests (not used in tests yet)

### DIFF
--- a/src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/Path.groovy
+++ b/src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/Path.groovy
@@ -1,0 +1,45 @@
+package org.openkilda.functionaltests.helpers.model
+
+import groovy.transform.EqualsAndHashCode
+import org.openkilda.testing.model.topology.TopologyDefinition.Isl
+import org.openkilda.northbound.dto.v2.flows.FlowPathV2
+import org.openkilda.testing.model.topology.TopologyDefinition
+
+/* This class represent any kind of flow path and is intended to help compare/manipulate paths
+(presented as lists of nodes), received from different endpoints in the same manner.
+To add the new way to represent the path, just add the one into PathFactory.get() method
+Usage example:
+def nodesFromRerouteResponse = haFlowHelper.reroute(haFlowId).getSharedPath().getNodes()
+def nodesFromPathsResponse = haPathHelper.get(haFlowId).getSharedPath().getForward()
+assert pathFactory.get(nodesFromRerouteResponse) == pathFactory.get(nodesFromPathsResponse)
+
+This class has to replace *PathHelper in future
+ */
+@EqualsAndHashCode(excludes = "topologyDefinition")
+class Path {
+    List<FlowPathV2.PathNodeV2> nodes
+    TopologyDefinition topologyDefinition
+
+    Path(List<FlowPathV2.PathNodeV2> nodes, TopologyDefinition topologyDefinition) {
+        if (nodes.size() % 2 != 0) {
+            throw new IllegalArgumentException("Path should have even amount of nodes")
+        }
+        this.nodes = nodes
+        this.topologyDefinition = topologyDefinition
+    }
+
+    List<Isl> getInvolvedIsls() {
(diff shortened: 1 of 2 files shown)


niksv (approved):

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/PathFactory.groovy
```diff
@@ -0,0 +1,35 @@
...
+import org.openkilda.testing.model.topology.TopologyDefinition
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.stereotype.Component
+
+@Component
+class PathFactory {
+    @Autowired
+    TopologyDefinition topologyDefinition
+
+    Path get(List nodes) {
```
rtretyak, line 15: My personal preference would be to go strongly typed here and create multiple overloads of 'get' for better IDE and refactoring support. Duck typing does not simplify anything here. But current implementation is 100% fine

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/PathFactory.groovy
```diff
@@ -0,0 +1,35 @@
...
+        }
+        if (nodes.get(0) instanceof PathNode || nodes.get(0) instanceof FlowPathV2.PathNodeV2) {
+            return new Path(nodes.collect { new FlowPathV2.PathNodeV2(it.getSwitchId(), it.getPortNo(), 0) }, topologyDefinition)
+        }
+        if (nodes.get(0) instanceof PathNodePayload) {
+            def convertedNodes = nodes.collect {
+                [new FlowPathV2.PathNodeV2(it.switchId, it.inputPort == null ? 0 : it.inputPort, 0),
+                 new FlowPathV2.PathNodeV2(it.switchId, it.outputPort == null ? 0 : it.outputPort, 0)]
+            }
+                    .flatten()
```
rtretyak, line 27: can use `collectMany` instead of `collect` and don't do flatten. Not really an issue as well

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/Path.groovy
```diff
@@ -0,0 +1,42 @@
...
+        def isls = topologyDefinition.getIsls() + topologyDefinition.getIsls().collect { it.reversed }
+        nodes.collate(2).collect { List<FlowPathV2.PathNodeV2> pathNodes ->
+            isls.find {
+                it.srcSwitch.dpId == pathNodes[0].switchId &&
+                        it.srcPort == pathNodes[0].portNo &&
+                        it.dstSwitch.dpId == pathNodes[1].switchId &&
+                        it.dstPort == pathNodes[1].portNo
+            }
+        }
+    }
```
rtretyak, line 41: This may potentially have issues if we 'create' new isls (by re-plugging via a-switch for example) that did not originally exist in the topologyDefinition. But it's ok for initial implementation

rtretyak (approved):

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/PathFactory.groovy
pkazlenka: Having everything strongly typed is an obvious requirement. Unfortunately, Java has problems with overloading generics: JVM considers `void set(List<Integer>)` and `void set(List<String>` as the same signature and raises compliation error.
The workaround could be to use wrapper classes, e.g. List<PathNodePayload> -> PathNodePayloads, but in current code state this would require massive refactoring (e.g. replacing application's DTOs with test's DTOs).

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/Path.groovy
pkazlenka: Probably, it worth to add a warning for future.
There should be another method (or flag), which is 'heavy' (with requests to database) and rarely used. Opposite to this 'light' one, often used.

PR #5220, opened 2023-05-25
[TEST] Add simple traffexam for HA-Flows

artHalMirantis:
-bidirectional ha flow traffexam implemented
-base test spec HaFlowTraffSpec added
-changed signatures for the ha flow request builders to add traffgen ports as parameter


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/haflows/HaFlowPathSwapSpec.groovy
```diff
@@ -85,7 +84,11 @@ class HaFlowPathSwapSpec extends HealthCheckSpecification {
             }
         }
 
-        // TODO check traffic
+        and: "Traffic passes through HA-Flow"
+        if (swT.isHaTraffExamAvailable()) {
+            haFlowHelper.getTraffExam(createdHaFlow).run().hasTraffic()
```
rtretyak, line 89: Implicit assertions in spock are applied only on top-level expressions. Means that when you are inside 'if', you'll have to add an explicit 'assert'. Otherwise this line won't fail even if there's no traffic. Check this across other changes in this PR as well
```suggestion
            assert haFlowHelper.getTraffExam(createdHaFlow).run().hasTraffic()
```

rtretyak (changes requested):

pkazlenka (approved):

PR #5228, opened 2023-06-02
[TEST] Refactor errors verification

pkazlenka:
* All error verifications are refactored to use new API (AbstractExpectedError)
* Re-enabled several tests after bugs are fixed
* Fixed misprints


src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/error/AbstractExpectedError.groovy
```diff
@@ -0,0 +1,39 @@
+package org.openkilda.functionaltests.error
+
+import org.openkilda.messaging.error.MessageError
+import org.springframework.http.HttpStatus
+import org.springframework.web.client.HttpClientErrorException
+
+import java.util.regex.Pattern
+
+abstract class AbstractExpectedError extends HttpClientErrorException{
```
rtretyak, line 9: Not sure why it extends HttpClientErrorException 🤔  It's not really an exception, is it? I mean, this is meant to help with verifications, not to be thrown

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/extension/env/EnvCleanupExtension.groovy
```diff
@@ -110,7 +110,7 @@ abstract class EnvCleanupExtension extends AbstractGlobalExtension implements Sp
                             it.dstSwitch.dpId == link.destination.switchId && it.dstPort == link.destination.portNo
                 }
                 if (!isl) {
-                    throw new IslNotFoundException("Wasn't able to find isl: $link")
+                    throw new IslNotFoundExpectedError("Wasn't able to find isl: $link")
```
rtretyak, line 113: I don't think these 'expectedErrors' should be thrown. I also don't think IslNotFoundException should be renamed to IslNotFoundExpectedError. It's a pretty specific Exception that is meant to indicate a missing Isl, not an 'expected error'

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/error/AbstractExpectedError.groovy
```diff
@@ -0,0 +1,39 @@
...
+    Pattern descriptionPattern
+
+    AbstractExpectedError(HttpStatus statusCode, String message, Pattern messagePattern) {
+        super(statusCode)
+        this.statusCode = statusCode
+        this.message = message
+        this.descriptionPattern = messagePattern
+    }
+
+    boolean equals(other) {
```
rtretyak, line 21: Will work, but comes with some compromises as well:
- you a forced to use `expectedError == actualError` and can't do `actualError == expectedError` (which is not obvious at all)
- also this comes a little bit out of order with all verifications in the code base, where we usually put 'actual' on the lhs of equals expression.

As a quick fix to existing design, I would suggest instead of overriding the 'equals' operator just rename it to sth like `matches`/`conforms`/`satisfies` etc. This will eliminate ambiguity resulting from the uncommon usage of equals operator, since usually it is expected to be 'symmetric'

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/HealthCheckSpecification.groovy
```diff
@@ -63,9 +63,9 @@ class HealthCheckSpecification extends HealthCheckBaseSpecification {
                     assert links.findAll { it.state != IslChangeType.DISCOVERED }.empty
                 }
                 def topoLinks = topology.islsForActiveSwitches.collectMany { isl ->
-                    [islUtils.getIslInfo(links, isl).orElseThrow { new IslNotFoundException(isl.toString()) },
+                    [islUtils.getIslInfo(links, isl).orElseThrow { new IslNotFoundExpectedError(isl.toString()) },
```
rtretyak, line 66: same here

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/error/AbstractExpectedError.groovy
pkazlenka: It eliminates hundreds of IDE and build warnings about comparison of incompatible types. Doesn't look clear, but I consider this 'hack' as acceptable in this particular case.

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/extension/env/EnvCleanupExtension.groovy
pkazlenka: Seems like automatic refactoring mistake. Will fix.

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/error/AbstractExpectedError.groovy
pkazlenka: I can't imagine the cases when this standard approach to testing can become ambiguous, and '==' is just shorter than .methodName(). But you're, probably, right. If I refactor to use custom method, there won't be need to inherit from HTTPClientErrorException and would make method signature more strict.

rtretyak: > this standard approach to testing

- '==' not being symmetric is not standard 
- 'actual' being on the 'right' is also not standard (at least in this codebase)
- hack with `extends HttpClientErrorException` adds to confusion as well

But I understand what you tried to do, just that it adds more unique approach to asserting rather than making it simpler to a wider auditory. I'm not insisting on any changes though, just suggesting, it's your call. I don't think that this is critical in any way

rtretyak: Also, it's a good idea to add javadocs/comments to any hack/tricky solution

pkazlenka: Overriding 'equals()' is a standard approach. This exact implementation doesn't fully implement it. It lacks spring web client listener that would convert HttpClientErrorException into ExpectError before.

Issue #5141, opened 2023-03-23
Excess rules exist on switch after flow with vlan stats partial update (remove stats)


artHalMirantis:
1) create flow with stats using POST [/v2/flows]
```
{
  "flow_id": "flow1",
  "source": {
    "inner_vlan_id": 0,
    "port_number": 10,
    "switch_id": "00:00:00:00:00:00:00:01",
    "vlan_id": 0
  },
  "destination": {
    "inner_vlan_id": 0,
    "port_number": 0,
    "switch_id": "00:00:00:00:00:00:00:02",
    "vlan_id": 0
  },
  "maximum_bandwidth": 10000,
  "statistics": {
    "vlans": [
      1,2,3
    ]
  }
}
```

2)partially update flow with empty stats set using PATCH [/v2/flows/{flowid}]

```
{
  "source": {
    "vlan_id": 15
  },
  "destination": {
     "vlan_id": 17
  },
  "statistics": {
    "vlans": []
  }
}
```

3) validate switch 

actual result: excess rules are on dst and src switches
expected: no excess rules on switches

artHalMirantis:
@durgakako actual behaviour is reproducible for patch, but please be aware it was also bug for put (keep in mind while fixing)

PR #5236, opened 2023-06-12
#5141: [TEST] Excess rules when update flow with stats VLANs

pkazlenka:
* Re-enabled appropriate test and added check to verify bugfix


artHalMirantis (approved):
approved
but @pkazlenka do we really need to cover every single bug with functional test
this case is quite rare and i thought manual check is enough

dmitrii-beliakov (approved):

PR #5252, opened 2023-06-20
Add NoviFlow switches support into lockkeeper

niksv:


src-python/lock-keeper/cli/__init__.py
artHalMirantis, line 1: do we need it in PR?

artHalMirantis (approved):
not critical, but worth to remove empty file

src-python/lock-keeper/cli/__init__.py
niksv: Yes. each python package contains __init__.py file 
https://stackoverflow.com/questions/10863268/how-is-an-empty-init-py-file-correct

src-java/testing/test-library/src/main/java/org/openkilda/testing/service/lockkeeper/LockKeeperServiceImpl.java
```diff
@@ -59,6 +60,9 @@
 @Profile("hardware")
 public class LockKeeperServiceImpl implements LockKeeperService {
 
+    public static final String SWITCH_TYPE_PARAMETER = "switch_type";
+    public static final String NOVIFLOW_SWITCH_TYPE = "noviflow";
```
pkazlenka, line 64: As long as you don't use these constants separately, it's better to join them into one, e.g.
```
private static final HashMap<String, String> NOFIFLOW_SWITCH_PARAMETERS = ImmutableMap.of("switch_type",  "noviflow") 
```
(For some reason we have guava in test-library dependencies).

src-python/lock-keeper/test/test_novi_cli.py
```diff
@@ -0,0 +1,75 @@
+from unittest import TestCase
+
+from cli.novi_cli import NoviCli
+
+class TestStringMethods(TestCase):
```
pkazlenka, line 5: Very nice!

src-python/lock-keeper/lockkeeper.py
```diff
@@ -80,37 +86,37 @@ def parse_dump_flows(raw):
...
 
 
 @app.route('/flows', methods=['POST'])
 def post_flows_route():
     payload = request.get_json()
-    commands = ['ovs-ofctl add-flow br0 in_port={in_port},' \
-                'actions=output={out_port}'.format(**flow) for flow in payload]
-    execute_remote_commands(commands)
+    switch_cli = _get_switch_cli(request.args.get(SWITCH_TYPE_PARAMETER))
+    switch_cli.create_flows(payload)
```
pkazlenka, line 97: If you pass `request` instead of particular argument into `_get_switch_cli()` and avoid temporary variables, you'll be able to shorten the code to 
```
_get_switch_cli(request).create_flows(payload)
```

pkazlenka (approved):

PR #5256, opened 2023-06-22
#4925: [TEST] Multiple TSDB targets

pkazlenka:
* Added support for Victoria TSDB
* OpenTSDB is now LegacyTSDB
* LegacyTSDB support is not implemented yet
* New way to work and verify statistics (via model.stats.*FlowStats)
* Refactored all tests that used TSDB
* New test configuration keys: legacy.tsdb.enpoint, tsdb.endpoint, tsdb.metric.prefix, use.legacy.tsdb (useless until LegacyTsdb interface is implemented)


artHalMirantis (approved):

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/model/stats/Direction.groovy
```diff
@@ -0,0 +1,13 @@
+package org.openkilda.functionaltests.model.stats
+
+enum Direction {
+    FORWARD("forward"),
+    REVERSE("reverse"),
+    BOTH("*")
```
dmitrii-beliakov, line 6: probably not important, but I'm just curious why a start and not a word "both" as in the other enum values?

src-java/testing/test-library/src/main/java/org/openkilda/testing/service/tsdb/model/RawDataPoints.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.stream.IntStream;
+
+@Data
+public class RawDataPoints {
+    private RawMetric metric;
+    private ArrayList<Long> values;
+    private ArrayList<Long> timestamps;
+
+    /**
+     * Converts result received from JSON to StatsResult object. If several datapoints have the same timestamp,
```
dmitrii-beliakov, line 36: It's a bit strange to read about received from JSON, it's not here and this class cannot enforce it (the same in other javadocs).
```suggestion
     * Converts this RawDataPoints to StatsResult object. If several datapoints have the same timestamp,
```

src-java/testing/test-library/src/main/java/org/openkilda/testing/service/tsdb/model/StatsResponseDto.java
```diff
@@ -13,15 +13,33 @@
...
-public class EmptyStatsResult extends StatsResult {
-    public EmptyStatsResult() {
-        super();
-        tags = Collections.emptyMap();
-        aggregateTags = Collections.emptyList();
-        dps = Collections.emptyMap();
+import java.util.HashMap;
+import java.util.List;
+
+@Data
```
dmitrii-beliakov, line 23: Here and in other places fields have a default access modifier, but a private is more appropriate for a 
DTO. You can use `@FieldDefaults(makeFinal=true, level=AccessLevel.PRIVATE)` on a class level.
Or is there a reason not to do so?

dmitrii-beliakov (approved):
👍 
just some smaller suggestions.

dmitrii-beliakov:
I squashed the commits

dmitrii-beliakov (approved):

pkazlenka:
This PR is closed because #5335 contains all it's changes and has support for legacy OpenTSDB.

Issue #5097, opened 2023-02-23
Add "protected" to available paths api


pzakatov:
As a user, I can request the availability of protected paths so I'd be able to see how much bandwidth is available to me in case I need a protected service.

API: /v1/network/paths

PR #5263, opened 2023-06-30
Feature/show protected paths

dmitrii-beliakov:
This feature adds calculation of a protected path for each path returned by `/network/paths` API end point and, if the path exists, adds it to the payload under the `protected_path` property.
To enable this functionality a user must specify a request param `include_protected` = true. If this parameter is omitted or false, `network/paths/` behaves as if this feature doesn't exist (nothing new in the response payload).
When a protected path doesn't exist for some path, payload doesn't contain the `protected_path` property (maybe worth to change it and show and empty object to highlight that the logic has been executed).

Closes https://github.com/telstra/open-kilda/issues/5097

### DIFF
--- a/src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/PathMapper.java
+++ b/src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/PathMapper.java
@@ -39,7 +39,8 @@ public abstract class PathMapper {
      */
     public org.openkilda.messaging.info.network.Path map(org.openkilda.pce.Path path) {
         if (path == null || path.getSegments().isEmpty()) {
-            return new org.openkilda.messaging.info.network.Path(0L, Duration.ZERO, new ArrayList<>(), false);
+            return new org.openkilda.messaging.info.network.Path(0L, Duration.ZERO, new ArrayList<>(), false,
+                    null);
         }
 
         List<PathNodePayload> nodes = new ArrayList<>();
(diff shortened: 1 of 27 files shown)


src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/PathMapper.java
```diff
@@ -61,7 +62,14 @@ public org.openkilda.messaging.info.network.Path map(org.openkilda.pce.Path path
             nodes.add(new PathNodePayload(lastSegment.getDestSwitchId(), lastSegment.getDestPort(), null));
         }
 
+        if (path.getProtectedPath() != null && path.getProtectedPath().getProtectedPath() != null) {
```
pablomuri, line 65: Why do we have a protected path inside the protected path?

src-java/kilda-pce/src/test/java/org/openkilda/pce/mapper/PathSegmentMapperTest.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.List;
+import java.util.Set;
+
+public class PathSegmentMapperTest {
+
+    @Test
+    public void toPathSegmentTest() {
+        Segment segment1 = Segment.builder()
+                .latency(12L)
+                .srcSwitchId(new SwitchId("00:01"))
```
pablomuri, line 39: [minor] this switchId could be also initialized with an int new SwitchId(1)
```suggestion
                .srcSwitchId(new SwitchId(1)))
```

In any case, I would create the 3 switchIds previously

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/PathMapper.java
dmitrii-beliakov: a protected path is actually the same type as a path. So a Path type contains a Path type variable; there are checks that don't allow this kind of recursion

src-java/kilda-pce/src/main/java/org/openkilda/pce/PathComputer.java
```diff
@@ -80,6 +80,15 @@ default GetHaPathsResult getHaPath(HaFlow haFlow, boolean isProtected)
     GetHaPathsResult getHaPath(HaFlow haFlow, Collection<PathId> reuseSubPathsResources, boolean isProtected)
             throws UnroutableFlowException, RecoverableException;
 
+    /**
+     * Calculates a protected path based on given parameters and configuration.
```
IvanChupin, line 84: should we use 'the' here?
```suggestion
     * Calculates a protected path based on the given parameters and configuration.
```

IvanChupin (commented):
Great job! Thanks,
Just a couple of notes from my side.

IvanChupin (approved):

dmitrii-beliakov:
I fixed the docs and squashed the commits again.

Issue #5313, opened 2023-07-27
Remove single table mode support


niksv:
Single table mode was deprecated and must be removed

PR #5319, opened 2023-07-28
Remove multi_table fields from DB objects

niksv:
Related to #5313


src-java/floodlight-service/floodlight-api/src/main/java/org/openkilda/floodlight/model/FlowSegmentMetadata.java
```diff
@@ -38,10 +38,9 @@ public class FlowSegmentMetadata implements Serializable {
     @JsonCreator
     public FlowSegmentMetadata(
             @JsonProperty("flowid") @NonNull String flowId,
-            @JsonProperty("cookie") @NonNull Cookie cookie,
-            @JsonProperty("multi_table") boolean multiTable) {
+            @JsonProperty("cookie") @NonNull Cookie cookie) {
         this.flowId = flowId;
         this.cookie = cookie;
-        this.multiTable = multiTable;
+        this.multiTable = true;
```
dmitrii-beliakov, line 44: Is it really needed? It can only have one value, right?

dmitrii-beliakov (approved):

src-java/floodlight-service/floodlight-api/src/main/java/org/openkilda/floodlight/model/FlowSegmentMetadata.java
niksv: yes, it will have only one value. I will have 2 more PRs (e.g #5382) which will remove fields from Java objects

PR #5326, opened 2023-08-01
Split connected devices messages

niksv:
Some messages can be too large to be pushed in Kafka so we have to split them

### DIFF
--- a/src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/split/SplitIterator.java
+++ b/src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/split/SplitIterator.java
@@ -1,0 +1,100 @@
+/* Copyright 2023 Telstra Open Source
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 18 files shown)


artHalMirantis:
automatic regression run is enough to cover test needs

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/split/SplitIterator.java
```diff
@@ -0,0 +1,100 @@
...
+
+        if (remainingChunkSize <= 0) {
+            remainingChunkSize = chunkSize;
+            addCurrent = false;
+            T result = buildFunction.apply(currentBuilder);
+            currentBuilder = defaultBuilderSupplier.get();
+            return Optional.of(result);
+        }
+        return Optional.empty();
+    }
```
pablomuri, line 99: Why didn't you implement the iterator interface?
https://www.baeldung.com/java-creating-custom-iterator

pablomuri (changes requested):

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/split/SplitIterator.java
niksv: Because this interface differs from Iterator interface. 
Itearator interface has methods `hasNext()` and `remove()` I dont need them. 
Also my interface can combine several iterms into one chunk. And `next` will return non empty value only when current chunk will be full. 

But iterator interface returns next element every call until the end of a collection. I don't need it.

PR #5333, opened 2023-08-07
Split GetFlowsPerPortForSwitchResponse because they can be too big for Kafka

niksv:


niksv:
automatic regression run is enough to cover test needs
Like in similar PR https://github.com/telstra/open-kilda/pull/5326#issuecomment-1663791267

IvanChupin (approved):

PR #5334, opened 2023-08-08
Fix/restore old otsdb config

niksv:


niksv:
do not need it anymore

PR #5340, opened 2023-08-18
Add ability to save history on flow PATCH operation.

### LINKED ISSUE #5237: Add support of HA-flow history to HA-flow patch
HA-flow patch has two options: lightweight - when there is no need to send new commands to switches (i.e. update the description) and via the update when patching one field it grabs the remaining field from the current HA-flow and performs the update. With lightweight patch we don't save the history; with the second one the history is saved because the normal update has its saving history functionality (it saves dumps before and after and other information such as resources allocation).
- [ ] add support of saving history when performing the lightweight patch
- [ ] review if it is needed to add some additional action to full-featured patch (for example, save the patch request itself or affected fields)

### LINKED ISSUE #5385: HA flow history on PATCH
Add saving a new event together with a dump on HA flow PATCH operation.

dmitrii-beliakov:
This PR moves history service classes to another module, so that they could be used in other modules. Then it adds support of saving history to flow PATCH when doing a partial update or invoking a full update via PATCH.
A new bolt and storm stream have been created to pass messages to history bolt that saves history events.

This PR also adds an ability to save history on HA-flow PATCH. It adds an event for the partial update, adds dumps before and after the partial update, adds a partial update request to the details of one of the actions. When the full update is required, the appropriate action is saved and the rest of the actions will be saved via the full update.

closes #5237 
closes #5385


src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/services/FlowOperationsService.java
```diff
@@ -429,6 +439,67 @@ UpdateFlowResult.UpdateFlowResultBuilder prepareFlowUpdateResult(FlowPatch flowP
...
+                .flowEventData(FlowEventData.builder()
+                        .event(FlowEventData.Event.PATCH)
+                        .flowId(flow.getFlowId())
+                        .details("Flow PATCH operation is invoked")
+                        .initiator(FlowEventData.Initiator.NB)
+                        .time(Instant.now())
+                        .build())
+                .flowHistoryData(FlowHistoryData.builder()
+                        .flowId(flow.getFlowId())
+                        .action("Flow PATCH parameters has been validated successfully.")
```
dmitrii-beliakov, line 496: typo

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/services/FlowOperationsService.java
```diff
@@ -392,13 +400,15 @@ public Flow updateFlow(FlowOperationsCarrier carrier, FlowPatch flowPatch)
 
         Flow updatedFlow = updateFlowResult.getUpdatedFlow();
         if (updateFlowResult.isNeedUpdateFlow()) {
+            saveHistoryActionFullUpdate(carrier, updatedFlow, flowPatch);
```
niksv, line 403: will we create 2 events in history? one here and one in flowhs topology?

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/model/FlowPatch.java
```diff
@@ -48,4 +48,6 @@ public class FlowPatch {
     private String description;
     private FlowEncapsulationType encapsulationType;
     private Set<Integer> vlanStatistics;
+
+    private String correlationId;
```
niksv, line 52: I'm not sure that it should be here. 
FlowPatch is just a model which is used here and in NB. In NB we are setting correclation id to the parent message
```
        FlowPatch flowPatch = flowMapper.toFlowPatch(flowPatchDto);
        flowPatch.setFlowId(flowId);
        CommandMessage request = new CommandMessage(new FlowPatchRequest(flowPatch), System.currentTimeMillis(),
                correlationId);
```
In this case FlowPatch.correlationId will be empty or it will duplicate information from CommandMessage.

You need correlationId only in one servce in nbWorker. Maybe it is worth to add `correlationId` argument into `saveHistoryAction*` methods?

niksv (commented):
In general topology looks correct

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/model/FlowPatch.java
dmitrii-beliakov: well, I was hesitating about putting it there and I wanted to know other's people opinion on that. It could be a parameter just next to flowPatch, or somehow commandContext could be injected into the operation service class.
Yes, probably it is better to have the correlation ID outside.

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/services/FlowOperationsService.java
dmitrii-beliakov: yes. And I think it make sense because this is actually what happens. The patch event would contain in this case that it's been invoked and then it passed the control to the update. The update history will follow.

dmitrii-beliakov:
I tested it manually. For lightweight update I used just priority:
```json
{
  "priority": 1
}
```
For full update, I created a flow with the `MAX_LATENCY` strategy and changed `max-latency`:
```json
{
  "priority": 8,
  "max-latency": 1112
}
```
In both cases the history is saved. There is a special message saying whether it was full or partial update.

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/model/FlowPatch.java
dmitrii-beliakov: I removed the variable

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/service/common/HistoryUpdateCarrier.java
dmitrii-beliakov, line 1: that is the coolest change 😎

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/bolts/HistoryOperationsBolt.java
```diff
@@ -58,7 +58,7 @@ public HistoryOperationsBolt(PersistenceManager persistenceManager) {
     public void init() {
         super.init();
 
-        historyService = new HistoryService(persistenceManager);
+        historyWriterService = new HistoryWriterService(persistenceManager);
```
dmitrii-beliakov, line 61: hmm, actually it's not the best name. I just wanted to have different names for the service that is invoked by a client and a service that does persistence job on the other side of the messaging. I'll try to come up with 
a better naming.

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/service/history/FlowHistory.java
```diff
@@ -0,0 +1,86 @@
...
+    private String action;
+    private String description;
+    private final String taskId;
+    private FlowDumpData flowDumpData;
+
+    private FlowHistory(String taskId) {
+        this.taskId = taskId;
+    }
+
+    public static FlowHistory withTaskId(String taskId) {
```
pkazlenka, line 41: IIRC, java conventions suggest using `of` word for static constructors. So I'd expect the one is named `ofTaskId()`.
Also, this would distinguish constructor from the rest pseudo-builder methods (`withX`).
Why won't you use @Builder annotation, by the way?

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/service/history/FlowHistory.java
```diff
@@ -0,0 +1,86 @@
...
+        this.flowDumpData = flowDumpData;
+        return this;
+    }
+
+    /**
+     * This method is a combination of {@link  #withFlowDump} and {@link #withFlowId}.
+     * @param flow save state of this HA-flow to history
+     * @return this object to chain other methods
+     */
+    public FlowHistory withFlowDumpBefore(Flow flow, FlowPath forward, FlowPath reverse) {
```
pkazlenka, line 70: From what I see in usage of this and the next methods, you, probably, can cut input parameters list to `Flow` only (anyway you extract paths from it before calling it in code):
```
    public FlowHistory withFlowDumpBefore(Flow flow) {
        this.flowDumpData = HistoryMapper.INSTANCE.map(flow, flow.getForwardPath(), flow.getReversePath(), STATE_BEFORE);
        this.flowId = flow.getFlowId();
        return this;
```

pkazlenka (approved):
Mostly good. Please, address the comments.

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/service/history/FlowHistory.java
dmitrii-beliakov: `with` makes the usage fluent, the whole thing looks like a sentence, like: save flow history with action and with dump. This one is made in a form of a static constructor so that the client can't skip adding the task ID and I actually made it the same as other method names on purpose. (in IDE it won't confuse, there is only one option after typing FlowHistory)
Builder annotation does the same, but without fluency and add builder() and build() methods. of() is also possible, probably it's just a personal preference. People make names different sometimes such as `emptyList` in java collecitons

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/service/history/FlowHistory.java
dmitrii-beliakov: agree that it needs to be removed, but other places is not ready for this change. When saving the history with dumping protected paths, we use different paths when using this method. 
There is a feature I created about changing the history structure. Once it is done, this method will no longer be needed and we can remove it.

src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/service/history/FlowHistory.java
dmitrii-beliakov: I made it of() kind of make sense as well, `save(FlowHistory.of(correlationId).withAction(...`
I left out `taskId` because there is correlation ID passed in in all cases. I don't know why it was called taskId in history, in other places it is called correlation ID

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/haflow/delete/actions/OnFinishedWithErrorAction.java
```diff
@@ -39,8 +39,8 @@ public OnFinishedWithErrorAction(FlowOperationsDashboardLogger dashboardLogger)
     public void perform(State from, State to, Event event, HaFlowDeleteContext context, HaFlowDeleteFsm stateMachine) {
         dashboardLogger.onFailedHaFlowDelete(stateMachine.getHaFlowId(), stateMachine.getErrorReason());
 
-        HaFlowHistoryService.using(stateMachine.getCarrier()).saveError(HaFlowHistory
-                .withTaskId(stateMachine.getHaFlowId())
+        FlowHistoryService.using(stateMachine.getCarrier()).saveError(HaFlowHistory
+                .of(stateMachine.getHaFlowId())
```
dmitrii-beliakov, line 43: oh, it is a bug

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/haflow/delete/actions/RevertHaFlowStatusAction.java
```diff
@@ -45,8 +45,8 @@ protected void perform(
             log.debug("Reverting the HA-flow status of {} to {}", haFlowId, originalStatus);
 
             haFlowRepository.updateStatus(haFlowId, originalStatus);
-            HaFlowHistoryService.using(stateMachine.getCarrier()).saveError(HaFlowHistory
-                    .withTaskId(stateMachine.getHaFlowId())
+            FlowHistoryService.using(stateMachine.getCarrier()).saveError(HaFlowHistory
+                    .of(stateMachine.getHaFlowId())
```
dmitrii-beliakov, line 49: oh, it is a bug

dmitrii-beliakov: I fixed it. Type safe correlation ID would really help here

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/BaseHaFlowPathRemovalAction.java
```diff
@@ -56,7 +56,7 @@ protected void removeFlowPaths(HaPathIdsPair haPathIdsPair,
                 if (oldSubPath != null) {
                     log.debug("Removed HA-flow sub path {}", oldSubPath);
                     updateIslsForFlowPath(oldSubPath);
-                    HaFlowHistoryService.using(carrier).save(HaFlowHistory.withTaskId(correlationId)
+                    FlowHistoryService.using(carrier).save(HaFlowHistory.of(correlationId)
```
IvanChupin, line 59: May we not create a new FlowHistoryService object for every iteration here?

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/BaseHaFlowPathRemovalAction.java
```diff
@@ -69,7 +69,7 @@ protected void removeFlowPaths(HaPathIdsPair haPathIdsPair,
                 HaFlowPath oldPath = haFlowPathRepository.remove(haFlowPathId).orElse(null);
                 if (oldPath != null) {
                     log.debug("Removed HA-flow path {}", oldPath);
-                    HaFlowHistoryService.using(carrier).save(HaFlowHistory.withTaskId(correlationId)
+                    FlowHistoryService.using(carrier).save(HaFlowHistory.of(correlationId)
```
IvanChupin, line 72: Here as well, and moreover, If I got it right we have FlowHistoryService.getNextHistoryEntryTime() method for these  cases.

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/BaseHaFlowPathRemovalAction.java
```diff
@@ -85,7 +85,7 @@ protected void removeRejectedPaths(Collection<PathId> subPathIds, Collection<Pat
             flowPathRepository.remove(subPathId)
                     .ifPresent(subPath -> {
                         updateIslsForFlowPath(subPath);
-                        HaFlowHistoryService.using(carrier).save(HaFlowHistory.withTaskId(correlationId)
+                        FlowHistoryService.using(carrier).save(HaFlowHistory.of(correlationId)
```
IvanChupin, line 88: The same here

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/BaseHaFlowPathRemovalAction.java
```diff
@@ -95,7 +95,7 @@ protected void removeRejectedPaths(Collection<PathId> subPathIds, Collection<Pat
         for (PathId haFlowPathId : haFlowPathIds) {
             haFlowPathRepository.remove(haFlowPathId)
                     .ifPresent(haFlowPath -> {
-                        HaFlowHistoryService.using(carrier).save(HaFlowHistory.withTaskId(correlationId)
+                        FlowHistoryService.using(carrier).save(HaFlowHistory.of(correlationId)
```
IvanChupin, line 98: Same here

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/BaseReceivedResponseAction.java
```diff
@@ -91,8 +91,8 @@ protected void handleResponse(
 
 
                 response.getFailedCommandIds().forEach((uuid, message) ->
-                        HaFlowHistoryService.using(stateMachine.getCarrier()).saveError(HaFlowHistory
-                                .withTaskId(stateMachine.getCommandContext().getCorrelationId())
+                        FlowHistoryService.using(stateMachine.getCarrier()).saveError(HaFlowHistory
```
IvanChupin, line 94: May we not create a new instance of FlowHistoryService each iteration?

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/DeallocateResourcesAction.java
```diff
@@ -44,8 +44,8 @@ public void perform(S from, S to, E event, C context, T stateMachine) {
         stateMachine.getOldResources().forEach(haFlowResources -> {
             transactionManager.doInTransaction(() ->
                     resourcesManager.deallocateHaFlowResources(haFlowResources));
-            HaFlowHistoryService.using(stateMachine.getCarrier()).save(HaFlowHistory
-                    .withTaskId(stateMachine.getCommandContext().getCorrelationId())
+            FlowHistoryService.using(stateMachine.getCarrier()).save(HaFlowHistory
```
IvanChupin, line 47: May we not create a new instance of FlowHistoryService each iteration?

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/HandleNotDeallocatedResourcesAction.java
```diff
@@ -31,8 +31,8 @@ public class HandleNotDeallocatedResourcesAction<T extends HaFlowPathSwappingFsm
     @Override
     public void perform(S from, S to, E event, C context, T stateMachine) {
         stateMachine.getOldResources().forEach(haFlowResources ->
-                HaFlowHistoryService.using(stateMachine.getCarrier()).saveError(HaFlowHistory
-                        .withTaskId(stateMachine.getCommandContext().getCorrelationId())
+                FlowHistoryService.using(stateMachine.getCarrier()).saveError(HaFlowHistory
```
IvanChupin, line 34: May we not create a new instance of FlowHistoryService each iteration?

IvanChupin (approved):
Great job! 
Just a small comment regarding FlowHistoryService:
You could consider to not create the FlowHistoryService every time you are going to use it. 
Apparently, creating a new service object on-request is a good practice with it's pros, but there are also cons exist(Performance, Resource Consumption, Creation Costs and so on). So do we actually need to create a new service object, particularly  in a iterations.. 

And also, you have the following method in FlowHistoryService:
 ```
  private Instant getNextHistoryEntryTime() {
        Instant now = Instant.now();
        if (lastHistoryEntryTime == null || lastHistoryEntryTime.isBefore(now)) {
            lastHistoryEntryTime = now;
        } else {
            lastHistoryEntryTime = lastHistoryEntryTime.plusMillis(1);
        }
        return lastHistoryEntryTime;
    }
```

And for the current implementation the this.lastHistoryEntryTime field for every getNextHistoryEntryTime() call should be null initially.

File: src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/BaseHaFlowPathRemovalAction.java (hunk 10 above)
Comment (dmitrii-beliakov) [reply src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/BaseHaFlowPathRemovalAction.java:72]: there is a TODO about it. Manipulating the time value is a hack and that method shouldn't exist in the first place. The effect is probably could be seen only in test env, or rare in prod.

File: src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/BaseHaFlowPathRemovalAction.java (hunk 9 above)
Comment (dmitrii-beliakov) [reply src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/BaseHaFlowPathRemovalAction.java:59]: Small Java objects are cheap, there are no long-running or heavy jobs done in the constructor, so it is not a big deal. There is a bigger task of unifying the carrier, once it is done, there will be no need for the using() static constructor, and it'll be easier to refactor further. 
But it is a good point! I would rather delegate it DI container so that app logic doesn't care about the scopes and lifecycle  of the service class.

**Comment by yuliiamir** [comment]
There are two 'stateBefore' blocks in the dump section during the partial update with the consecutive update(event: "Flow partial updating",). Need to recheck logic to avoid duplication of the data.

File: src-java/base-topology/base-storm-topology/src/main/java/org/openkilda/wfm/share/mappers/HaFlowHistoryMapper.java [hunk 16]
```diff
@@ -80,6 +80,7 @@ public abstract class HaFlowHistoryMapper {
     public abstract HaFlowHistoryEntry createHaFlowHistoryEntry(
             HaFlowEvent flowEvent, List<HaFlowHistoryPayload> payloads, List<HaFlowDumpPayload> dumps);
 
+    @Mapping(source = "taskId", target = "taskId")
     public abstract HaFlowEventDumpDataImpl createHaFlowEventDump(HaFlowDumpData haFlowDumpData);
```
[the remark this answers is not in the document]
dmitrii-beliakov: No, some renaming leftover. I'll fix it.

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/BaseHaFlowPathRemovalAction.java
dmitrii-beliakov: I created a local variable for the flow history service.

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/BaseHaFlowPathRemovalAction.java
dmitrii-beliakov: I created a local variable for the flow history service.

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/BaseHaFlowPathRemovalAction.java
dmitrii-beliakov: I created a local variable for the flow history service.

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/BaseHaFlowPathRemovalAction.java
dmitrii-beliakov: I created a local variable for the flow history service.

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/DeallocateResourcesAction.java
dmitrii-beliakov: I created a local variable for the flow history service.

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/common/actions/haflow/HandleNotDeallocatedResourcesAction.java
dmitrii-beliakov: I created a local variable for the flow history service.

dmitrii-beliakov:
@yuliiamir I double checked why we have two state before and it seems to me not a bug, but it is confusing definitely.
The first state before is saved when we receive a PATCH request, we save there a flow status before doing any operations. Flow patch implemented in a way so that it modifies the flow and only then sends the request to the full update. The update is not aware if it is requested from patch or not and it does its own dump before.
The dumps are actually different. I used the following request:
```json
{
  "priority": 8,
  "max-latency": 1112
}
```
that creates two dumps before, one of them have priority = 1 (it is saved before all operations), the second dump before has priority = 8 (modified by a patch, but before the update is processed). Given different request, it is possible that two dumps are the same.
So, I think this is still useful information given how the operations are implemented. We can change the patch logic, so that it modifies the flow only if it doesn't require the full update. But it seems to me a different issue. It is not just saving history during the patch, but how the patch itself works.
I created a proposal about changing the history structure, so it is more clear which action created a dump. Given that implemented, we could see that the dump is not actually a duplicate, but just performed in different places.

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/service/haflow/HaFlowUpdateService.java
```diff
@@ -162,9 +171,28 @@ public void handlePartialUpdateRequest(
             }
             carrier.sendNorthboundResponse(buildResponseMessage(
                     updateHaFlowResult.partiallyUpdatedHaFlow, commandContext));
+            FlowHistoryService.using(carrier).save(HaFlowHistory
```
dmitrii-beliakov, line 174: @niksv @yuliiamir 
this is an alternative way of how to save history on patch. Here we just wait until the decision of making the full update is made and the data is written to DB and only then save the "dump before". With this approach we don't save an extra dump before in case of the full update (the update makes its own dump before).
In contrast, in the partial update for the simple flow, we make a dump before _actually_ before any modifications are done. The problem there is that the decision and modifying the object is done after the dump is made, so we cannot not save the dump in case of the full update. It leads to the situation, when we save dump before two times. However, they are different dumps, because we can first modify some fields in the partial update and then pass the modified flow to the full update.

What do you think? Should make one way or another in both places? Should we fix it somehow differently in both cases? (move the decision to earlier stage, before modifying the flow)?

IvanChupin (approved):

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/service/haflow/HaFlowUpdateService.java
niksv: I think 2 "before" is better, because we will have more info for an operator

niksv (approved):

Issue #5170, opened 2023-04-19
[TEST] add tests to cover #5165 (HA-flow protected path is equal to main path issue)


artHalMirantis:
cause due to resources lack the issue is not covered by autotests
so we need to create one to cover it

yuliiamir:
The appropriate TC has been added.

PR #5350, opened 2023-08-25
5170: [TEST] Ha flow: Verify the absence of common ISLs in main and protected paths

yuliiamir:
related to #5170

### DIFF
--- a/src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/FlowPathModel.groovy
+++ b/src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/FlowPathModel.groovy
@@ -1,0 +1,20 @@
+package org.openkilda.functionaltests.helpers.model
+
+
+import org.openkilda.testing.model.topology.TopologyDefinition.Isl
+
+import groovy.transform.ToString
+import groovy.transform.TupleConstructor
+
+@TupleConstructor
+@ToString(includeNames = true, includePackage = false)
+class FlowPathModel {
+    String flowId
+    PathModel path
+    PathModel protectedPath
+
+    List<Isl> getCommonIslsWithProtected(boolean isForward = true) {
+        isForward ? path.forward.getInvolvedIsls().intersect(protectedPath.forward.getInvolvedIsls()) :
+                path.reverse.getInvolvedIsls().intersect(protectedPath.reverse.getInvolvedIsls())
+    }
+}
--- a/src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/HaFlowAllEntityPaths.groovy
+++ b/src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/HaFlowAllEntityPaths.groovy
@@ -1,0 +1,47 @@
+package org.openkilda.functionaltests.helpers.model
+
+import org.openkilda.northbound.dto.v2.haflows.HaFlowPaths
+import org.openkilda.testing.model.topology.TopologyDefinition
+
+import groovy.transform.EqualsAndHashCode
+import groovy.transform.ToString
+import groovy.transform.builder.Builder
+
+@EqualsAndHashCode(excludes = "topologyDefinition")
+@Builder
+@ToString(includeNames = true, excludes = 'topologyDefinition', includePackage = false)
+class HaFlowAllEntityPaths {
+
+    FlowPathModel sharedPath
+    List<FlowPathModel> subFlowPaths
+
+    TopologyDefinition topologyDefinition
+
+    HaFlowAllEntityPaths(HaFlowPaths haFlowPaths, TopologyDefinition topologyDefinition) {
+        this.sharedPath = new FlowPathModel(
+                path: new PathModel(
+                        forward: new Path(haFlowPaths.sharedPath.forward, topologyDefinition),
+                        reverse: new Path(haFlowPaths.sharedPath.reverse, topologyDefinition)
+                ),
+                protectedPath: !sharedPath?.protectedPath ? null : new PathModel(
+                        forward: new Path(haFlowPaths.sharedPath.protectedPath.forward, topologyDefinition),
+                        reverse: new Path(haFlowPaths.sharedPath.protectedPath.reverse, topologyDefinition)
+                ))
+
+        this.subFlowPaths = haFlowPaths.subFlowPaths.collect { subFlow ->
+            new FlowPathModel(
+                    flowId: subFlow.flowId,
+                    path: new PathModel(
+                            forward: new Path(subFlow.forward, topologyDefinition),
+                            reverse: new Path(subFlow.reverse, topologyDefinition)
+                    ),
+                    protectedPath: !subFlow?.protectedPath ? null : new PathModel(
+                            forward: new Path(subFlow.protectedPath.forward, topologyDefinition),
+                            reverse: new Path(subFlow.protectedPath.reverse, topologyDefinition)
+                    )
+            )
+        }
+
+        this.topologyDefinition = topologyDefinition
+    }
+}
--- a/src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/HaFlowExtended.groovy
+++ b/src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/HaFlowExtended.groovy
@@ -1,0 +1,95 @@
+package org.openkilda.functionaltests.helpers.model
+
+import org.openkilda.functionaltests.helpers.builder.HaFlowBuilder
+import org.openkilda.northbound.dto.v2.haflows.HaFlow
+import org.openkilda.northbound.dto.v2.haflows.HaFlowPaths
+import org.openkilda.northbound.dto.v2.haflows.HaFlowSharedEndpoint
+import org.openkilda.northbound.dto.v2.haflows.HaSubFlow
+import org.openkilda.testing.model.topology.TopologyDefinition
+import org.openkilda.testing.service.northbound.NorthboundServiceV2
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import groovy.transform.EqualsAndHashCode
+import groovy.transform.ToString
+import groovy.transform.builder.Builder
+
+/* This class represents any kind of interactions with HA flow
+This class has to replace *HaFlowHelper in future
+ */
+@EqualsAndHashCode(excludes = 'northboundV2, topologyDefinition')
+@Builder
+@ToString(includeNames = true, excludes = 'northboundV2, topologyDefinition')
+class HaFlowExtended {
+    String haFlowId
+    String status
+    String statusInfo
+
+    HaFlowSharedEndpoint sharedEndpoint
+
+    long maximumBandwidth
+    String pathComputationStrategy
+    String encapsulationType
+    Long maxLatency
+    Long maxLatencyTier2
+    boolean ignoreBandwidth
+    boolean periodicPings
+    boolean pinned
+    Integer priority
+    boolean strictBandwidth
+    String description
+    boolean allocateProtectedPath
+
+    Set<String> diverseWithFlows
+    Set<String> diverseWithYFlows
+    Set<String> diverseWithHaFlows
+
+    List<HaSubFlow> subFlows
+
+    String timeCreate
+    String timeUpdate
+
+    @JsonIgnore
+    NorthboundServiceV2 northboundV2
+
+    @JsonIgnore
+    TopologyDefinition topologyDefinition
+
+    HaFlowExtended(HaFlow haFlow, NorthboundServiceV2 northboundV2, TopologyDefinition topologyDefinition) {
+        this.haFlowId = haFlow.haFlowId
+        this.status = haFlow.status
+        this.statusInfo = haFlow.statusInfo
+        this.sharedEndpoint = haFlow.sharedEndpoint
+        this.maximumBandwidth = haFlow.maximumBandwidth
+        this.pathComputationStrategy = haFlow.pathComputationStrategy
+        this.encapsulationType = haFlow.encapsulationType
+        this.maxLatency = haFlow.maxLatency
+        this.maxLatencyTier2 = haFlow.maxLatencyTier2
+        this.ignoreBandwidth = haFlow.ignoreBandwidth
+        this.periodicPings = haFlow.periodicPings
+        this.pinned = haFlow.pinned
+        this.priority = haFlow.priority
+        this.strictBandwidth = haFlow.strictBandwidth
+        this.description = haFlow.description
+        this.allocateProtectedPath = haFlow.allocateProtectedPath
+
+        this.diverseWithFlows = haFlow.diverseWithFlows
+        this.diverseWithYFlows = haFlow.diverseWithYFlows
+        this.diverseWithHaFlows = haFlow.diverseWithHaFlows
+        this.subFlows = haFlow.subFlows
+        this.timeCreate = haFlow.timeCreate
+        this.timeUpdate = haFlow.timeUpdate
+
+        this.northboundV2 = northboundV2
+        this.topologyDefinition = topologyDefinition
+    }
+
+    HaFlowAllEntityPaths getAllEntityPaths() {
+        HaFlowPaths haFlowPaths = northboundV2.getHaFlowPaths(haFlowId)
+        new HaFlowAllEntityPaths(haFlowPaths, topologyDefinition)
+    }
+
+    static HaFlowBuilder build(SwitchTriplet swT, NorthboundServiceV2 northboundV2, TopologyDefinition topologyDefinition,
+                               boolean useTraffgenPorts = true, List<Integer> busyEndpoints = []) {
+        return new HaFlowBuilder(swT, northboundV2, topologyDefinition, useTraffgenPorts, busyEndpoints)
+    }
+}
--- a/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/haflows/HaFlowPathsSpec.groovy
+++ b/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/haflows/HaFlowPathsSpec.groovy
@@ -1,12 +1,13 @@
 package org.openkilda.functionaltests.spec.flows.haflows
 
+import org.openkilda.functionaltests.helpers.model.HaFlowExtended
+
 import groovy.util.logging.Slf4j
 import org.openkilda.functionaltests.HealthCheckSpecification
 import org.openkilda.functionaltests.error.haflow.HaFlowNotFoundExpectedError
 import org.openkilda.functionaltests.extension.failfast.Tidy
 import org.openkilda.functionaltests.extension.tags.Tags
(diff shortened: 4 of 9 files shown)


src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/GroupFlowPathModel.groovy
```diff
@@ -0,0 +1,13 @@
+package org.openkilda.functionaltests.helpers.model
```
pkazlenka, line 1: Do we use group flow paths somewhere in tests? If no, I guess we don't need to create classes and deserialize it.

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/FlowPathModel.groovy
```diff
@@ -0,0 +1,29 @@
+package org.openkilda.functionaltests.helpers.model
```
pkazlenka, line 1: It becomes quite crowded in this package. I'd recommend to add new package, e.g. `path`.

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/HaFlowAllEntityPaths.groovy
```diff
@@ -0,0 +1,56 @@
...
+import groovy.transform.ToString
+import groovy.transform.builder.Builder
+
+@EqualsAndHashCode(excludes = "topologyDefinition")
+@Builder
+@ToString(includeNames = true, excludes = 'topologyDefinition', includePackage = false)
+class HaFlowAllEntityPaths {
+
+    FlowPathModel shared
+    List<FlowPathModel> subFlowPaths
```
pkazlenka, line 17: Let's be consistent. Either `sharedPaths` or `subFlows`.

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/HaFlowExtended.groovy
```diff
@@ -0,0 +1,94 @@
...
+        this.diverseWithHaFlows = haFlow.diverseWithHaFlows
+        this.subFlows = haFlow.subFlows
+        this.timeCreate = haFlow.timeCreate
+        this.timeUpdate = haFlow.timeUpdate
+
+        this.northboundV2 = northboundV2
+        this.topologyDefinition = topologyDefinition
+    }
+
+    HaFlowPaths getFlowPath() {
```
pkazlenka, line 86: Is this one-string method used somewhere, except of the next method? If no, it can be removed.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/haflows/HaFlowPathsSpec.groovy
```diff
@@ -38,4 +35,30 @@ class HaFlowPathsSpec extends HealthCheckSpecification {
         def actualException = thrown(HttpClientErrorException)
         new HaFlowNotFoundExpectedError(~/HA-flow ${flowId} not found\./).matches(actualException)
     }
+
+    @Tidy
+    def "HA flow main path is not overlapped with protected path"() {
+        assumeTrue(useMultitable, "HA-flow operations require multiTable switch mode")
+        given: "An HA-flow with protected path"
+        def swT = topologyHelper.findSwitchTripletForHaFlowWithProtectedPaths()
+        assumeTrue(swT != null, "These cases cannot be covered on given topology:")
```
pkazlenka, line 44: You won't need this special assumption if you follow the approach from SwitchPairs class.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/haflows/HaFlowPathsSpec.groovy
```diff
@@ -38,4 +35,30 @@ class HaFlowPathsSpec extends HealthCheckSpecification {
...
+
+        def haFlowRequest = haFlowHelper.randomHaFlow(swT)
+        haFlowRequest.allocateProtectedPath = true
+        HaFlowExtended haFlow = swT.addHaFlow(haFlowRequest)
```
pkazlenka, line 48: I'd recommend to hide this creation into SwitchTriplet class. Keeping in mind, that random Ha-Flow won't work here, I'd recommend to use builder. I.e. code in spec should look like this:
```
HaFlowExtended haFlow = swT.customHaFlow() //this is static method that returns builder
                                                        .withProtectedPath(true)
                                                        .with....() //for now we don't need anything else
                                                        .create()
```

For easier cases, `swt.createRandomHaFlow()` would be good.
You can either call HaFlowHelper's methods from SwitchTriplet, or copy-paste them, or rewrite yourself, whatever suits you more at the moment.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/haflows/HaFlowPathsSpec.groovy
```diff
@@ -38,4 +35,30 @@ class HaFlowPathsSpec extends HealthCheckSpecification {
         def actualException = thrown(HttpClientErrorException)
         new HaFlowNotFoundExpectedError(~/HA-flow ${flowId} not found\./).matches(actualException)
     }
+
+    @Tidy
+    def "HA flow main path is not overlapped with protected path"() {
+        assumeTrue(useMultitable, "HA-flow operations require multiTable switch mode")
```
pkazlenka, line 41: Remove this, single-table mode is already obsolete in kilda.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/haflows/HaFlowPathsSpec.groovy
```diff
@@ -38,4 +35,30 @@ class HaFlowPathsSpec extends HealthCheckSpecification {
...
+        given: "An HA-flow with protected path"
+        def swT = topologyHelper.findSwitchTripletForHaFlowWithProtectedPaths()
+        assumeTrue(swT != null, "These cases cannot be covered on given topology:")
+
+        def haFlowRequest = haFlowHelper.randomHaFlow(swT)
+        haFlowRequest.allocateProtectedPath = true
+        HaFlowExtended haFlow = swT.addHaFlow(haFlowRequest)
+        assert haFlow.allocateProtectedPath
```
pkazlenka, line 49: We already have this assertion in HaFlowProtectedPathSpec. No need to check it every time.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/haflows/HaFlowPathsSpec.groovy
```diff
@@ -38,4 +35,30 @@ class HaFlowPathsSpec extends HealthCheckSpecification {
...
+
+        when: "Request path for HA flow with protected path"
+        def haFlowPaths = haFlow.getAllEntityPaths()
+        assert haFlowPaths.subFlowPaths.protectedPath.size() == 2
```
pkazlenka, line 53: This assertion should be in `then` block.

pkazlenka, line 53: This check is not necessary if you verify that each subflow path has protected path.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/haflows/HaFlowPathsSpec.groovy
```diff
@@ -38,4 +35,30 @@ class HaFlowPathsSpec extends HealthCheckSpecification {
...
+
+        then: "HA flow main and protected paths do not have common ISLs"
+        haFlowPaths.subFlowPaths.each {subFlowPath ->
```
pkazlenka, line 56: What if rewrite loop using `every` method? E.g. like
```
haFlowPaths.subFlowPaths.every {it.forward.getCommonIsls(it.protectedPath.forward).isEmpty() }
```
Also, now, when you have FlowPathModel, you can add method .getCommonIslsWithProtected(), so your assertion would look like 
```
haFlowPaths.subFlowPaths.every {it.getCommonIslsWithProtected().isEmpty() }
```

pkazlenka (changes requested):

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/haflows/HaFlowPathsSpec.groovy
yuliiamir: Done.
Do we need to remove this verification from the rest of the Ha-flow-related tests?

pkazlenka: I think for now it's not necessary (unless you want to solve a bunch of conflicts after eventual merge/rebase).

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/haflows/HaFlowPathsSpec.groovy
yuliiamir: I added this check` haFlowPaths.subFlowPaths.protectedPath.size() == 2 `
to clarify that Ha flow has a protected path.  The following check (`getCommonIsls()`) will be skipped as we don't have entities to compare (decreasing time execution).

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/haflows/HaFlowPathsSpec.groovy
yuliiamir: The appropriate method has been added to the FlowPathModel.
As for the usage of each{}, let's keep it as there is no need to iterate through all elements when we find a corrupted one.

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/GroupFlowPathModel.groovy
yuliiamir: sure, let's keep it simple

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/FlowPathModel.groovy
yuliiamir: Done.

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/HaFlowAllEntityPaths.groovy
yuliiamir: Done.

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/HaFlowExtended.groovy
yuliiamir: Done.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/haflows/HaFlowPathsSpec.groovy
yuliiamir: The new logic has been added. Could you please check?

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/Helper.groovy
```diff
@@ -0,0 +1,21 @@
...
+import com.github.javafaker.Faker
+
+import java.text.SimpleDateFormat
+
+class Helper {
+
+    static def random = new Random()
+    static def faker = new Faker()
+
+    static String generateDescription(String flowType) {
```
yuliiamir, line 12: Each flow-specific helper has generateDescription() and generateFlowId() methods, it seems like we need to move these methods to the separate class and reuse for different flow payload creation.
WDYT: As for the param 'flowType', we can add enum for the main flow types (ha-, y, regular).

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/SwitchTriplet.groovy
```diff
@@ -42,4 +66,64 @@ class SwitchTriplet {
     static Closure ONE_SUB_FLOW_IS_ONE_SWITCH_FLOW = {
         SwitchTriplet swT -> swT.shared == swT.ep1 && swT.shared != swT.ep2
     }
-}
+
+    /**
+     * Adds ha-flow and waits for it to become UP.
+     */
+    HaFlowExtended addHaFlow(HaFlowCreatePayload flow) {
```
yuliiamir, line 73: unused method (deleted locally)

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/haflows/HaFlowPathsSpec.groovy
```diff
@@ -38,4 +34,25 @@ class HaFlowPathsSpec extends HealthCheckSpecification {
         def actualException = thrown(HttpClientErrorException)
         new HaFlowNotFoundExpectedError(~/HA-flow ${flowId} not found\./).matches(actualException)
     }
+
+    @Tidy
+    def "HA flow main path is not overlapped with protected path"() {
+        given: "An HA-flow with protected path"
+        def swT = topologyHelper.findSwitchTripletForHaFlowWithProtectedPaths()
+        assumeTrue(swT != null, "These cases cannot be covered on given topology:")
+        HaFlowExtended haFlow = swT.customHaFlow().withProtectedPath(true).create()
```
pkazlenka, line 43: Good!

pkazlenka (approved):

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/SwitchTriplet.groovy
```diff
@@ -1,24 +1,48 @@
...
+@EqualsAndHashCode(excludes = 'northboundV2Service, topologyDefinition')
 class SwitchTriplet {
     Switch shared
     Switch ep1
     Switch ep2
     List<List<PathNode>> pathsEp1
     List<List<PathNode>> pathsEp2
 
+    @JsonIgnore
+    NorthboundServiceV2 northboundV2Service
```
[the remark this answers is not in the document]
pkazlenka: HaFlow and SwitchTriplet is really logically bonded. You admit yourself, that you can't create HaFlow without SwitchTriplet as mandatory parameter. So, between 
`HaFlowFactory.create(switchTriplet, params)` and `SwitchTriplet.createHaFlow(params)` the second seems more readable for me.
HTTP (and kafka, and DB) client must present in all business entity models so the instances could read/update their state without passing clients each time you need to do something. I think that model 
```
class Flow {
    def create(){...}
    def update(){...}
    def delete(){...}
    def getStatus(){...}
}
```
is much more clear than
```
class FlowCreator{}
class FlowUpdater{}
class FlowStatusVerifier{}
```
Actually, if you can create a PoC, which demonstrates how to initialize an object, that can represent HaFlow with `new()`, it would be even better.

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/Helper.groovy
```diff
@@ -0,0 +1,21 @@
+package org.openkilda.functionaltests.helpers
+
+import com.github.javafaker.Faker
+
+import java.text.SimpleDateFormat
+
+class Helper {
```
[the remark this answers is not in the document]
pkazlenka: Probably worth moving to Flow (or, temporary, to HaFlow) model. So it could be called like 
`Flow.generateId()`/`Flow.generateDescription()`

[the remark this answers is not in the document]
yuliiamir: Done.

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/SwitchTriplet.groovy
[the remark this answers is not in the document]
rtretyak: Switches that are used to build haFlow is just another param similar to its bandwidth, name and protectedPath flag.
Switches are not really 'logically bound' to any type of flow. They represent switch entities. Either 1 switch, switchPair or switchTriplet. And moreover **Switches** are not logically bound to neither northbound nor kafka and not **responsible** for creating or updating **flows**.
> SwitchTriplet.createHaFlow(params)

So no, switchTriplet is not 'creating a flow'. Northbound is creating a flow, using a HaFlow entitiy which represents a flow, which uses SwitchTriplet entity which represents switches over which this flow is going to be built.


> HTTP (and kafka, and DB) client must present in all business entity models so the instances could read/update their state without passing clients each time you need to do something.

Unless you want to follow any of established software design patterns, then probably yes, you can do this. Universe will not collapse, but this is not something you will often see anywhere else, except maybe an ORM pattern. But again, you could probably defend your point if it was applied to HaFlow and you called ```haFlow.create()```, but currently it is called on a 'switchTriplet' itself which doesn't make sense. Not going to describe what is SRP/SOLID here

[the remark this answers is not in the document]
pkazlenka: @rtretyak , you are right. We shouldn't break the logic to just minimize amount of parameters which must be passed to the methods.
@yuliiamir , let's create flows using factory classes (e.g. `HaFlowFactory.custom(swT).withProtectedPath().create()`). 

@rtretyak , @yuliiamir , or maybe you have better ideas?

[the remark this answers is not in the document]
yuliiamir: The appropriate changes have been added.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/haflows/HaFlowPathsSpec.groovy
yuliiamir: Let's add all required changes to the SwitchTriplet (using approach from SwitchPairs) as a part of another PR.

PR #5364, opened 2023-09-05
Fix PathValidator to consider all flows in a diverse group

### LINKED ISSUE #5351: Path check: consider all flows in the diverse group when validate for intersections
Currently, when executing the path validation and providing the diverse flow ID it shows intersections only with this flow's segments.
Add ability to validate whether there are intersections with a segment of any of the flow in the flow group.

Steps to reproduce are: 
1. Create two or more flows that have the same diverse group ID
2. Execute the path validation with the path that intersects with the first flow, but provide in the parameter diverseWith the second flow.
Expected result: validation says that there are intersections 
Actual result: validation doesn't say anything about the intersections.

dmitrii-beliakov:
This fix adds an ability to find intersections with all flows in a diverse group (instead of checking only with the specific flow).
Now, when a parameter `diverseWithFlow` is present, the system will fetch the diverse group ID from that flow and, then, validate all flows from that group for intersections with the given path.

(The parent branch must be integrated into develop first)
closes #5351


src-java/nbworker-topology/nbworker-storm-topology/src/test/java/org/openkilda/wfm/topology/nbworker/validators/PathValidatorTest.java
```diff
@@ -496,6 +506,46 @@ public void whenDiverseWith_andExistsIntersection_validatePathReturnsError() {
                         + " port 6 and destination switch 00:00:00:00:00:00:00:03 port 6.");
     }
 
+    @Test
+    void whenDiverseWith_andIntersectionWithGroup_validatePathReturnsAllIntersections() {
+        Switch switch1 = Switch.builder().switchId(SWITCH_ID_1).build();
+        Switch switch2 = Switch.builder().switchId(SWITCH_ID_2).build();
+        Switch switch3 = Switch.builder().switchId(SWITCH_ID_3).build();
```
pablomuri, line 513: I see some places where these switches are created. Maybe we can make them static?

pablomuri (approved):

src-java/nbworker-topology/nbworker-storm-topology/src/test/java/org/openkilda/wfm/topology/nbworker/validators/PathValidatorTest.java
dmitrii-beliakov: yes, could be, thanks for the suggestion. I'll refactor it

niksv (approved):

niksv:
this PR should not has label `bug` because it is a `bugfix`

PR #5370, opened 2023-09-11
Feature/path check structure

### LINKED ISSUE #5362: Path check: create a field for information and move appropriate messages from "error"
This enhancement adds a new field to the path validation response that contains information that is not necessary an error, just something that is true according to DB. For example, when a user enters a parameter diverse flow, but there is no flow with that ID, it is not an error from the point of view of building a path. In this case, the response should contain a valid or invalid result and at the same time a warning that the diverse from has not been found and therefore it wasn't a part of the path validation.
Another example, is when we have an intersection with a diverse flow. It means that this path will contain penalties (might be not the best path), but nevertheless could be built and is valid otherwise.
Similarly with other messages.
Info messages doesn't affect the value of the field "is_valid"

Response:
```json
{
    "is_valid" : true,
    "errors" : [],
    "information" : ["There is an intersection with the flow X on a segment ... "],
    "pce_response" : "The path has been computed successfully"
}
```

dmitrii-beliakov:
This change introduces a new field to the path check response called validationMessages. It contains messages from the validator; it gives information about what we know about the path. Errors is remained, but it contains errors if it is not possible to execute validations.
`isValid` in the path check response now returns true only if the PCE can compute the path.

It also contains a fix for a flow without diverse group issue: when a `diverse_with_flow` provided, but this flow doesn't contain a diverse group, only this flow in considered for the intersection checking (otherwise as before all flows from the diverse group are considered).

closes #5362

### DIFF
--- a/src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
+++ b/src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
@@ -49,6 +49,7 @@ import com.google.common.collect.Sets;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.collections4.map.LazyMap;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.storm.shade.com.google.common.collect.Maps;
@@ -66,6 +67,7 @@ import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
+@Slf4j
 public class PathValidator {
 
     private static final String LATENCY = "latency";
@@ -114,8 +116,7 @@ public class PathValidator {
                 .flatMap(Set::stream)
                 .collect(Collectors.toSet());
 
-        validationsResult.addAll(executeValidations(InputData.of(data),
-                getPerPathValidations(data)));
+        validationsResult.addAll(executeValidations(InputData.of(data), getPerPathValidations(data)));
 
         GetPathsResult pceResult = validatePathComputation(data);
         String pceResponse = pceResult.isSuccess() ? "The path has been computed successfully" :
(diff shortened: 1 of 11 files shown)


dmitrii-beliakov:
On the demo there were comments about the fields naming. In the end I decided to leave them as is. To me, it makes sense to preserve the source of the information. PCE is not an implementation detail, it is advertised as a feature of OpenKilda even on web, so I think it is OK to have it as field name.
Nevertheless, if there are any suggestions for the names, please write it here. If we agree on some other naming, I'll change it.

dmitrii-beliakov:
There are some problems with the branch hierarchy, I am looking into it, but most likely it will be resolved by merging the parent branches

niksv (approved):

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
```diff
@@ -114,17 +116,16 @@ public PathValidationResult validatePath(PathValidationData pathValidationData)
...
         GetPathsResult pceResult = validatePathComputation(data);
         String pceResponse = pceResult.isSuccess() ? "The path has been computed successfully" :
                 format("The path has not been computed successfully due to the following reasons: %s",
                         pceResult.getFailReasons().values());
 
         return PathValidationResult.builder()
-                .isValid(validationsResult.isEmpty() && pceResult.isSuccess())
-                .errors(new LinkedList<>(validationsResult))
+                .isValid(pceResult.isSuccess())
+                .validationMessages(new LinkedList<>(validationsResult))
```
IvanChupin, line 128: should we use addValidationMessages() method here and pass only the element(validationsResult) instead of  the current implementation? 
```suggestion
                .addValidationMessages(validationsResult)
```

IvanChupin (approved):
Great job!
Just one suggestion

src-java/nbworker-topology/nbworker-storm-topology/src/main/java/org/openkilda/wfm/topology/nbworker/validators/PathValidator.java
dmitrii-beliakov: Actually, we can't do that. It works a bit differently. We add a `@Singular` annotation on a collection so that we have a shortcut for adding a single element to this collection. The method accepts the collection generic type. In this case `validationMessages` is a `List<String>`, so `addValidationMessage` accepts a String.
This is just a syntactic sugar for 
```java
    .validationMessages(Collections.singletonList("foo");
    .addValidationMessage("foo");
```

PR #5386, opened 2023-09-15
[TEST] Fix victoriametrics config key

pkazlenka:
* Updated confd key for victoriametrics

### DIFF
--- a/confd/templates/functional-tests/kilda.properties.tmpl
+++ b/confd/templates/functional-tests/kilda.properties.tmpl
@@ -9,7 +9,7 @@ zookeeper.connect_string = {{ getv "/kilda_tests_zookeeper_hosts" "n/a" }}/{{ ge
 kafka.bootstrap.server={{ getv "/kilda_tests_kafka_hosts" "n/a" }}
 kafka.bootstrap.server.internal={{ getv "/kilda_tests_kafka_hosts_internal" "n/a" }}
 legacy.tsdb.endpoint=http://{{ getv "/kilda_tests_opentsdb_hosts" }}:{{ getv "/kilda_tests_opentsdb_port" }}
-tsdb.endpoint=http://{{ getv "/kilda_tests_victoriametrics_hosts" }}:{{ getv "/kilda_tests_victoriametrics_port" }}
+tsdb.endpoint=http://{{ getv "/kilda_tests_victoriametrics_host" }}:{{ getv "/kilda_victoriametrics_read_port" }}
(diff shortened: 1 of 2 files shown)


confd/vars/test-vars.yaml
```diff
@@ -10,8 +10,7 @@ kilda_tests_zookeeper_hosts: "localhost"
 
 kilda_tests_opentsdb_hosts: "localhost"
 kilda_tests_opentsdb_port: "4242"
-kilda_tests_victoriametrics_hosts: "localhost"
-kilda_tests_victoriametrics_port: "8428"
```
niksv, line 14: Why did you remove it? We do not want to test Victoria metrics in func tests?

pkazlenka: I found that we can (and, in fact, we did) use the application variables from main.yaml instead of test-vars.yam.

confd/templates/functional-tests/kilda.properties.tmpl
```diff
@@ -8,8 +8,8 @@ elasticsearch.endpoint={{ getv "/kilda_tests_logging_elasticsearch_proto" "n/a"
 zookeeper.connect_string = {{ getv "/kilda_tests_zookeeper_hosts" "n/a" }}/{{ getv "/kilda_zookeeper_state_root" }}
 kafka.bootstrap.server={{ getv "/kilda_tests_kafka_hosts" "n/a" }}
 kafka.bootstrap.server.internal={{ getv "/kilda_tests_kafka_hosts_internal" "n/a" }}
-legacy.tsdb.endpoint=http://{{ getv "/kilda_tests_opentsdb_hosts" }}:{{ getv "/kilda_tests_opentsdb_port" }}
-tsdb.endpoint=http://{{ getv "/kilda_tests_victoriametrics_hosts" }}:{{ getv "/kilda_tests_victoriametrics_port" }}
+legacy.tsdb.endpoint=http://{{ getv "/kilda_opentsdb_hosts" }}:{{ getv "/kilda_opentsdb_port" }}
+tsdb.endpoint=http://{{ getv "/kilda_victoriametrics_host" }}:{{ getv "/kilda_victoriametrics_write_port" }}{{ getv "/kilda_tests_victoriametrics_query_path" }}
```
niksv, line 12: Note: I think there is a typo here. here we have write port: `kilda_victoriametrics_write_port` but I think tests only read. Maybe we need read port?

pkazlenka: Good catch. Fixed.

niksv (approved):

Issue #5346, opened 2023-08-23
Collect application logs if test is failed


pkazlenka:
It is complicated to understand why some of the tests failed during full regression run: there are gigabytes of application logs collected. On  the other hand, without logs, it is hard to debug flaky tests. So it would be nice to collect the logs for failed tests only and attach them to test report. For junit.xml there's CDATA field, where such an information can be stored. But for spock report, the place to store the logs of failed tests should be investigated.
The mechanism should be the next:
1. Test start time should be remembered in pre-test.
2. If the test fails, some mechanism (I guess, it will be a spock listener), detects failure.
3. Listener, having test start and test end time, dumps log from kibana (or directly from elasticsearch) between time when test started and test finished.
4. Listener attaches log to the failed test report, so developer can see what has happened to the system exactly when test failed.

PR #5387, opened 2023-09-15
#5346: [TEST] Collect application logs if test is failed

pkazlenka:
Implements #5346

* Creates json file with kilda application logs in build/logs/spec/<specgroup>/<specname>/ directory
if test is failed.

### DIFF
--- a/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/listeners/CollectFailedTestLogsListener.groovy
+++ b/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/listeners/CollectFailedTestLogsListener.groovy
@@ -1,0 +1,48 @@
+package org.openkilda.functionaltests.listeners
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.spockframework.runtime.model.ErrorInfo
+import org.spockframework.runtime.model.IterationInfo
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.web.client.RestTemplate
+
+import java.time.ZoneOffset
+import java.util.concurrent.ConcurrentHashMap
+
+import static java.time.LocalDateTime.now
+import static java.time.format.DateTimeFormatter.ISO_DATE_TIME
+import static org.openkilda.functionaltests.listeners.LogParallelSpecsListener.getIterationPath
+
+class CollectFailedTestLogsListener extends AbstractSpringListener{
+    private ConcurrentHashMap<String, String> startTime = new ConcurrentHashMap<>()
+    @Value('${elasticsearch.endpoint}')
+    private String elasticSearchEndpoint
+    @Value('${elasticsearch.index}')
+    private String elasticSearchIndex
+
+
+    @Override
+    void beforeIteration(IterationInfo iterationInfo) {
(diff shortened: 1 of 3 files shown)


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/listeners/CollectFailedTestLogsListener.groovy
```diff
@@ -0,0 +1,59 @@
...
+
+    private static String utcTimeNow() {
+        return now(ZoneOffset.UTC).format(ISO_DATE_TIME)
+    }
+
+    private static File getTargetLogFile(ErrorInfo error) {
+        def specGroupName = error.getMethod().getParent().getPackage().split('\\.').last()
+        def specName = error.getMethod().getParent().getName()
+        def iterationName = error.getMethod().getIteration().getDisplayName()
+        return new File("build/logs/spec/${specGroupName}/${specName}/${iterationName}.server.log.json")
```
pkazlenka, line 53: Please, let me know it there's a way to retrieve gradle's build directory root in runtime.

dmitrii-beliakov: Not sure how to use gradle API from here, but you can save something as a system property in gradle.build and then read it here. For example, in a random test I had opened it looks like this:
```java
getClass().getClassLoader().getResourceAsStream("topology.properties")
```
then if you write sources root to a property, then you can get it here back.
But maybe this is overkill for this case.

dmitrii-beliakov (approved):

rtretyak (approved):

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/listeners/CollectFailedTestLogsListener.groovy
```diff
@@ -0,0 +1,52 @@
...
+                        "@timestamp:[${startTime} TO ${utcTimeNow()}]", String.class).getBody()
+        def beautifiedString = objectMapper.readValue(logs, Object.class)
+        objectMapper.writerWithDefaultPrettyPrinter().writeValue(getTargetLogFile(error), beautifiedString)
+    }
+
+    private static String utcTimeNow() {
+        return now(ZoneOffset.UTC).format(ISO_DATE_TIME)
+    }
+
+    private static File getTargetLogFile(ErrorInfo error) {
```
rtretyak, line 44: It'd be a good idea for this and https://github.com/telstra/open-kilda/blob/558e34f3f9476aa0f83b024657ff982d862271dc/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/listeners/LogParallelSpecsListener.groovy#L43
to find 'logs directory' for current test using some common shared method

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/listeners/CollectFailedTestLogsListener.groovy
pkazlenka: Hardcoding would be easier, I'm afraid.

PR #5409, opened 2023-09-29
[TEST] Fix Mirror and some other tests

pkazlenka:
* Fixed Mirror tests to not fail if there are no traffgens
* Fixed/ignored tests that failed because of minor bugs in app or in themselves.

### DIFF
--- a/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/grpc/LogSpec.groovy
+++ b/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/grpc/LogSpec.groovy
@@ -85,9 +85,10 @@ class LogSpec extends GrpcBaseSpecification {
         then: "Current remote log server configuration is deleted"
         response.deleted
 
+        /* https://github.com/telstra/open-kilda/issues/5408
         def response1 = grpc.getRemoteLogServerForSwitch(sw.address)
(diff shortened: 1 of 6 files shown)


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/grpc/LogSpec.groovy
```diff
@@ -85,9 +85,10 @@ class LogSpec extends GrpcBaseSpecification {
         then: "Current remote log server configuration is deleted"
         response.deleted
 
+        /* https://github.com/telstra/open-kilda/issues/5408
```
dmitrii-beliakov, line 88: A minor thing: I'd recommend to use `*` on all lines of the multiline comment, in the same manner as in copywrite.

dmitrii-beliakov (approved):

yuliiamir (approved):

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/grpc/LogSpec.groovy
pkazlenka: I'll do it in the next PR's. I wouldn't like to re-start PR regression for one cosmetic fix.

PR #5413, opened 2023-10-02
Victoria metrics url fixes for multicluster victoria configuration.

IvanChupin:
added new parameter
kilda_victoriametrics_read_path. 
On the **RIGEL/VEGA/PROD** environments the value for this parameter should be set to 
/select/0
Since we have multi cluster configuration for victoria Db on the rugel/vega/prod envs

This PR should be tested at least on the RIGEL env.

### DIFF
--- a/confd/templates/gui/application.properties.tmpl
+++ b/confd/templates/gui/application.properties.tmpl
@@ -11,6 +11,15 @@ server.ssl.key-alias = kilda
 server.ssl.key-store = classpath:keystore-kilda.jks
 server.ssl.key-store-password = openkilda
 
+#SSL Victoria configuration. Configure the SSL keystore file for Victoria Metrics in Java with the certificate required
+#for the HTTPS schema. Provide this file to enable REST API calls over HTTPS to the Victoria DB.
+victoria.trust.store=file:./victoria-keystore-kilda.jks
+{{- if exists "/victoria_keystore_password" }}
+victoria.trust.store.password = {{ getv "/victoria_keystore_password" }}
+{{- else }}
+victoria.trust.store.password =
+{{- end }}
+
 #Derby database configuration (In Memory)
 spring.jpa.database=default
 spring.jpa.properties.hibernate.dialect={{ getv "/kilda_gui_db_dialect" }}
@@ -40,7 +49,7 @@ opentsdb.base.url=http://{{ getv "/kilda_opentsdb_hosts" }}:{{ getv "/kilda_open
 opentsdb.metric.prefix = {{ getv "/kilda_opentsdb_metric_prefix" }}
 
 #VICTORIA METRICS Base URL and metric prefix
-victoria.base.url=http://{{ getv "/kilda_victoriametrics_host" }}:{{ getv "/kilda_victoriametrics_read_port" }}/prometheus
+victoria.base.url={{ getv "/kilda_victoria_read_url_schema" }}://{{ getv "/kilda_victoriametrics_host" }}:{{ getv "/kilda_victoriametrics_read_port" }}{{ getv "/kilda_victoriametrics_read_path" }}/prometheus
 
 #Kilda username and password
 kilda.username = {{ getv "/kilda_northbound_username" }}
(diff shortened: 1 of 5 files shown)


confd/vars/main.yaml
```diff
@@ -24,6 +24,7 @@ kilda_victoriametrics_host: "victoriametrics.pendev"
 kilda_victoriametrics_write_port: "4242"
 kilda_victoriametrics_read_port: "8428"
 kilda_victoriametrics_path: ""
+kilda_victoriametrics_read_path: ""
```
niksv, line 27: can `kilda_victoriametrics_path` be different with `kilda_victoriametrics_read_path` ?

IvanChupin: It should be different if not empty

confd/vars/main.yaml
```diff
@@ -24,6 +24,8 @@ kilda_victoriametrics_host: "victoriametrics.pendev"
...
+victoria_read_url_prefix: "http"
```
pkazlenka, line 28: I'd say that this is rather 'scheme' than 'url prefix'.

pkazlenka (approved):

confd/vars/main.yaml
IvanChupin: Rename it, right after will test it on Rigel.

confd/templates/gui/application.properties.tmpl
```diff
@@ -11,6 +11,11 @@ server.ssl.key-alias = kilda
 server.ssl.key-store = classpath:keystore-kilda.jks
 server.ssl.key-store-password = openkilda
 
+#SSL Victoria configurations
+#victoria.trust.store=file:./keystore-victoria-kilda.jks
```
dmitrii-beliakov, line 15: It is confusing for someone not familiar with the area. Please consider to add some more information such as why would one need it from this folder. Or maybe just remove comment if the usage is unlikely.

src-gui/src/main/java/org/openkilda/security/CustomRestTemplateConfiguration.java
```diff
@@ -0,0 +1,52 @@
...
+            CloseableHttpClient httpClient = HttpClients.custom().setSSLSocketFactory(sslConFactory).build();
+            LOGGER.info("httpClient has been initialized");
+            ClientHttpRequestFactory requestFactory = new HttpComponentsClientHttpRequestFactory(httpClient);
+            LOGGER.info("requestFactory has been initialized");
+            RestTemplate restTemplate = new RestTemplate(requestFactory);
+            LOGGER.info("restTemplate has been initialized");
+            return restTemplate;
+        } catch (Exception e) {
+            LOGGER.error("Failed to initialize RestTemplate with SSL context. Using a simple RestTemplate.", e);
+            return new RestTemplate();
```
dmitrii-beliakov, line 64: That seems suspicious to me. We try to use a template with a certificate, but for whatever reason we use a simple one? Isn't it the case that falling back to lower level should be done only in special cases/environments?

confd/vars/main.yaml
IvanChupin: renamed

confd/templates/gui/application.properties.tmpl
IvanChupin: thanks, will write more description

confd/templates/gui/application.properties.tmpl
```diff
@@ -11,6 +11,15 @@ server.ssl.key-alias = kilda
 server.ssl.key-store = classpath:keystore-kilda.jks
 server.ssl.key-store-password = openkilda
 
+#SSL Victoria configuration. Configure the SSL keystore file for Victoria Metrics in Java with the certificate required
+#for the HTTPS schema. Provide this file to enable REST API calls over HTTPS to the Victoria DB.
+victoria.trust.store=file:./victoria-keystore-kilda.jks
+{{- if exists "/victoria_keystore_password" }}
+victoria.trust.store.password = {{ getv "/victoria_keystore_password" }}
+{{- else }}
+# victoria.trust.store.password =
```
niksv, line 20: this is a comment. Is it correct?

dmitrii-beliakov (approved):

yuliiamir:
The issue has been fixed(all required environment have been checked). 
GraphVictoria
GraphTSDB

Victoria_Graph30minperiod
27842/3beac3e1-ec00-417a-968f-a1bf5a0502a3)

PR #5416, opened 2023-10-04
[TEST]: Fix victoria metrics key(query_path)

yuliiamir:


pkazlenka (approved):

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/listeners/CollectFailedTestLogsListener.groovy
```diff
@@ -28,7 +30,7 @@ class CollectFailedTestLogsListener extends AbstractSpringListener{
 
     @Override
     void error(ErrorInfo error) {
-        if (!isFailedInPreTest(error)) {
+        if (!isFailedInPreTest(error) && profile == "virtual") {
```
niksv, line 33: why only for vortual?

yuliiamir: There is no need to collect logs for profile=hardware as we can obtain them after tests execution. As for the "virtual", all logs will be unavailable due to the JJ cleanup.

dmitrii-beliakov (approved):

PR #5418, opened 2023-10-05
Add support of HA-flow endpoints update.

### LINKED ISSUE #5415: HA flow endpoints update
Implement endpoints update for HA flow: when an update changes only parameters that don't affect the path, do not invoke path recalculation logic. Parameters for endpoints update:
* port
* inner VLAN
* outer VLAN

dmitrii-beliakov:
This feature introduces an endpoint update for HA-flows: a partial update without paths recalculation. The condition for executing this kind of update is the updating only fields of endpoint that don't affect paths between them:
* port
* inner VLAN
* outer VLAN

_There is a small fix by the way an action deallocation resources on HA-flow delete now saves history correctly._

closes #5415

### DIFF
--- a/src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/haflow/delete/actions/DeallocateResourcesAction.java
+++ b/src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/haflow/delete/actions/DeallocateResourcesAction.java
@@ -47,7 +47,7 @@ public class DeallocateResourcesAction extends
                     resourcesManager.deallocateHaFlowResources(resources));
 
             FlowHistoryService.using(stateMachine.getCarrier()).save(HaFlowHistory
-                    .of(stateMachine.getHaFlowId())
+                    .of(stateMachine.getCommandContext().getCorrelationId())
                     .withAction("Flow resources have been deallocated")
                     .withDescription(format("The ha-flow resources for %s / %s have been deallocated",
                             resources.getForward().getPathId(), resources.getReverse().getPathId()))
--- a/src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/haflow/update/HaFlowUpdateFsm.java
+++ b/src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/haflow/update/HaFlowUpdateFsm.java
@@ -56,12 +56,14 @@ import org.openkilda.wfm.topology.flowhs.fsm.haflow.update.actions.RevertFlowAct
 import org.openkilda.wfm.topology.flowhs.fsm.haflow.update.actions.RevertFlowStatusAction;
 import org.openkilda.wfm.topology.flowhs.fsm.haflow.update.actions.RevertNewRulesAction;
 import org.openkilda.wfm.topology.flowhs.fsm.haflow.update.actions.RevertPathsSwapAction;
+import org.openkilda.wfm.topology.flowhs.fsm.haflow.update.actions.SkipResourceManagementOnEndpointsUpdateAction;
 import org.openkilda.wfm.topology.flowhs.fsm.haflow.update.actions.SwapFlowPathsAction;
 import org.openkilda.wfm.topology.flowhs.fsm.haflow.update.actions.UpdateFlowStatusAction;
 import org.openkilda.wfm.topology.flowhs.fsm.haflow.update.actions.UpdateHaFlowAction;
 import org.openkilda.wfm.topology.flowhs.fsm.haflow.update.actions.ValidateFlowAction;
 import org.openkilda.wfm.topology.flowhs.service.FlowProcessingEventListener;
 import org.openkilda.wfm.topology.flowhs.service.haflow.HaFlowGenericCarrier;
+import org.openkilda.wfm.topology.flowhs.utils.EndpointUpdateType;
 
 import io.micrometer.core.instrument.LongTaskTimer;
 import io.micrometer.core.instrument.LongTaskTimer.Sample;
@@ -85,6 +87,7 @@ public final class HaFlowUpdateFsm extends HaFlowPathSwappingFsm<HaFlowUpdateFsm
         HaFlowGenericCarrier, FlowProcessingEventListener> {
     private HaFlowRequest targetHaFlow;
     private FlowStatus newFlowStatus;
+    private EndpointUpdateType endpointUpdateType = EndpointUpdateType.NONE;
 
     public HaFlowUpdateFsm(@NonNull CommandContext commandContext, @NonNull HaFlowGenericCarrier carrier,
                            @NonNull String flowId,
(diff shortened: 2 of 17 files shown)


src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/haflow/delete/actions/DeallocateResourcesAction.java
```diff
@@ -47,7 +47,7 @@ public void perform(State from, State to, Event event, HaFlowDeleteContext conte
                     resourcesManager.deallocateHaFlowResources(resources));
 
             FlowHistoryService.using(stateMachine.getCarrier()).save(HaFlowHistory
-                    .of(stateMachine.getHaFlowId())
+                    .of(stateMachine.getCommandContext().getCorrelationId())
```
dmitrii-beliakov, line 50: actually there is a separate issue for that, I'll move that fix there.

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/haflow/update/HaFlowUpdateFsm.java
```diff
@@ -293,9 +296,14 @@ public Factory(@NonNull HaFlowGenericCarrier carrier, @NonNull Config config,
...
 
             builder.transition().from(State.NEW_RULES_REVERTED)
-                    .to(State.REVERTING_ALLOCATED_RESOURCES)
+                    .to(State.SKIP_REVERTING_RESOURCES_ALLOCATION)
                     .on(Event.NEXT);
 
+            builder.transitions().from(State.SKIP_REVERTING_RESOURCES_ALLOCATION)
+                    .toAmong(State.REVERTING_ALLOCATED_RESOURCES, State.REVERTING_FLOW)
+                    .onEach(Event.NEXT, Event.UPDATE_ENDPOINTS_ONLY)
+                    .perform(new SkipResourceManagementOnEndpointsUpdateAction(persistenceManager));
```
niksv, line 305: I'm not sure it will work.

It say: 
go from state `SKIP_REVERTING_RESOURCES_ALLOCATION` to state `REVERTING_FLOW` on event `UPDATE_ENDPOINTS_ONLY`. After changing state please run `SkipResourceManagementOnEndpointsUpdateAction` 

But this action it the only action which can call `UPDATE_ENDPOINTS_ONLY` event. I think wemust call this action it the previous transition 
```
builder.transition().from(State.NEW_RULES_REVERTED)
                    .to(State.SKIP_REVERTING_RESOURCES_ALLOCATION)
                    .on(Event.NEXT);
```

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/haflow/update/HaFlowUpdateFsm.java
```diff
@@ -335,6 +343,18 @@ public Factory(@NonNull HaFlowGenericCarrier carrier, @NonNull Config config,
...
+                    .from(State.FLOW_UPDATED)
+                    .to(State.RESOURCE_ALLOCATION_COMPLETED)
+                    .on(Event.UPDATE_ENDPOINTS_ONLY)
+                    .perform(new PostResourceAllocationAction(persistenceManager));
+
+            builder.transition()
+                    .from(State.OLD_RULES_REMOVED)
+                    .to(State.UPDATING_FLOW_STATUS)
+                    .on(Event.UPDATE_ENDPOINTS_ONLY)
+                    .perform(new SkipResourceManagementOnEndpointsUpdateAction(persistenceManager));
```
niksv, line 356: same

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/haflow/update/actions/UpdateHaFlowAction.java
```diff
@@ -58,13 +59,25 @@ protected Optional<Message> performWithResponse(State from, State to, Event even
...
         });
 
+        if (stateMachine.getEndpointUpdateType().isPartialUpdate()) {
+            stateMachine.setNewPrimaryPathIds(stateMachine.getOldPrimaryPathIds());
+            stateMachine.setNewProtectedPathIds(stateMachine.getOldProtectedPathIds());
+
+            FlowHistoryService.using(stateMachine.getCarrier()).save(HaFlowHistory
+                    .of(stateMachine.getCommandContext().getCorrelationId())
+                    .withAction("HA-flow endpoints update is executed."));
+            stateMachine.fire(Event.UPDATE_ENDPOINTS_ONLY);
```
niksv, line 78: FSM calls this action in state `FLOW_UPDATED`. But there is no any transition from state `FLOW_UPDATED` on event  `UPDATE_ENDPOINTS_ONLY`

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/utils/EndpointUpdateType.java
```diff
@@ -0,0 +1,179 @@
...
+
+        boolean subflowEndpointChanged = isSubflowEndpointChanged(
+                originalHaFlow.getHaSubFlows(), targetHaFlow.getSubFlows());
+
+        if (updateEndpointOnly && sharedEndpointChanged && subflowEndpointChanged) {
+            return ALL;
+        } else if (updateEndpointOnly && sharedEndpointChanged) {
+            return SHARED;
+        } else if (updateEndpointOnly && subflowEndpointChanged) {
+            return SUBFLOWS;
```
niksv, line 79: What if we need to update rules only at the subflow1 endpoint but not on subflow2? will we update both in this case?

niksv (changes requested):

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/utils/EndpointUpdateType.java
dmitrii-beliakov: I asked you this question last week that we don't have specified several things in the task, including this one. You answered that an HA flow works as a unit and we cannot see or update individual subflows.

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/fsm/haflow/update/HaFlowUpdateFsm.java
dmitrii-beliakov: yes, I now think I was confused on how and when this framework manages transitions and events (thanks to their brilliant javadocs) I am double checking this now, But it is a bit tricky to reproduce and to test. Thanks for catching it!

dmitrii-beliakov: still, it is not the case that the action is executed after. I added tracing into history and it says that the action is executed while the state is what is specified in from clause. But the confusing point is when the event goes into action, as I understand now it takes effect in the next transition.

dmitrii-beliakov:
Here are pictures of the new paths in the update FSM workflow. A new path skips old resources de-allocation (because the new are the same as the old):
HaFlowUpdateFsm_change_1
and resources reverting on error (again, because the new and the old rules are the same):
HaFlowUpdateFsm_change_2

dmitrii-beliakov:
As for the rules: at this point rule manager creates meters to add, which cannot be added, because they are already on the switch. So. apparently these rules could be skipped. However, it is not that straightforward how to implement it. One idea was to pass an additional flag to the rule manager from the build rules action and adjust the rule manager to use that. Probably there is a better way to implement it.
With the diverse group ID: changes in the diverse group ID says that the path might require recalculation and therefore endpoints update cannot be executed. However, database contains the current state and it is not clear whether the data from the update request lead to the diverse group ID change, it is not a part of the request object. There is a need to implement a method that will be invoked when a decision about the endpoints update is made, whether the diverse group ID changes.

PR #5438, opened 2023-10-11
Run unit tests via github actions in parallel

niksv:
If saves us 3.5 mins:

Before:
old_not_parallel

After:
only_parallel


.github/workflows/unittest.yml
```diff
@@ -34,6 +34,7 @@ jobs:
         mkdir -p $HOME/.gradle
         echo "org.gradle.jvmargs=-Xmx4096M" >> $HOME/.gradle/gradle.properties
```
dmitrii-beliakov, line 35: maybe it worth to add `Xms` parameter to some higher value (that's minimum memory, or in other words starting memory). Otherwise if it starts with 256M, it will resize/reallocate memory more times. Say 2048, I think we can afford it

dmitrii-beliakov (approved):

niksv:
It is related to unit tests. No addition tests required

PR #5437, opened 2023-10-11
Ignore lock-keeper tests in docker directory

niksv:

PR #5442, opened 2023-10-16
add lock-keeper/test to .gitignore

pablomuri:


pkazlenka:
It's a duplicate: https://github.com/telstra/open-kilda/pull/5437

PR #5456, opened 2023-10-26
Store datapoints into special storage to save memory

niksv:
New Storage was introduced in OpenTsdbTopology to save RAM.
It stores datapoints using string keys. It saves 40% of memory in average


pablomuri (approved):
LGTM

PR #5461, opened 2023-10-27
Change health check test for the MySQL container. Update docs.

### LINKED ISSUE #5459: MySQL container: improve healthcheck
Sometimes MySQL migration container fails with an error that the dependent container is unhealthy. However, that container appears healthy when inspecting later (yes, that's strange!). 

Improve MySQL container health check.

dmitrii-beliakov:
This change is an attempt to improve health checking for the MySQL container. It's difficult to say for sure, if there is only one problem, but probably, the previous test command could report that the container is healthy, however it doesn't yet accept connections (mysqladmin doesn't say that the DB is created and ready). Also, I suspect, that the health checking could report false information because of starting and stopping a temporary server during the initialization stage: this container starts a service, creates a DB called `kilda` then stops the temporary server, then starts the actual service. 

I added the command to select DB and run some query afterwards. Together, it says that selecting a DB is successful and therefore it's been created during the init stage. A select statement says that the server can execute queries.
Start period 

closes #5459

### DIFF
--- a/confd/templates/docker-compose/docker-compose.tmpl
+++ b/confd/templates/docker-compose/docker-compose.tmpl
@@ -140,7 +140,7 @@ services:
     ports:
       - 8101:3306
     healthcheck:
-      test: "mysql -u kilda -pkilda -e 'use kilda; select count(user());'"
+      test: "mysql -u kilda -pkilda --protocol=tcp -e 'use kilda; select count(user());'"
       interval: 3s
       timeout: 3s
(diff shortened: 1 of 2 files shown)


niksv (approved):

pkazlenka (approved):

yuliiamir:
Retested and redeployed Kilda 5 times without any failures. The problem seems to have disappeared.

PR #5469, opened 2023-11-09
[TEST] A unit test for generating a DOT-file representation of an FSM

dmitrii-beliakov:
This is a proof of concept how the DiagramBuilder could look like.
The problem of generating an FSM graph is that it requires to instantiate the FSM, which, in turn, requires to provide all its dependencies. Therefore, we can't have a single place for the diagram builder for all FSMs, because we don't have such a module that have all required dependencies (and probably we don't want to have it). 

Even with the single FSM builder there is this problem with providing dependencies and instantiating all required classes, which is a big deal. They are actually not needed, but the method calls and constructors are invoked in the eager manner and they actually use some business logic, so providing nulls or mocks is not an option for many of them.

Unfortunately, our FSMs are strongly coupled with the resources they use and the framework handles the entire FSM instead of its metadata, so, there are few options. At the same time, rewriting the structure of FSM seems to be a bigger task just for a sake of having a convenient way to generate a DOT file. I thought that it's good enough to spend 5-15 minutes and copy-paste the code with reformatting, then rewriting all FSMs.

This POC could be used as a template and once somebody wants to generate a picture of the FSM, they will create a similar test in the same module where the required FSM live (this gives access to all necessary dependencies). Once such test is created it could be committed for the future use. This doesn't look like a very good approach, but it is a working approach at hand.

I investigated alternative ways and, currently, I don't see how to resolve it quickly and other ways seem to be too expansive. We can leave this approach for now and we can replace it once somebody provides a better one.

Since this project is an open source one, we can use various tools available online for converting a DOT file to a picture. There is an Intellij Idea plugin for it as well, but it doesn't work on my PC. It's possible to make the conversion in java, but we'll need a library for it, seems like it is not worth it


dmitrii-beliakov:
I didn't mention that actions are not a part of the diagram. To make them there we'll need to created the actual action objects, or provide strings instead of the calls, or find some other way.

yuliiamir:
The appropriate .dot file and graph have been generated successfully. 
graph

niksv (approved):

IvanChupin (approved):

PR #5472, opened 2023-11-14
Remove opentsdb from Kilda-gui

IvanChupin:
This pr completely remove all the logic that process the opentsdb stats requests.(including datasource switch-checkbox on the flow-detail page)
either form gui part of the app and from the gui bff part.


src-gui/src/main/java/org/openkilda/security/CustomRestTemplateConfiguration.java
```diff
@@ -41,7 +41,7 @@ public class CustomRestTemplateConfiguration {
     @Value("${victoria.trust.store}")
     private Resource victoriaTrustStore;
 
-    @Value("${victoria.trust.store.password}")
+    @Value("${ссvictoria.trust.store.password}")
```
niksv, line 44: ccvictoria? what does it mean?

niksv (commented):
Also please remove otsdb props from `confd/templates/gui/application.properties.tmpl`

src-gui/src/main/java/org/openkilda/security/CustomRestTemplateConfiguration.java
IvanChupin: oh, thanks, that is typo

niksv (approved):

dmitrii-beliakov (approved):
As a matter of fact, no test is broken after these changes, which is an indicator of the test coverage.

niksv:
@dmitrii-beliakov 
image

yuliiamir:
It seems like a graph (image) has not been updated when Meter type is selected, but new data is received and the appropriate call has been executed.
FlowData

DataForMeterButImageForFlow

IvanChupin:
@yuliiamir 
We agreed with @niksv on the following solution:

**Graph: Flow**
Both directions at the same time
image
**Graph: Loss packets,** for forward both graphs should be for forward, for reverse both should be revers.
image
**Graph: Meter**, both direction should be represented at the same time
image

yuliiamir:
The issue has been fixed, victoria/otsdb toggle has been removed, the appropriate changes have been added to the Graph (described above).

pablomuri (approved):

PR #5478, opened 2023-11-17
[TEST]: Improvement: Flows(regular, y-, ha-) leftovers cleanup

yuliiamir:
Add cleanup before spec as sometimes during flow creation we have a timeout and this flow is created after some period of time. As a result, we lose its tracking and cannot delete it. Except for this problem, some tests fail due to this leftover.
In the scope of this task, the flow type was added to the flow name to make the debugging process easier.


src-java/testing/test-library/src/main/java/org/openkilda/testing/service/northbound/IslandNorthboundServiceV2.java
```diff
@@ -47,7 +50,21 @@ public class IslandNorthboundServiceV2 extends NorthboundServiceV2Impl {
     public List<FlowResponseV2> getAllFlows() {
         List<FlowResponseV2> result = super.getAllFlows();
         return result.stream().filter(flow -> switchIds.stream().anyMatch(swId ->
-                flow.getSource().getSwitchId().equals(swId) || flow.getDestination().getSwitchId().equals(swId)))
+                flow.getSource().getSwitchId().equals(swId) && Objects.isNull(flow.getYFlowId())))
```
yuliiamir, line 53: There is no need to check both source and destination as flow is deployed inside one topology lab.
As yFlow's sub-flows are regular flows, we need to exclude them from the list of regular flows.

pkazlenka (approved):

niksv (approved):

Issue #5390, opened 2023-09-18
Stabilize flaky tests


pkazlenka:
This is a placeholder issue to track the activities on tests stabilization

PR #5482, opened 2023-11-22
#5390: [TEST] Debug enabled for flaky Flow Monitoring spec

pkazlenka:
Implements #5390

* Failing Flow Monitoring tests now show why they can't find RTT stats
* Increased timeout to wait for links to resurrect after storm restart
* Added debug log to store the actual time of waiting in Wrappers.wait() method

### DIFF
--- a/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowMonitoringSpec.groovy
+++ b/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowMonitoringSpec.groovy
@@ -15,6 +15,7 @@ import static org.openkilda.functionaltests.helpers.Wrappers.wait
 import static org.openkilda.functionaltests.model.stats.Direction.FORWARD
 import static org.openkilda.functionaltests.model.stats.Direction.REVERSE
 import static org.openkilda.functionaltests.model.stats.FlowStatsMetric.FLOW_RTT
+import static org.openkilda.functionaltests.model.stats.Origin.FLOW_MONITORING
 import static org.openkilda.testing.Constants.WAIT_OFFSET
 
 import org.openkilda.functionaltests.HealthCheckSpecification
@@ -97,7 +98,7 @@ class FlowMonitoringSpec extends HealthCheckSpecification {
         flowHelperV2.addFlow(flow)
         //wait for generating some flow-monitoring stats
         wait(flowSlaCheckIntervalSeconds + WAIT_OFFSET) {
-            flowStats.rttOf(flow.getFlowId()).get(FLOW_RTT, FORWARD).hasNonZeroValues()
+            assert flowStats.rttOf(flow.getFlowId()).get(FLOW_RTT, FORWARD, FLOW_MONITORING).hasNonZeroValues()
         }
 
         def path = northbound.getFlowPath(flow.flowId)
@@ -160,7 +161,7 @@ and flowLatencyMonitoringReactions is disabled in featureToggle"() {
         flowHelperV2.addFlow(flow)
         //wait for generating some flow-monitoring stats
         wait(flowSlaCheckIntervalSeconds + WAIT_OFFSET) {
-            flowStats.rttOf(flow.getFlowId()).get(FLOW_RTT, REVERSE).hasNonZeroValues()
+            assert flowStats.rttOf(flow.getFlowId()).get(FLOW_RTT, REVERSE, FLOW_MONITORING).hasNonZeroValues()
         }
         pathHelper.convert(northbound.getFlowPath(flow.flowId)) == mainPath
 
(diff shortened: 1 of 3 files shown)


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/FlowMonitoringSpec.groovy
```diff
@@ -97,7 +97,7 @@ class FlowMonitoringSpec extends HealthCheckSpecification {
         flowHelperV2.addFlow(flow)
         //wait for generating some flow-monitoring stats
         wait(flowSlaCheckIntervalSeconds + WAIT_OFFSET) {
-            flowStats.rttOf(flow.getFlowId()).get(FLOW_RTT, FORWARD).hasNonZeroValues()
+            assert flowStats.rttOf(flow.getFlowId()).get(FLOW_RTT, FORWARD).hasNonZeroValues()
```
yuliiamir, line 100: Could you please add 'origin: flow_monitoring' to the stats filter?

pkazlenka: Done.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/xresilience/StormLcmSpec.groovy
```diff
@@ -140,7 +140,7 @@ class StormLcmSpec extends HealthCheckSpecification {
         !networkDeployed && wfmManipulator.deployTopology("network")
         srcBlockData && lockKeeper.reviveSwitch(islUnderTest.srcSwitch, srcBlockData)
         dstBlockData && lockKeeper.reviveSwitch(islUnderTest.dstSwitch, dstBlockData)
-        Wrappers.wait(WAIT_OFFSET + discoveryTimeout) {
+        Wrappers.wait(WAIT_OFFSET + discoveryTimeout * 3) {
```
dmitrii-beliakov, line 143: a minor thing or maybe not for this PR, but seems like the `WAIT_OFFSET` doesn't do its job well, now there are in fact two offsets. Also, as I understand, the bigger offset is only needed some times, but other times it's fine? Maybe then a retrying mechanism would work better here.

dmitrii-beliakov (approved):

yuliiamir (approved):

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/xresilience/StormLcmSpec.groovy
pkazlenka: In theory, we have several timeouts for different cases, like discoveryTimeout, statsCollectionInterval, etc. 
Also there's 'WAIT_OFFSET', which is some kind of common time gap that we allow kilda to late on. I.e, in this particular case, we expect that after reviving switches new ISL will be discovered not later than in `discoveryTimeout`, but giving it some additional time (offset).
Unfortunately, in most cases we don't have hard specifications, how long would this or that event should be waited for, before we decide that it takes too long. Also, having 'asynchronous world' inside kilda (I mean kafka and storm) doesn't make things easier for us.

So yes, in theory, we should report bug on each event which didn't happen in (`defined in specifications timeout` + WAIT_OFFSET), but in practice, I'm afraid, this will only end up in hundreds of issues, which nobody is going to investigate and fix :\ So we just try to find the amount of time which makes test both stable enough and still be able to discover kilda slownesses.

The highlighted part of the code has a logical mistake, though. We should always wait for only one  `defined in specifications timeout`, but for different offsets. Probably, we need something like `MINOR_WAIT_OFFSET`, `STANDARD_WAIT_OFFSET` and `LONG_WAIT_OFFSET`. 

`Wrappers.wait()` is a retrying mechanism, it is not an explicit waiter.

pkazlenka: Changed wait time calculation logic (it's a coincidence, that WAIT_OFFSET equals to discovery.timeout at the moment).

dmitrii-beliakov: With retrying I primarily mean that one execution requires to wait let's say 1s, other require 5s, we wait here 8s just to be sure. With retrying we assert something is true and on fail just repeat it a couple of times later and only then fail. So those times when the assertion is true, we don't wait 8s.

This method on 143 is retrying? I thought you need to pass the timeout and the retry timeout into it.

pkazlenka: Yes, it retries for given amount of time waiting for X seconds between tries (by default we don't  wait, so don't pass the second argument to function.
E.g. in this particular case, we will be waiting for 60 seconds (15 + 3 * 15), evaluating the closure again and again, until it returns true or until 60 seconds have passed.
Logarithmic (or fibonacci's) increase of wait interval between tries is not implemented, though, if you meant the one.

PR #5488, opened 2023-11-27
#5390: [TEST] Fix flaky HA-Flow ping test

pkazlenka:
Implements #5390

* Fixed incorrect condition in StatsResult.hasNonZeroValuesAfter() method which caused test to fail sometimes.
* Added waiter for ping success after changing flow incapsulation (seems like there was a race condition, when FlowFetcher couldn't receive flow incapsulation after flow update)

### DIFF
--- a/src-java/testing/test-library/src/main/java/org/openkilda/testing/service/tsdb/model/StatsResult.java
+++ b/src-java/testing/test-library/src/main/java/org/openkilda/testing/service/tsdb/model/StatsResult.java
@@ -62,7 +62,7 @@ public class StatsResult {
 
     public boolean hasNonZeroValuesAfter(long timestamp) {
         return dataPoints.keySet().stream()
-                .anyMatch(tstamp -> tstamp >= timestamp && dataPoints.get(tstamp) > 0);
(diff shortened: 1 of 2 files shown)


src-java/testing/test-library/src/main/java/org/openkilda/testing/service/tsdb/model/StatsResult.java
```diff
@@ -62,7 +62,7 @@ public boolean hasNonZeroValues() {
 
     public boolean hasNonZeroValuesAfter(long timestamp) {
         return dataPoints.keySet().stream()
-                .anyMatch(tstamp -> tstamp >= timestamp && dataPoints.get(tstamp) > 0);
+                .anyMatch(tstamp -> tstamp >= timestamp && dataPoints.get(tstamp) != 0);
```
yuliiamir, line 65: Is this fix for the TC 'Unable to ping one of the HA-subflows via periodic pings if related ISL is broken' or for some other HA-Flow-related cases?

pkazlenka: Yep.

dmitrii-beliakov (approved):

PR #5489, opened 2023-11-28
Improve dump payload format to not display empty protected path objects.

dmitrii-beliakov:
After this fix the history dump in the history entry of HA-flow no longer contains fields `protected_forward_path` and `protected_reverse_path` if they are empty: that is all fields are null and lists are empty.


src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/validation/JsonIncludeObjectHavingNonNullFieldOrNonEmptyList.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+import java.util.List;
+
+/**
+ * This class is a custom Jackson's JsonInclude validation intended to exclude from JSON
+ * an object containing only null fields and non-null but empty Lists.
+ */
+public final class JsonIncludeObjectHavingNonNullFieldOrNonEmptyList {
```
pkazlenka, line 26: Looking at the class usage, I'd rename it to something closer to class description, i.e. `ExcludeObjectsWithOnlyNullFieldsOrEmptyLists`.
I don't think we should mention `JacksonInclude` mechanism in class name if we don't implement this interface.

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/validation/JsonIncludeObjectHavingNonNullFieldOrNonEmptyList.java
```diff
@@ -0,0 +1,65 @@
...
+                        return false;
+                    }
+                } else {
+                    if (field.get(obj) != null) {
+                        return false;
+                    }
+                }
+            }
+        } catch (IllegalAccessException e) {
+            throw new IllegalStateException("Cannot determine whether to include a field because cannot access it", e);
```
pkazlenka, line 55: I'm wondering if we should fail the response in case we have problem filtering out some fields just for convenience. I mean, for me this doesn't sound like ERROR. May be warning in logs would be enough?

src-java/base-topology/base-storm-topology/src/test/java/org/openkilda/wfm/share/mappers/HaFlowHistoryMapperTest.java
```diff
@@ -452,8 +458,47 @@ public void createHaFlowDump() {
...
+                null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null,
+                null, null, null, new HaFlowPathPayload(null, null, null, null, null, null, null, null, null, null,
+                null, null, Collections.EMPTY_LIST, Collections.EMPTY_LIST), new HaFlowPathPayload(null, null, null,
+                null, null, null, null, null, null, null, null, null, Collections.EMPTY_LIST, Collections.EMPTY_LIST));
+
+        ObjectMapper objectMapper = new ObjectMapper();
+
+        String json = objectMapper.writeValueAsString(payloadWithEmptyPaths);
+        assertFalse(json.contains("protected_reverse_path"),
+                "Protected forward must not be present, because all fields are null and lists are empty");
```
pkazlenka, line 473: s/forward/reverse/

src-java/base-topology/base-storm-topology/src/test/java/org/openkilda/wfm/share/mappers/HaFlowHistoryMapperTest.java
```diff
@@ -452,8 +458,47 @@ public void createHaFlowDump() {
...
+        assertFalse(json.contains("protected_forward_path"),
+                "Protected forward must not be present, because all fields are null and lists are empty");
+    }
+
+    @Test
+    void whenDumpContainsNonEmptyListsInPathPayload_theWholePathPayloadIsIncluded() throws JsonProcessingException {
+        HaFlowDumpPayload payloadWithPaths = new HaFlowDumpPayload("1", null, null, null, null, null, null, null, null,
```
pkazlenka, line 480: I suggest to move creation of totally empty  `HaFlowDumpPayload` to the method and modify needed values in this test. 
Also, Would it play trick if you add @NoArgsConstructor for HaFlowDumpPayload class?

pkazlenka (changes requested):

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/payload/history/HaFlowDumpPayload.java
```diff
@@ -63,6 +65,10 @@ public class HaFlowDumpPayload implements Serializable {
     List<HaSubFlowPayload> haSubFlows;
     HaFlowPathPayload forwardPath;
     HaFlowPathPayload reversePath;
+    @JsonInclude(value = JsonInclude.Include.CUSTOM,
+            valueFilter = JsonIncludeObjectHavingNonNullFieldOrNonEmptyList.class)
     HaFlowPathPayload protectedForwardPath;
+    @JsonInclude(value = JsonInclude.Include.CUSTOM,
+            valueFilter = JsonIncludeObjectHavingNonNullFieldOrNonEmptyList.class)
```
IvanChupin, line 72: Just for the information to consider:
Have you thought about to move the logic of setting the values for the protectedForwardPath and protectedReversePath fields to the HaFlowHistoryMapper? I mean if we have not null, but completely empty inside objects, do we actually need to instantiate them at all? what if prevent the instantiation of such objects?
I think the best place to to it is the HaFlowHistoryMapper. or HistoryService.getHaFlowHistoryEvents() method.
As result we do need to implement JsonIncludeObjectHavingNonNullFieldOrNonEmptyList with the reflection hacks of setting the private fields accessible true.
:)

IvanChupin (approved):
Great job! 👍 
Just one moment to consider.

pkazlenka:
Tested manually. No autotests needed.

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/payload/history/HaFlowDumpPayload.java
dmitrii-beliakov: Just another argument to have work-in-progress code reviews, because I also thought and tried it first via the mapper. It doesn't work unfortunately. Even if we don't create object at all, Jackson still renders the model according to what we have in this POJO by default including all null fields. Basically, POJO defines a model with its metadata, doesn't matter what's inside.
There is a mechanism to not display null or empty fields using `JsonInclude`, but it is tricky. The collections are never null for Jackson, so if you have a nested object with a list, this object always has a member: a collection, even if it is empty. If you put the appropriate annotation on that field, it hides that field, but still consider the parent object as non-empty. So, the object is going to be displayed, but with a hidden collection. There is no built-in annotation `NON_EMPTY` but for objects, so I had to implement it.

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/validation/JsonIncludeObjectHavingNonNullFieldOrNonEmptyList.java
dmitrii-beliakov: I considered it, but then decided to leave it like it is. This exception means that the logic of include/exclude is not compatible with the given model class. Meaning either change the class, or implement different logic for include/exclude.

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/validation/JsonIncludeObjectHavingNonNullFieldOrNonEmptyList.java
dmitrii-beliakov: In fact, they made it in Jackson in a way that you don't need any interface (in my opinion, that's weird, but it is like it is). In Jackson, they have a hidden requirement to the equals method and naming/idea of include with negation (i.e. include NON_NULL), so I just followed something people could find on web. 
Frankly, I don't think some names would help here. I'm sure people would be confused anyway whether equals returns true means include or exclude. I can create an interface for it with a single usage for it or rename it,  but then on the usage side it's going to be 
```java
@JsonInclude(value = JsonInclude.Include.CUSTOM, valueFilter = ExcludeObjectsWithOnlyNullFieldsOrEmptyLists.class)
```
kind of still confusing.
So, what is it you proposing here? Maybe I can add some more text explanation

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/validation/JsonIncludeObjectHavingNonNullFieldOrNonEmptyList.java
dmitrii-beliakov: I changed the type and the message. Should be more clear now.

src-java/base-topology/base-messaging/src/main/java/org/openkilda/messaging/validation/JsonIncludeObjectHavingNonNullFieldOrNonEmptyList.java
dmitrii-beliakov: I added a javadoc with the explanation.

pkazlenka: Fair enough.

pkazlenka (approved):

niksv (approved):

PR #5491, opened 2023-11-28
Remove the deprecated authentication plugin for the MySQL container.

### LINKED ISSUE #5460: MySQL container: change deprecated default_authentication_plugin
`default_authentication_plugin=mysql_native_password` is deprecated and pollutes log files with messages about the deprecation.
Investigate whether we can change it to the new auth type. 

[Here are more information](https://dev.mysql.com/doc/refman/8.0/en/upgrading-from-previous-series.html#upgrade-caching-sha2-password).

dmitrii-beliakov:
After this fix the newly created MySQL container no longer uses the deprecated authentication plugin and uses the recommended one.
In order to test, deploy OpenKilda, find the ID of the MySQL container:
```bash
docker ps | grep mysql
```
and then display the logs of that container:
```bash
docker logs be35
```
note that there are no more warning messages in the log about the deprecated authentication plugin.

closes #5460

### DIFF
--- a/confd/templates/docker-compose/docker-compose.tmpl
+++ b/confd/templates/docker-compose/docker-compose.tmpl
@@ -127,7 +127,7 @@ services:
   mysql_db:
     hostname: mysql.pendev
     image: mysql:8.0.34
-    command: --default-authentication-plugin=mysql_native_password
+    command: --default-authentication-plugin=caching_sha2_password
     volumes:
       - ./docker/mysql/setup_data:/docker-entrypoint-initdb.d
       - sql_data:/var/lib/mysql


IvanChupin (approved):

pkazlenka (approved):

niksv (approved):

niksv:
> This change is compatible with the all-in-one, but it's not displayed here.
You can run all in one with PR id and jenkins should put its approval here

niksv:
@dmitrii-beliakov Are we sure it will work with current staging/prod DBs?

pkazlenka:
Successfully verified on a hardware environment as well.

PR #5492, opened 2023-11-28
GUI: Add inner vlans into flows/add-new page

IvanChupin:
Added inner vlan field for source and dest switch for "add new Flow page."

In order to test functionality:
Go to Flows page and press "Add new Flow" button
image
On the add new Flow page find new fields for inner vlan.
image.
we can fill in all the required fields and create a new flow.
The selected inner vlans should be presented on the flow-detail page.
image


dmitrii-beliakov (approved):

IvanChupin:
> @IvanChupin , functionality works as expected, but, IIRC, we agreed to allow special purpose VLAN 4095 to be used within others. So, please, add this option (at the moment VLAN 4094 is the highest available).

fixed

Issue #4485, opened 2021-09-17
[GUI] missing permission validation on the 'useractivity' page


andriidovhan:
**Steps to reproduce:**
1. create (or update existing user) without permission to see a list of users.
2. go to the `useractivity` page and press in the `Username` field.

**Expected result:**
It should only show the activities of logged-in user if that user does not have permission to view users.

**Actual result:**
The system is allowing to see the list of users in this(`Username`) field.

<img width="1380" alt="OPEN_KILDA_-_User_Activity" src="https://user-images.githubusercontent.com/19863573/133797906-34a0aee0-8258-4a87-bab2-52b4bba3c925.png">

PR #5495, opened 2023-11-29
open-kilda-4485 [GUI] missing permission validation on the 'useractivity' page

IvanChupin:
added permission (menu_user_management) check for user activity page

problem described here: closes: https://github.com/telstra/open-kilda/issues/4485

Steps to reproduce:
1) Create user with permission 'menu_user_activity', but without 'menu_user_management'
2) change the password for the created user:
image
3) Copy the password to the clipboard
4) login to the newly created user and go to the 'user activity page'
image
5) Result: Username selectbox  is not shown for the users who have no 'menu_user_management' permission. Also as you may see on the last screenshot the user activity data is loaded via the following get request:
`/openkilda/api/useractivity/log?userId=2&startTime...`  it has userId urlParameter that is used to pass the needed user, this parameter is not new and I decided to use it to load only activity data for loged in user. For example for users with the 'menu_user_management' permission get request looks like this:
`/openkilda/api/useractivity/log?userId=&startTime...`. So while approach that I filter the activityData is not the best(because still it is possible to load needed userActivity data just passing zero arguments, but this is another scope..), I see no alternative option to solve the task without significant code changes.

closes: #4485

### DIFF
--- a/src-gui/ui/src/app/modules/useractivity/useractivity.component.html
+++ b/src-gui/ui/src/app/modules/useractivity/useractivity.component.html
@@ -1,110 +1,116 @@
 <div id="switchdetails_div">
-  <form [formGroup]="userActivityForm" id="userActivityForm" #roleform="ngForm">
-  <div class="row form-group clearfix">
-    <div class="col-sm-6">
-      <div class="row">
-        <label class='col-sm-3 col-form-label'>Start Date:</label>
-        <div class='col-sm-6'>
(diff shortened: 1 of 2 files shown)


pkazlenka:
Verified in Chromium. Throws exception and doesn't show user activity at all in firefox, but I don't think we support all the browsers in GUI.

pkazlenka (approved):

PR #5514, opened 2023-12-14
[TEST]: 5162: Refactoring HC

yuliiamir:


src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/extension/env/EnvExtension.groovy
```diff
@@ -93,6 +106,22 @@ class EnvExtension extends AbstractGlobalExtension implements SpringContextListe
             throw new RuntimeException("Provided profile '$profile' is unknown. Select one of the following profiles:" +
                     " hardware, virtual")
         }
+        if((profile == "hardware" && enable) || (profile == "virtual" && !isTopologyRebuildRequired && enable)) {
```
pkazlenka, line 109: It's useful to move complicated conditions into boolean methods with readable name. E.g. this one can be moved into something like `boolean isHealthcheckNeeded(String profile, Boolean enable, Boolean isTopologyRebuildRequired)`.

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/extension/env/EnvExtension.groovy
```diff
@@ -93,6 +106,22 @@ class EnvExtension extends AbstractGlobalExtension implements SpringContextListe
...
+            log.info("Starting topology-related HC: lab_id=" + topology.labId)
+            Wrappers.wait(30) {
```
pkazlenka, line 111: Let's replace magical number with clear constant name.

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/extension/env/EnvExtension.groovy
```diff
@@ -93,6 +106,22 @@ class EnvExtension extends AbstractGlobalExtension implements SpringContextListe
...
+                     allLinksAreActive,
+                     noFlowsLeft,
+                     noLinkPropertiesLeft,
+                     linksBandwidthAndSpeedMatch,
+                     noExcessRulesMeters,
+                     allSwitchesConnectedToExpectedRegion,
+                     switchesConfigurationIsCorrect].eachParallel { it(topology) }
+                }
+            }
+            log.info("Topology-related HC passed: lad_id=" + topology.labId + " time: " + System.currentTimeSeconds())
```
pkazlenka, line 123: Misprint: should be `lab_id`

Also, either add start time to the line 110 or remove it from this line. Otherwise logging this doesn't make sense.
As an option, you may wrap method into `Wrappers.benchmark()` (though, you'll have to add logger parameter to it to output into console instead of logstash).

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/extension/env/EnvExtension.groovy
```diff
@@ -186,6 +215,100 @@ class EnvExtension extends AbstractGlobalExtension implements SpringContextListe
                     server42IslRtt = (sw.prop.server42IslRtt == null ? "AUTO" : (sw.prop.server42IslRtt ? "ENABLED" : "DISABLED"))
                 })
             }
+
+            log.info("Starting topology-related HC: lab_id=" + topo.labId)
```
pkazlenka, line 219: Why do you copy-paste this healthcheck instead of reusing it (if it is needed at all)?

src-java/testing/test-library/src/main/java/org/openkilda/testing/config/DefaultServiceConfig.java
```diff
@@ -131,6 +132,29 @@ public List<Floodlight> floodlights(@Value("#{'${floodlight.openflows}'.split(',
...
 
+    @Bean
+    @Scope(SCOPE_PROTOTYPE)
+    public List<Floodlight> commonFloodlights(@Value("#{'${floodlight.openflows}'.split(',')}") List<String> openflows,
+                                        @Value("#{'${floodlight.endpoints}'.split(',')}") List<String> endpoints,
+                                        @Value("#{'${floodlight.containers}'.split(',')}") List<String> containers,
+                                        @Value("#{'${floodlight.regions}'.split(',')}") List<String> regions,
+                                        @Value("#{'${floodlight.modes}'.split(',')}") List<String> modes,
+                                        @Autowired TopologyDefinition topo) {
+        if (openflows.size() != endpoints.size() || openflows.size() != containers.size()
```
pkazlenka, line 143: Move this to boolean method with self-explaining name.

src-java/testing/test-library/src/main/java/org/openkilda/testing/config/DefaultServiceConfig.java
```diff
@@ -131,6 +132,29 @@ public List<Floodlight> floodlights(@Value("#{'${floodlight.openflows}'.split(',
...
+                || openflows.size() != regions.size() || openflows.size() != modes.size()) {
+            throw new IllegalArgumentException("Floodlight test properties are illegal. Sizes of floodlight.openflows,"
+                    + " floodlight.endpoints, floodlight.containers, floodlight.regions, floodlight.modes should be "
+                    + "equal");
+        }
+        List<Floodlight> floodlights = new ArrayList<>();
```
pkazlenka, line 149: Probably this would be shorter with streams api.

pkazlenka (changes requested):

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/extension/env/EnvExtension.groovy
yuliiamir: In this case we check topology readiness per topology from topologies pool after its creation. In the case of rebuilding topology, only this part will be checked and the HC above #110 will be skipped.

pkazlenka: Can we move this to method with List<Closure> parameter to not repeat code?

src-java/testing/test-library/src/main/java/org/openkilda/testing/config/DefaultServiceConfig.java
yuliiamir: As we use the same index for the rest of the props (openflows.get(i), containers.get(i), regions.get(i), modes.get(i)) to set up each floodlight, it seems like using streams is not applicable here.

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/extension/env/EnvExtension.groovy
yuliiamir: done

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/extension/env/EnvExtension.groovy
yuliiamir: sorry, it was my debug time, I don't want to use TOPOLOGY_DISCOVERING_TIME, added the new constant for that HEALTH_CHECK_TIME.

src-java/testing/test-library/src/main/java/org/openkilda/testing/config/DefaultServiceConfig.java
pkazlenka: Hm. AI says, that it can be written as 
```
return IntStream.range(0, regions.size())
    .mapToObj(i -> new Floodlight(
        openflows.get(i),
        containers.get(i),
        regions.get(i),
        new FloodlightServiceImpl(endpoints.get(i)),
        EnumUtils.getEnumIgnoreCase(FloodlightConnectMode.class, modes.get(i))
    ))
 ```
 , but it it doesn't look more readable. 
 I'd still move each `Floodlight` object constructor parameter to the new line, but it's not mandatory in our coding convention.

pkazlenka (approved):

PR #5520, opened 2023-12-19
Add ordering for simple flow history actions

### LINKED ISSUE #5519: Remove workaround for history actions ordering with fake time
Implement for simple flow the same solution that was done for HA-flows in the issue #5501

dmitrii-beliakov:
This change removes a hack with the ordering based on the fake history action time. Previously when multiple events happened in exact same time, the history service artificially added a millisecond to 2nd and further actions, and then they were ordered by this time. Now, this hack is removed, actions are saved with the actual time of their creation and an additional ordering is added.

closes #5519


pkazlenka (approved):

src-java/kilda-persistence-hibernate/src/main/java/org/openkilda/persistence/hibernate/entities/history/HibernateFlowEvent.java
```diff
@@ -74,8 +74,14 @@ public class HibernateFlowEvent extends EntityBase implements FlowEventData {
...
 
+    /**
+     * The ordering relies on grouping messages in topologies: history messages for the same correlation ID arrive to
+     * the same worker and processed sequentially within a single operation (see fieldsGrouping in topology builders).
+     * However, timestamps are not generated by DB, so assumptions about the relation between timestamp and
+     * id cannot be made.
+     */
     @OneToMany(mappedBy = "event", cascade = CascadeType.ALL)
-    @OrderBy("event_time")
+    @OrderBy("eventTime , id")
```
IvanChupin, line 84: while it is possible to use both the column name or the field name as a value, it seems like we use snake case preferably, it might be better to stick to one style.

IvanChupin (approved):

src-java/kilda-persistence-hibernate/src/main/java/org/openkilda/persistence/hibernate/entities/history/HibernateFlowEvent.java
dmitrii-beliakov: I think its better to stick with the field name then. I chose this one because IDE can highlight typos, with the snake case it has false negatives, it underlines all fields as incorrect.

PR #5522, opened 2023-12-19
remove opentsdb and hbase

pablomuri:
part of #4925


pkazlenka (approved):

dmitrii-beliakov:
I think that we need to get rid of otsdb completely. Now there is still a Storm topology with that name, template opentsdb-topology.tmpl. I don't see tests removed, we didn't have tests for it? If some parts are still in use, they need to be renamed.

pablomuri:
> I think that we need to get rid of otsdb completely. Now there is still a Storm topology with that name, template opentsdb-topology.tmpl. I don't see tests removed, we didn't have tests for it? If some parts are still in use, they need to be renamed.

@dmitrii-beliakov I was thinking about this, but I am not sure. Consider that even if we don't have an opentsdb running, the [victoria metrics is getting the data exposing an opentsdb interface](https://github.com/telstra/open-kilda/blob/70aa4490d27b102f5a4c2dc4055d2e65235a5454/confd/templates/docker-compose/docker-compose.tmpl#L586). In fact, the topology is still sending the statistics in opentsdb format.

The functional tests were removed by @pkazlenka in a previous PR.

dmitrii-beliakov:
@pablomuri I just wanted to make sure there is no leftovers after the removal and no confusion. We can still find something like:
```
This sub-project holds the Storm topology "OpenTSDB" that is used to submit 
the statistics from OpenKilda to OpenTSDB.
```
that might be confusing. There is some config that might be needed only for OpenTSDB. If you say you updated all required places, then all is good.

dmitrii-beliakov (approved):

PR #5526, opened 2023-12-21
Add a new workflow path in rerouting FSMs to skip rerouting

### LINKED ISSUE #5291: Misleading message in the flow history when doing a reroute
There is a misleading message when doing a reroute: 
```json
"action": "Failed to reroute the flow",
"details": "Reroute is unsuccessful. Couldn't find new path(s)"
```
 but at the same time reroute could actually bring the flow from DOWN to UP. This message actually means that new paths were not found, but this might be correct and the flow goes UP using the old paths that were unavailable before, but can be used now. 

Steps to reproduce:
1. Create a flow with any parameters, make sure the flow status is UP
2. Take down all ports on one of the switches (for example by using the switch controller `/switches/00:00:00:00:00:00:00:01/port/1/config` with a body `{"status":"DOWN"}`. Make sure the flow is DOWN
3. Bring one of the ports UP, so that the flow could use it. Make sure the flow is UP again
4. Navigate to the flow's history, find the automatic reroute event:
Actual result: one of the actions says: `Failed to reroute the flow` (other action says `"action": "The flow status was reverted to UP"`)
Expected result: the message does not imply that a reroute was unsuccessful if it is not the case.

dmitrii-beliakov:
This PR introduces a new behavior when a flow or HA-flow transits from DOWN to UP state and automatic reroute is invoked. In case, when there is only only path with goes down and then up, automatic reroute cannot find a new better path, but at the same time it is not needed because the flow is working using the old paths that are now ok. In this case the reroute is not needed and skipped. In the history we can see messages about the skipped rerouting, so that we can distinguish it with the rerouting that allocates new resources. 
The message in the history that reroute is unsuccessful in this case is now removed.

closes #5291


IvanChupin (approved):

dmitrii-beliakov:
It's been some time ago, I think there was a problem because some of the tests relies on the history outcome, which has changed. Please double check the test failures.

PR #5533, opened 2024-01-10
[TEST]: 5222: Enabling test: Bulk rerouting due to ISLs blinking

yuliiamir:


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/xresilience/ChaosSpec.groovy
```diff
@@ -83,6 +105,10 @@ class ChaosSpec extends HealthCheckSpecification {
                 }.empty
             }
         }
+
+        topology.activeSwitches.each {
```
pkazlenka, line 109: I'd replace with asynchronous switch synchronization. But leave it as is, I'll cover it in my refactoring.

pkazlenka (approved):

yuliiamir:
The further investigation is required as excess rules are still present after flow deletion. It is hard to catch this issue by executing tests on VM, but this test failed 2/3 on JJ.

PR #5541, opened 2024-01-17
5390: [TEST] Replace switch validation with switch synchronization where possible pt. I

pkazlenka:
* Added methods synchronizeAndGetFixedEntries() into SwitchHelper
* Replaced validation calls with new methods in org.openkilda.functionaltests.spec.flows,
org.openkilda.functionaltests.spec.links,
org.openkilda.functionaltests.spec.server42,
org.openkilda.functionaltests.spec.switches (partially)
* Added method SwitchPair.toList()


pkazlenka:
I've almost done with the second part, so no need to review two big PRs.

PR #5543, opened 2024-01-18
3566 Add validation for remotelog server ip

oleksir:


src-java/grpc-speaker/grpc-service/src/main/java/org/openkilda/grpc/speaker/client/GrpcSession.java
```diff
@@ -290,7 +290,7 @@ public CompletableFuture<Optional<CliReply>> setConfigRemoteLogServer(RemoteLogS
         GrpcResponseObserver<CliReply> observer = new GrpcResponseObserver<>(address, SET_CONFIG_REMOTE_LOG_SERVER);
 
         RemoteLogServer logServer = RemoteLogServer.newBuilder()
-                .setIpaddr(remoteServer.getIpAddress())
+                .setIpaddr(verifyHostAddress(remoteServer.getIpAddress()))
```
pablomuri, line 293: I've noticed this is not being tested. Could you please test at least the static method?

pablomuri (changes requested):

src-java/grpc-speaker/grpc-service/src/main/java/org/openkilda/grpc/speaker/client/GrpcSession.java
oleksir: hi @pablomuri added tests

PR #5549, opened 2024-01-22
Improvement/migration java 8 17 kilda gui

IvanChupin:
Kilda-gui:
* reverted to java 8
* dockerfile logic that remove java 17 deps
* all deps kept the same
* removed old jUnit transitive deps from com.googlecode.json-simple
* Unit tests for statsController fixed
* small refactoring for FlowConverter.java,FlowPathConverter.java,IslLinkConverter.java

### DIFF
--- a/src-gui/src/main/java/org/openkilda/security/filter/ResponseWrapper.java
+++ b/src-gui/src/main/java/org/openkilda/security/filter/ResponseWrapper.java
@@ -15,14 +15,13 @@
 
 package org.openkilda.security.filter;
 
-import jakarta.servlet.ServletOutputStream;
-import jakarta.servlet.http.HttpServletResponse;
-import jakarta.servlet.http.HttpServletResponseWrapper;
-
 import java.io.IOException;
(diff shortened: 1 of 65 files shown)


src-gui/src/main/java/org/openkilda/security/filter/ResponseWrapper.java
```diff
@@ -15,14 +15,13 @@
...
-import jakarta.servlet.http.HttpServletResponse;
-import jakarta.servlet.http.HttpServletResponseWrapper;
-
 import java.io.IOException;
 import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
 import java.util.UUID;
+import javax.servlet.ServletOutputStream;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpServletResponseWrapper;
```
dmitrii-beliakov, line 24: Just curious: why is it so? In other places I had to do the opposite migration.

dmitrii-beliakov (approved):

PR #5553, opened 2024-01-24
Migrate to java 17: functional tests

pkazlenka:
* Updated functional tests to work with java 17

### DIFF
--- a/src-java/testing/test-library/src/main/java/org/openkilda/testing/config/KafkaTopicNameConfig.java
+++ b/src-java/testing/test-library/src/main/java/org/openkilda/testing/config/KafkaTopicNameConfig.java
@@ -1,4 +1,4 @@
-/* Copyright 2019 Telstra Open Source
+/* Copyright 2024 Telstra Open Source
  *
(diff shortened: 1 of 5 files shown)


src-java/testing/test-library/src/main/java/org/openkilda/testing/config/KafkaTopicNameConfig.java
```diff
@@ -37,11 +37,37 @@
...
 @Configuration
-@EnableAutoConfiguration(exclude = {ErrorMvcAutoConfiguration.class})
-@ConfigurationType(name = "kafkaTopicsConfig", value = KafkaTopicsConfig.class)
-@ConfigurationType(EnvironmentConfig.class)
 public class KafkaTopicNameConfig {
 
+    private KafkaTopicsConfig kafkaTopicsConfig;
+    private EnvironmentConfig environmentConfig;
+
+    @Bean
```
dmitrii-beliakov, line 45: @IvanChupin maybe we still need to think about the profiles or it's the task for the future? What do you think?

dmitrii-beliakov (approved):

PR #5568, opened 2024-01-31
[Java 8-17] Migrate from docker-compose to docker compose v2.

dmitrii-beliakov:
This change removes `docker-compose` which is no longer supported. After this change we can upgrade `docker` on our Ubuntu and remove `docker-compose` at all.
Some Ubuntu machines might have a warning: `WARNING: Your kernel does not support swap limit capabilities. Limitation discarded.` Which could be resolved as it is described on [the docker page](https://docs.docker.com/engine/install/troubleshoot/#kernel-cgroup-swap-limit-capabilities). I'll copy-paste it here:

Edit the `/etc/default/grub` file. Add or edit the `GRUB_CMDLINE_LINUX` line to add the following two key-value pairs:
```
GRUB_CMDLINE_LINUX="cgroup_enable=memory swapaccount=1"
```
Save and close the file.
Update the GRUB boot loader.
```
sudo update-grub
```
The changes take effect when you reboot the system.

There is another bug in the current docker that makes the output flickering. I believe there is an issue for that on docker and it will be resolved in the following versions. (It could be seen when doing `make up-stable` just before wfm installs the topologies)

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,11 +42,6 @@ The following packages are required for building OpenKilda controller:
  - GNU Make 4.1+
  - Open vSwitch 2.9+
 
-#### Python dependency notice
-
-We do not recommend upgrading pip and install docker-compose using the methods described below, bypassing the packer managers. Instead, please read the documentation for installing the [pip](https://pip.pypa.io/en/stable/installation/#upgrading-pip) and the [docker-compose](https://docs.docker.com/compose/install/).
-
-
 #### Dependency installation on Ubuntu 18.04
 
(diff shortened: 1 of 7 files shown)


pablomuri:
> There is another bug in the current docker that makes the output flickering. I believe there is an issue for that on docker and it will be resolved in the following versions. (It could be seen when doing make up-stable just before wfm installs the topologies)

Coold you please link this issue?

IvanChupin (approved):

dmitrii-beliakov:
@pablomuri 
> Coold you please link this issue?

Right now I lost the exact link and can only find an issue about docker stats flickering screen and docker desktop flickering. 
It's a minor thing anyway. We install docker without specifying the version. We'll get an updated version automatically.

PR #5571, opened 2024-02-02
Migration-java-8-17. Northbound, Grpc-speaker, test-library config fixes

IvanChupin:
Northbound config fix, unit tests fixes
GRPC-speaker config fix, unit tests fixes
test-library config fix
In this PR sabre library moved away from the following kilda modules: **Northbound**, **GRPC** and **test-library**


src-java/grpc-speaker/grpc-service/src/main/java/org/openkilda/grpc/speaker/config/KafkaTopicsGrpcConfig.java
```diff
@@ -0,0 +1,416 @@
+/* Copyright 2019 Telstra Open Source
```
dmitrii-beliakov, line 1: Not requesting a change, but just some thoughts: copyright since 2019 includes copyright since 2024 😄, but for the new files it's better to use the current year? The caveat is that for old files this new claim could be understood as it removes the claim from the previous version or the opposite, it's been claimed in 19, but the edit in 24 goes without the copyright statement. To resolve ambiguity, some companies put two years: Copyright 2019,2024, the year of the first publication and when the last edit has been made.

src-java/grpc-speaker/grpc-service/src/main/java/org/openkilda/grpc/speaker/config/KafkaTopicsGrpcConfig.java
```diff
@@ -0,0 +1,416 @@
...
+    @Value("${kafka.topic.topo.router.connected.devices.storm:kilda.connected.devices.priv.storm}")
+    private String topoConnectedDevicesTopic;
+
+    @Value("${kafka.topic.topo.floodlight.connected.devices.region:kilda.floodlight.connected.devices.priv}")
+    private String topoConnectedDevicesRegionTopic;
+
+    @Value("${kafka.topic.topo.history.storm:kilda.topo.history.storm.priv}")
+    private String topoHistoryTopic;
+
+    // TODO(surabujin): check usage
```
dmitrii-beliakov, line 162: maybe create an issue for it? Seems like nobody remembers that this TODO exists.

dmitrii-beliakov (approved):

PR #5573, opened 2024-02-02
5517 Separate docker run command

oleksir:


nikitamarchenko:
@oleksir please don't add new RUN commands in docker if possible. Every single RUN command creates an additional layer in the docker image. https://docs.docker.com/develop/develop-images/instructions/#run. tldr that bad practice don't do that.

pablomuri:
Hi @oleksir,  it was my fault, @nikitamarchenko is right. So, as he said we can't separate the RUN commands. I like your suggestion but let's discuss it tomorrow with the team.

PR #5577, opened 2024-02-06
[Java 8-17] Improvement/migration java 8 17 gradle

dmitrii-beliakov:
This change removes usages of deprecated methods, constants, classes, gradle config with their more modern alternatives. 

docker-compose here merged for convenience, there is a separate PR for that one.


src-java/build.gradle
```diff
@@ -56,28 +63,24 @@ subprojects {
...
-                version {
-                    strictly '2.3.1'
-                }
-            }
-            implementation('org.apache.kafka:kafka_2.11') {
-                version {
-                    strictly '2.3.1'
-                }
-            }
+//            implementation('org.apache.kafka:kafka-clients') {
```
dmitrii-beliakov, line 66: There is a continuation task for this. At this time different modules use different versions. I move them gradually to the new version.

PR #5588, opened 2024-02-19
[Java 8-17] Add JMX monitoring settings. Adjust Storm settings so that it doesn't consume 100% of all cores.

dmitrii-beliakov:
This PR introduces JMX remote monitoring settings. It also adjust settings for Storm so that it doesn't utilize so many CPU resources.


dmitrii-beliakov:
I tested NB JMX connection manually it works fine. 
For Storm, there is a need for some other adjustments to enable monitoring. I left these setting and the explanation, so that the next person who will try JMX won't start from scratch.

IvanChupin (approved):

PR #5589, opened 2024-02-21
GenerateTopologyConfig functional-test fix

IvanChupin:
*ClassNotFoundException fix for PostgreSQL82Dialect
update deps: com.vladmihalcea:hibernate-types
52:2.20.0 => 60:2.21.1

*UuidGenerator java.lang.NoSuchMethodException fixes for HibernatePortEvent entity.

*Fixed the problem for the create query for HibernatePortEvent entity with incorrect string value for the column Id(uuid format) of HibernatePortEvent entity.

*Fixed the problem with Hibernate update query for HibernatePortEvent entity ...could not execute statement ... Column 'time_create' cannot be null java.sql.SQLIntegrityConstraintViolationException.

* Above fixes fixed the GenerateTopologyConfig functional-test run.

### DIFF
--- a/src-java/server42/server42-control/src/main/java/org/openkilda/server42/control/kafka/KafkaConfig.java
+++ b/src-java/server42/server42-control/src/main/java/org/openkilda/server42/control/kafka/KafkaConfig.java
@@ -15,7 +15,11 @@
 
 package org.openkilda.server42.control.kafka;
 
+import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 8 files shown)


src-java/server42/server42-control/src/main/java/org/openkilda/server42/control/kafka/KafkaConfig.java
IvanChupin, line 1: @IvanChupin remove logging before merge

PR #5591, opened 2024-02-22
Remove redundant dependencies from shadowJar task and classpath.

dmitrii-beliakov:
This change removes the logger dependencies from Storm topology modules. After this change there is no longer a conflict of the logger implementation when the topology jar is uploaded to the Storm cluster.

There is still a need to shrink the jar, but this is a more tricky task, because some dependencies are needed at runtime only. I'll create a separate task for it.

### DIFF
--- a/src-java/build.gradle
+++ b/src-java/build.gradle
@@ -221,27 +221,15 @@ subprojects {
 
     tasks.withType(ShadowJar).tap {
         configureEach {
-            // The step detects a conflict in dependency jars' content and fails the build if found.
-            doFirst {
-                Map<String, String> classes = new HashMap<>()
-                getIncludedDependencies().each { jar ->
-                    if (jar.isFile()) {
-                        JarFile jf = new JarFile(jar)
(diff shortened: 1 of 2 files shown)


src-java/build.gradle
```diff
@@ -221,27 +221,15 @@ subprojects {
...
-                        jf.close()
-                    }
-                }
+            /**
+             * Topology Jars that are uploaded into a Storm cluster shouldn't have duplicate classes that are already
+             * provided in the cluster (via Storm-....jar).
+             */
+            dependencies {
+                exclude(dependency('org.slf4j:slf4j-api'))
+                exclude(dependency('org.slf4j:slf4j-api'))
```
pablomuri, line 230: this line is duplicated

pablomuri (approved):

src-java/build.gradle
dmitrii-beliakov: nice catch, thanks! I removed the extra line

IvanChupin (approved):

PR #5604, opened 2024-03-05
Clean up dependency constraints. Remove unused code.

dmitrii-beliakov:
This PR removes unnecessary dependency constraints and updates dependencies.


src-java/testing/test-library/src/main/java/org/openkilda/testing/service/traffexam/model/EndpointReport.java
```diff
@@ -38,7 +37,7 @@ public EndpointReport(String error) {
     }
 
     public EndpointReport(ReportResponse report) {
-        if (Strings.isEmpty(report.getError())) {
+        if (report.getError() == null || report.getError().isEmpty()) {
```
IvanChupin, line 40: We could also use `org.apache.commons.lang3.StringUtils.isEmpty()` check here, since we already have implementation("org.apache.commons:commons-lang3") as a dependency

IvanChupin (approved):

PR #5611, opened 2024-03-14
[TEST]: Y-Flow: Adding a new approach for Y-Flow interaction

yuliiamir:
This PR is initial and contains the main methods to interact with Y-Flow. As Y-Flow is a complex flow that contains two regular flows, required changes have been added (history-related). In the scope of this PR, only two first Specs have been modified.


src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/FlowWithSubFlowsEntityPath.groovy
```diff
@@ -0,0 +1,106 @@
...
+                            diverseGroup: subFlow.diverseGroup
+                    )
+            )
+        }
+
+        this.topologyDefinition = topologyDefinition
+    }
+
+    List<Isl> getInvolvedIsls(boolean isForward = true) {
+        subFlowPaths.collect { it.getInvolvedIsls(isForward) }.flatten().unique() as List<Isl>
```
pkazlenka, line 88: I think in this particular part we should return Set. 
In case of simple flow, returning list makes sense because we have a clear straight order of ISLs. But in case of y- and ha-flows we don't have an order behind shared part.

pkazlenka (approved):

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/FlowWithSubFlowsEntityPath.groovy
yuliiamir: `List<Isl>` was used as methods to manipulate with ISLs(set up bandwidth/cost/etc) is for this type. We can overload existing methods as a part of another PR.
As for the order, if I am not mistaken, we have the same structure as for the regular flow due to collecting ISL per subflow but then we merge each subflow's ISLs.

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/FlowHelperV2.groovy
```diff
@@ -251,6 +251,10 @@ class FlowHelperV2 {
         return (KILDA_ALLOWED_VLANS - exclusions).shuffled().first()
     }
 
+    static List<Integer> availableVlanList(List<Integer> exclusions) {
```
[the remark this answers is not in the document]
yuliiamir: There is no need to use Set as it is a range of unique numbers from 1 to 4095.

pkazlenka (approved):

PR #5615, opened 2024-03-19
[Java 8-17] Temporary fix for the PortAntiflapSpec

dmitrii-beliakov:
This change resolves the test failure. However, it is not clear what the root cause is, as well as whether the new interval still makes sense for the test cases.

Anyway, without this change there are the following series of errors:
```
Disconnecting switch [00:00:00:00:00:00:00:02(0x0) from lab_service-1.kilda_default/192.168.32.24:41782] due to IO Error: Connection reset 
```

```
Can't process speaker command ...
org.openkilda.floodlight.error.SwitchNotFoundException: Switch 00:00:00:00:00:00:00:09 was not found
```

```
ISL controller for 00:00:00:00:00:00:00:02_6 <===> 00:00:00:00:00:00:00:07_19 not found
```

```
Port 10 for switch 1 not found.
```


dmitrii-beliakov:
@pkazlenka could you please look at this issue? Is the test case still valid when the interval is increased?

IvanChupin (approved):

Issue #3770, opened 2020-10-09
system is leaving excess/missing rules after swapEnpoints operation for two inactive flow


andriidovhan:
**Step to reproduce:**
The issue is reproduced in both modes(singleTable/multiTable) 
1. create two flows;
2. make created flows inactive;
3. `swapEnpoints` for the created flow
4. validate all involved switches

**Actual relust:**
one of involved switches has excess/missing rules;

**Expected result:**
all involved switches should be valid.

**NOTE:**
this issue can be reproduced by autotest (`"Unable to swap endpoints for two flows when one of them is inactive"`)

pkazlenka:
Issue is not reproducible.

PR #5625, opened 2024-04-04
5390: [TEST]: Stabilize flaky Swap Endpoint test

izadorozhna:
Implements #5390

* The spec failed when the empty map was returned in the synchronizeAndCollectFixedDiscrepancies() method.
* However it is expected for this method to return an empty map.
* Due to the known issue #3770, it returns non-empty map in the most of the cases.
* Now the check .isEmpty() is temporary commented to stabilaze the test.

Related to #3770

### DIFF
--- a/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/SwapEndpointSpec.groovy
+++ b/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/SwapEndpointSpec.groovy
@@ -21,15 +21,15 @@ import org.openkilda.northbound.dto.v2.flows.SwapFlowPayload
 import org.openkilda.testing.model.topology.TopologyDefinition.Switch
 import org.openkilda.testing.service.traffexam.TraffExamService
 import org.openkilda.testing.tools.FlowTrafficExamBuilder
+
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.web.client.HttpClientErrorException
 import org.springframework.web.client.HttpServerErrorException
-import spock.lang.Shared
-
+import spock.lang.Ignore
 import javax.inject.Provider
-
 import static groovyx.gpars.GParsPool.withPool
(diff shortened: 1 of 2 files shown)


pkazlenka (approved):

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/SwapEndpointSpec.groovy
```diff
@@ -991,13 +994,19 @@ switches"() {
...
+        when: "Get actual sources of flow1 and flow2"
+        def actualFlow1Src = northbound.getFlow(flow1.id).source
+        def actualFlow2Src = northbound.getFlow(flow2.id).source
+
+        then: "Actual flow1, flow2 sources are different"
+        Set actualFlow1Set = [actualFlow1Src.datapath, actualFlow1Src.portNumber,
+                              actualFlow1Src.vlanId, actualFlow1Src.innerVlanId].toSet()
+        Set actualFlow2Set = [actualFlow2Src.datapath, actualFlow2Src.portNumber,
+                              actualFlow2Src.vlanId, actualFlow2Src.innerVlanId].toSet()
+        assert !(actualFlow1Set == actualFlow2Set)
```
yuliiamir, line 1006: you can simplify the check:
`assert actualFlow1Set != actualFlow2Set`
Do we really need to collect actualFlow1Set/ actualFlow2Set data separately? What is the purpose of this collecting and following check?

izadorozhna: Thank you! Changed to `assert actualFlow1Set != actualFlow2Set`.
I am collecting the actual data of the flows to verify that after the failed attempt to swap the endpoints, the flow sources are not the same due to the bug #5635. Currently, each time the flows have the same source due to the effect which is not expected.

oleksir (approved):

yuliiamir (approved):

PR #5630, opened 2024-04-09
Prevent error messages in test java 8-17

oleksir:


src-java/testing/functional-tests/build.gradle
```diff
@@ -36,6 +36,7 @@ dependencies {
 
     testImplementation("org.assertj:assertj-core:3.24.2")
     testImplementation("org.apache.logging.log4j:log4j-api:2.23.0")
+    testImplementation("org.apache.logging.log4j:log4j-core:2.23.0")
```
dmitrii-beliakov, line 39: I bet runtime only is enough, but the question is whether it is needed at all. Logs are collected differently, without this dependency.

src-java/blue-green/src/main/java/org/openkilda/bluegreen/ZkClient.java
```diff
@@ -86,7 +86,7 @@ public synchronized void initAndWaitConnection() {
                 .onRetry(e -> {
                     String message = format("Failed to init zk client, retrying... Attempt: %d", e.getAttemptCount());
                     if (e.getAttemptCount() <= 10) {
-                        log.info(message, e.getLastException());
+                        log.info(format("%s. Error: %s", message, e.getLastException().getMessage()));
```
IvanChupin, line 89: It might be better to use already existed "{}" placeholders for Slf4j here, instead of using String.format placeholders..

```suggestion
                        log.info("{}. Error: {}", message, e.getLastException().getMessage());
```

IvanChupin (approved):
Please, consider to change the placeholders format.

src-java/testing/functional-tests/build.gradle
oleksir: fixed

Issue #5608, opened 2024-03-11
'Couldn't find non overlapping protected path' when restore broken ISL on protected path.


pkazlenka:
Currently our automated test "ProtectedPathSpec.Unable to swap paths for an inactive flow()" fails each time when we use switches 8 and 9 as switch pair (for some reason, the rest works ok).
At some moment of test (after the step `when: "Restore ISL for the protected path"`) both protected and main paths are up, but the flow stays in `Degraded` state with error : `Couldn't find non overlapping protected path. Skipped creating it`:
![image](https://github.com/telstra/open-kilda/assets/44895713/6441e086-5d11-4142-ae02-6d6166e750c4)
Flow history entries:
```
{
    "clazz": "org.openkilda.messaging.payload.history.FlowHistoryEntry",
    "flow_id": "11Mar085334_860_cloves0146",
    "timestamp": 1710146451,
    "timestamp_iso": "2024-03-11T08:40:51.324Z",
    "actor": "AUTO",
    "action": "Flow rerouting",
    "task_id": "88bc4925-e861-40f4-98ae-41cf39788b44 : W : 00:00:00:00:00:00:00:08_16 : DM : 00:00:00:00:00:00:00:08_16 : 8208bb89-97f6-47c2-b662-6f71b92fd81a : 11Mar085334_860_cloves0146",
    "details": "Reason: ISL 00:00:00:00:00:00:00:03_6 <===> 00:00:00:00:00:00:00:08_16 status become ACTIVE",
    "payload": [
      {
        "timestamp": 1710146451,
        "action": "Flow rerouting operation has been started.",
        "details": null,
        "timestamp_iso": "2024-03-11T08:40:51.324Z"
      },
      {
        "timestamp": 1710146451,
        "action": "The flow has been validated successfully",
        "details": null,
        "timestamp_iso": "2024-03-11T08:40:51.341Z"
      },
      {
        "timestamp": 1710146451,
        "action": "Couldn't find non overlapping protected path. Skipped creating it",
        "details": null,
        "timestamp_iso": "2024-03-11T08:40:51.394Z"
      },
      {
        "timestamp": 1710146451,
        "action": "The flow status was set to DEGRADED",
        "details": null,
        "timestamp_iso": "2024-03-11T08:40:51.407Z"
      },
      {
        "timestamp": 1710146451,
        "action": "Failed to reroute the flow",
        "details": "Couldn't find non overlapping protected path",
[...]

PR #5648, opened 2024-04-25
Fixed the flaky test in the ProtectedPathSpec

izadorozhna:
Implements #5390
Related to #5608

* Fixed the test "Flow swaps to protected path when main path gets broken, becomes DEGRADED if protected path is unable to reroute(no bw)"
* Earlier in some cases when the switchPair was set to 2-3 or 8-9, the protected path had only 1 ISL and it had enough BW, so the test failed because the BW was not reduced for some protected path ISLs.
* Now this test passes even when the swPair is 2-3 or 8-9, and other switches. So the temporary fix to skip 8-9 switches is removed.
* Removed the workaround to skip 8-9 switches from 2 tests (valid when #5645 is merged)

### DIFF
--- a/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/ProtectedPathSpec.groovy
+++ b/src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/ProtectedPathSpec.groovy
@@ -373,13 +373,7 @@ class ProtectedPathSpec extends HealthCheckSpecification {
     @Tags([ISL_RECOVER_ON_FAIL, ISL_PROPS_DB_RESET])
     def "Flow swaps to protected path when main path gets broken, becomes DEGRADED if protected path is unable to reroute(no bw)"() {
         given: "Two switches with 2 diverse paths at least"
-        //def switchPair = switchPairs.all().withAtLeastNNonOverlappingPaths(2).random()
-        //https://github.com/telstra/open-kilda/issues/5608
-        def switchesWhere5608IsReproducible = topology.activeSwitches.findAll {it.dpId.toString().endsWith("08")
-        ||it.dpId.toString().endsWith("09")}
-        def switchPair = switchPairs.all()
-                .excludeSwitches(switchesWhere5608IsReproducible)
-                .withAtLeastNNonOverlappingPaths(2).random()
+        def switchPair = switchPairs.all().withAtLeastNNonOverlappingPaths(2).random()
 
         when: "Create flow with protected path"
         def flow = flowHelperV2.randomFlow(switchPair).tap { allocateProtectedPath = true }
@@ -387,7 +381,7 @@ class ProtectedPathSpec extends HealthCheckSpecification {
         def path = northbound.getFlowPath(flow.flowId)
 
         and: "Other paths have not enough bandwidth to host the flow in case of reroute"
-        def otherIsls = switchPair.paths.findAll { it != pathHelper.convert(path.protectedPath) &&
+        def otherIsls = switchPair.paths.findAll {
                 it != pathHelper.convert(path) }.collectMany { pathHelper.getInvolvedIsls(it) }
                 .unique { a, b -> a == b || a == b.reversed ? 0 : 1 }
         otherIsls.collectMany{[it, it.reversed]}.each {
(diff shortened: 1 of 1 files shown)


izadorozhna:
Please see my comments and screenshots under #5608 issue

izadorozhna:
Work in progress. Will try to fix the other tests in the same Spec

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/ProtectedPathSpec.groovy
```diff
@@ -373,21 +373,15 @@ class ProtectedPathSpec extends HealthCheckSpecification {
...
     def "Flow swaps to protected path when main path gets broken, becomes DEGRADED if protected path is unable to reroute(no bw)"() {
         given: "Two switches with 2 diverse paths at least"
-        //def switchPair = switchPairs.all().withAtLeastNNonOverlappingPaths(2).random()
-        //https://github.com/telstra/open-kilda/issues/5608
-        def switchesWhere5608IsReproducible = topology.activeSwitches.findAll {it.dpId.toString().endsWith("08")
-        ||it.dpId.toString().endsWith("09")}
-        def switchPair = switchPairs.all()
-                .excludeSwitches(switchesWhere5608IsReproducible)
-                .withAtLeastNNonOverlappingPaths(2).random()
+        def switchPair = switchPairs.all().withAtLeastNNonOverlappingPaths(2).random()
```
pkazlenka, line 376: What exactly is the fix? https://github.com/telstra/open-kilda/issues/5608 is still open and not solved, as far as I can see.

pkazlenka (changes requested):

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/ProtectedPathSpec.groovy
izadorozhna: Yes, let me add some details to explain my PR.

So, I was debugging a couple of tests (namely, `"Flow swaps to protected path when main path gets broken, becomes DEGRADED if protected path is unable to reroute(no bw)"` and `"Flow swaps to protected path when main path gets broken, becomes DEGRADED if protected path is unable to reroute(no path)"` because they failed sometimes for the past month. The failure was related to the same `Couldn't find non overlapping protected path. Skipped creating it` error message mentioned in #5608. So, even with the WA to skip 8-9 switches, the overlapping issue was still reproducible.

* `Flow swaps to protected path when main path gets broken, becomes DEGRADED if protected path is unable to reroute(no bw)` - here the issue was still reproducible between the switches 2-3, please see the screenshot below.
image
When I investigated this test, it turned out that the protected path ISL has enough bw, that is why the test cannot get expected "Not enough bandwidth or no path found" message.  So the fix in this PR is done to reduce the BW on the original protected path ISLs. So, with this fix, the test passes because the flow is degraded due to the expected "Not enough bandwidth or no path found" message. So, now no need to skip 8-9 switches since with the correct BW reducing, the test passes.
* `Flow swaps to protected path when main path gets broken, becomes DEGRADED if protected path is unable to reroute(no path)` - this test also failed sometimes due to Overlapping issue - even with the WA to skip 8-9 switches. I made the PR #5645 to fix this test because the ISLs were not broken correctly there. But now I see that with this fix from #5645, the WA to skip 8-9 can be removed, it is not reproduced anymore.

So, I have executed both these tests 20 times with just 8-9 switchpair, 2-3 switchpair, random switchpair and the overlapping issue is not reproduced anymore - the tests are passing. Thus, there is no need to exclude 8-9 switches from the switchpair anymore.

pkazlenka (approved):

yuliiamir (approved):

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/ProtectedPathSpec.groovy
```diff
@@ -373,21 +373,15 @@ class ProtectedPathSpec extends HealthCheckSpecification {
...
 
         when: "Create flow with protected path"
         def flow = flowHelperV2.randomFlow(switchPair).tap { allocateProtectedPath = true }
         flowHelperV2.addFlow(flow)
         def path = northbound.getFlowPath(flow.flowId)
 
         and: "Other paths have not enough bandwidth to host the flow in case of reroute"
-        def otherIsls = switchPair.paths.findAll { it != pathHelper.convert(path.protectedPath) &&
```
yuliiamir, line 390: when we removed path.protectedPath from exclusion, we set up bandwidth less that required (protected path), but after isl break, our new main path is UP (very strange behaviour). Need to discuss.

izadorozhna:
Closing since the fix was not correct. It was not correct to break the used ISLs in protected or main path.

PR #5649, opened 2024-04-29
Updated liquibase for db_mysql_migration container.

IvanChupin:
from version 4.4 to 4.27


dmitrii-beliakov (approved):

oleksir:
Please check if we need to align with other places.

For example docker/db-migration/Dockerfile use 3.5.5

PR #5657, opened 2024-05-09
GUI fix for gui switch-datatable and switch-detail components

IvanChupin:
This fix  change the logic of showing the orange triangle that shows the discrepancies for switch name that was received from controller(Northbound API) and from inventory service. 

With the new logic, we do not have the triangle. Now on the switch-detail page we have two parts that represents the switch details data from **northbound** and switch details data from **inventory** service(switch-store).

Before the fix:
image

After the fix:
Example of switch-detail page where switch exist only in Controller.
image
Example of switch-detail page where switch exist only in Inventory.
image

Example of switch-detail page where switch exist in controller and in inventory, but with some discrepancies:
image

Example of switch-detail page, where switch(with switch_id=A) exist in controller and multiple switches with the same switch_id exist in inventory service(not usual behaviour, but possible)
image

Also it should not show any discrepancy tab in case if we do not have inventory-service connected(setup) to Kilda-gui. 

Discrepancy tab shows only the information about discrepancy in switch-state, as it was in an old logic.

In order to test this, you need to use rigel. 
I verified it on my local vm slightly tweaking  the needed part of code. But for Rigel we do not need this tweak.   

Possible test cases:
Switch A exists only in Controller.
Switch A exists only in Inventory service.
Switch A exists both in Controller and Inventory service.
Switch A exists both in Controller and Inventory service with different state.

closes https://github.com/telstra/open-kilda/issues/5661

### DIFF
--- a/src-gui/src/main/java/org/openkilda/config/DatabaseConfigurator.java
+++ b/src-gui/src/main/java/org/openkilda/config/DatabaseConfigurator.java
@@ -19,7 +19,7 @@ import org.openkilda.dao.entity.VersionEntity;
 import org.openkilda.dao.repository.VersionRepository;
 
 import com.ibatis.common.jdbc.ScriptRunner;
-
+import org.apache.commons.collections4.CollectionUtils;
 import org.apache.log4j.Logger;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.core.io.Resource;
@@ -32,9 +32,7 @@ import org.springframework.transaction.annotation.Transactional;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
-
 import java.math.BigInteger;
-
 import java.sql.Connection;
 import java.sql.Timestamp;
 import java.util.ArrayList;
@@ -42,7 +40,6 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
-
 import javax.persistence.EntityManager;
 import javax.persistence.PersistenceContext;
 import javax.sql.DataSource;
@@ -66,7 +63,7 @@ public class DatabaseConfigurator {
     private final DataSource dataSource;
 
     public DatabaseConfigurator(@Autowired final VersionRepository versionRepository, final DataSource dataSource,
-            final ResourceLoader resourceLoader, EntityManager em) {
+                                final ResourceLoader resourceLoader, EntityManager em) {
         this.versionEntityRepository = versionRepository;
         this.dataSource = dataSource;
         this.resourceLoader = resourceLoader;
@@ -82,7 +79,7 @@ public class DatabaseConfigurator {
     private void loadInitialData() {
         List<Long> versionNumberList = versionEntityRepository.findAllVersionNumber();
 
-        if (versionNumberList.isEmpty()) {
+        if (CollectionUtils.isEmpty(versionNumberList)) {
             try {
 
                 List<VersionEntity> list = new ArrayList<VersionEntity>();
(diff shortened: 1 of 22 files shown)


oleksir (approved):

src-gui/src/main/java/org/openkilda/config/DatabaseConfigurator.java
```diff
@@ -81,15 +81,15 @@ public void init() {
 
     private void loadInitialData() {
         List<Long> versionNumberList = versionEntityRepository.findAllVersionNumber();
-        
-        if (versionNumberList.size() == 0) {
+
+        if (versionNumberList.isEmpty()) {
```
oleksir, line 85: please consider to use CollectionUtils.isEmpty()

src-gui/src/main/java/org/openkilda/controller/SwitchController.java
```diff
@@ -277,8 +273,8 @@ public class SwitchController {
      * Configure switch port.
      *
      * @param configuration the configuration
-     * @param switchId the switch id
-     * @param port the port
+     * @param switchId      the switch id
```
oleksir, line 276: not sure if "@param switchId the switch id" make any sense. it is not your code, but probably would be better to remove obvious statements

oleksir (approved):

src-gui/src/main/java/org/openkilda/controller/SwitchController.java
IvanChupin: thanks, fixed

pablomuri (approved):

yuliiamir:
The following changes have been verified: 
1. A switch details page when the switch exists on both Controller and Inventory sides => Both` CONTROLLER DETAILS` and `INVENTORY DETAILS `sections are displayed + _optional: Discrepancy tab(states mismatch)_.
2. A switch details page when the switch exists ONLY on the Controller side => Only the `“CONTROLLER DETAILS”` section is displayed + **Discrepancy tab** `“*Note: Switch doesn’t exist in the inventory”.`
3. A switch details page when the switch exists ONLY on the Inventory side => Only the `“INVENTORY DETAILS” `section is displayed + **Discrepancy tab**` “*Note: Switch doesn’t exist in the controller”`.
4. Switches with the same ID in upper/lower case Inventory/Controller are treated as the same object.
5. Switches with the same ID on the Inventory side are treated as different entities => There is a discrepancy on the Switch details page: Switch ID field: `‘There is another switch with the same switch ID in the inventory service.’`
6. In the case of opening the `NON-EXISTING` switch details page, the redirection to the `‘Switches’ `page is executed.
7. Pop can be updated on the Controller side.

yuliiamir:
The following changes have been verified: 
1. A switch details page when the switch exists on both Controller and Inventory sides => Both` CONTROLLER DETAILS` and `INVENTORY DETAILS `sections are displayed + _optional: Discrepancy tab(states mismatch)_.
2. A switch details page when the switch exists ONLY on the Controller side => Only the `“CONTROLLER DETAILS”` section is displayed + **Discrepancy tab** `“*Note: Switch doesn’t exist in the inventory”.`
3. A switch details page when the switch exists ONLY on the Inventory side => Only the `“INVENTORY DETAILS” `section is displayed + **Discrepancy tab**` “*Note: Switch doesn’t exist in the controller”`.
4. Switches with the same ID in upper/lower case Inventory/Controller are treated as the same object.
5. Switches with the same ID on the Inventory side are treated as different entities => There is a discrepancy on the Switch details page: Switch ID field: `‘There is another switch with the same switch ID in the inventory service.’`
6. In the case of opening the `NON-EXISTING` switch details page, the redirection to the `‘Switches’ `page is executed.
7. Pop can be updated on the Controller side.

PR #5662, opened 2024-05-14
introduce simple commands for the build

oleksir:


oleksir:
Would be nice to change it a bit and not redeploy external services (like kafka and kibana) each time

Issue #5560, opened 2024-01-29
Add automatic cleanup in functional test


pkazlenka:
It would be nice to create a mechanism which avoid using spock native cleanup feature.

PR #5672, opened 2024-05-27
5560: [TEST]: Eliminate cleanup sections pt.4 (the rest)

pkazlenka:
Implements #5560

* Removed cleanup actions which weren't removed before
* Some cleanups for really complicated cases were left intact
* Some non-repeating cleanup actions weren't hid from specs to methods.


src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/ASwitchFlows.groovy
```diff
@@ -0,0 +1,39 @@
...
+@Component
+@Scope(SCOPE_PROTOTYPE)
+class ASwitchFlows {
+    @Autowired
+    LockKeeperService lockKeeperService
+    @Autowired
+    CleanupManager cleanupManager
+
+    void addFlows(List<ASwitchFlow> flows) {
+        cleanupManager.addAction(RESTORE_A_SWITCH_FLOWS, {lockKeeperService.removeFlows(flows)})
```
yuliiamir, line 26: WDYT: do we need to add some clarification of our topology.isls recovering (without a-switch flow we can have some other isls that will be auto-discovered by Kilda)?

yuliiamir (approved):

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/ASwitchFlows.groovy
pkazlenka: At the moment, in practice we only restore a-switch-related ISLs at this step and these ISLs return into expected state while other post-test actions happen (probably due to DoCleanupListener). If it would start to break, then yes, we'll need to add here some action to wait for ISL to appear/disappear.
But this doesn't seem like a trivial task for me.

src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/links/LinkSpec.groovy
```diff
@@ -776,18 +725,6 @@ class LinkSpec extends HealthCheckSpecification {
             !it.prop || (it.prop.server42IslRtt == "DISABLED")
         }
         switchHelper.validateAndCollectFoundDiscrepancies(switchesNotAffectedBy3906*.getDpId()).isEmpty()
-
-        cleanup:
-        aSwitchForwardRuleIsDeleted && lockKeeper.addFlows([isl.aswitch])
-        aSwitchReverseRuleIsDeleted && lockKeeper.addFlows([isl.aswitch.reversed])
-        Wrappers.wait(RULES_INSTALLATION_TIME + discoveryInterval + WAIT_OFFSET) {
```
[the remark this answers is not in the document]
pkazlenka: No. Either DoCleanupListener masks the problem, or just is not necessary and was added to the tests 'just in case'.

PR #5695, opened 2024-06-25
Fix deserializer for AddFlow

oleksir:


src-java/server42/server42-control-messaging/src/main/java/org/openkilda/server42/control/messaging/flowrtt/AddFlow.java
```diff
@@ -43,5 +43,6 @@ public class AddFlow extends Message {
     FlowDirection direction;
     @JsonProperty("port")
     Integer port;
+    @JsonProperty("dst_mac")
```
pablomuri, line 46: Theoretically, this is done by line 31 @JsonNaming(value = SnakeCaseStrategy.class) and it was working on the develop branch. Do you have any idea why is not working more? maybe we can remove it.

pablomuri (approved):
Theorically, thi

src-java/server42/server42-control-messaging/src/main/java/org/openkilda/server42/control/messaging/flowrtt/AddFlow.java
oleksir: Yes, it was tricky. Builder is used as a deserializer and @JsonNaming doesn't propagate there, but @JsonProperty does. Proved by delombok code and after applying the error gone

PR #5696, opened 2024-06-26
[TEST]: Regular Flow: New interaction approach(switches)

yuliiamir:


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/switches/FlowRulesSpec.groovy
```diff
@@ -776,24 +774,7 @@ class FlowRulesSpec extends HealthCheckSpecification {
         return rules
     }
 
-    void checkTrafficCountersInRules(FlowEndpointV2 flowEndpoint, isTrafficThroughRuleExpected) {
```
yuliiamir, line 779: This method was used to check the traffic counter for the test with singleTable mode (the test was deleted as SingleTable is deprecated, but without this leftover).

izadorozhna (approved):

PR #5706, opened 2024-07-11
[TEST]: Improvement: Fix test discrepancies

yuliiamir:


src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/PortAntiflapHelper.groovy
```diff
@@ -22,6 +25,7 @@ import static org.openkilda.testing.Constants.WAIT_OFFSET
  * port was brought down and forces sleep for required 'cooldown' amount of time when one wants to bring that port 'up'.
  */
 @Component
+@Scope(SCOPE_PROTOTYPE)
```
IvanChupin, line 28: Is that scope on purpose? or because other helpers have scope prototype?

IvanChupin (approved):

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/PortAntiflapHelper.groovy
yuliiamir: to avoid errors during parallel execution

PR #5713, opened 2024-07-19
[TEST]: Regular Flow: Performance tests: Eliminating flowHelper(v1/v2) usage

yuliiamir:
Currently, performance tests are out of testing scope. Due to the new implementation of flow interaction, the required changes have been added. The majority of the tests are passed, but there is some flakiness (EnduranceV2Spec)) that requires further investigation.


izadorozhna (approved):

src-java/testing/performance-tests/src/test/groovy/org/openkilda/performancetests/BaseSpecification.groovy
```diff
@@ -86,13 +80,21 @@ class BaseSpecification extends Specification {
     @Value('${perf.debug}')
     boolean debug
 
+    //CleanupManager is not called during perf-tests execution due to the implementation
+    //to use automatic cleanup the appropriate listeners should be added
+    static ThreadLocal<CleanupManager> threadLocalCleanupManager = new ThreadLocal<>()
```
yuliiamir, line 85: Listeners(AbstractSpringListener, AutomaticCleanupListener, RegisterListenersExtension) have not been added to avoid auto-cleanup evocation. It was done intentionally (some tests purge topology in the scope of cleanup).  It helps us to avoid unnecessary API calls(delete flow+check history).

src-java/testing/performance-tests/src/test/groovy/org/openkilda/performancetests/BaseSpecification.groovy
```diff
@@ -125,4 +127,19 @@ class BaseSpecification extends Specification {
...
         //this can have implementation if required
     }
+
+    Switch pickRandom(List<Switch> switches) {
+        switches[new Random().nextInt(switches.size())]
+    }
+
+    void deleteFlows( List<FlowExtended> flows) {
+        flows.each { it.sendDeleteRequest() }
+        def waitTime = northbound.getAllFlows().size()
```
IvanChupin, line 137: waitTime is equal to flow size.
just in case, is it correct?

IvanChupin (approved):

src-java/testing/performance-tests/src/test/groovy/org/openkilda/performancetests/BaseSpecification.groovy
yuliiamir: yes, this logic was used before, but I modified it to wait for 'existing flows' seconds. This is mostly for the case when we have 1000 or 300+ flows to decrease the time of waiting in case of any errors (anyway we have topology purge at the beginning of each spec).

PR #5716, opened 2024-07-24
Add split kafka message support for flow validation process

IvanChupin:
This PR adds the split message mechanism for the flow validation process. 

In particular during the flow validation procedure  the entities(FlowDumpResponse) that we gather from the floodlight side needs to be splitted before sending to the underlying processing topologies via Kafka. 
How it goes:
1)Floodlight module sends the chunked(splitted) messages via Kafka topic towards the floodlight-router topology, 
then 
2)floodlight-router topology just proxies all the messages and sends it via the kafka topic to the flowHs topology.
3) The flowHs topology waits and gathers chunked messages and then does some processing according to the flow validation logic and then sends some results to the northbound modules. (not in chunks, since at this step we do not work with chunks anymore, we have other entities)

So in short the chunked messages starts from floodlight module, then floodlight-router topology, then ends up on the flowHs topology.
So the changes that I made in this PR only related to the described message path above. 

Testing:
I did not implemented any func tests, so need to discuss with @yuliiamir. 
But all-in-one passed. I would say this is quite serious change that requires some attention from the func test point of view.

closes: #5718

### DIFF
--- a/src-java/floodlight-service/floodlight-api/src/main/java/org/openkilda/floodlight/api/response/ChunkedSpeakerDataResponse.java
+++ b/src-java/floodlight-service/floodlight-api/src/main/java/org/openkilda/floodlight/api/response/ChunkedSpeakerDataResponse.java
@@ -1,0 +1,76 @@
+/* Copyright 2024 Telstra Open Source
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+package org.openkilda.floodlight.api.response;
+
+import org.openkilda.messaging.MessageContext;
+import org.openkilda.messaging.MessageData;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.ToString;
+import org.apache.commons.collections4.CollectionUtils;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+@Getter
+@ToString(callSuper = true)
+@EqualsAndHashCode(callSuper = true)
+public class ChunkedSpeakerDataResponse extends SpeakerDataResponse {
+
+    @JsonProperty("message_id")
+    private String messageId;
+
+    @JsonProperty("total_messages")
+    private int totalMessages;
+
+    public ChunkedSpeakerDataResponse(@JsonProperty("data") MessageData data,
+                                      @JsonProperty("message_context") @NonNull MessageContext messageContext,
+                                      @JsonProperty("total_messages") int totalMessages) {
+        super(messageContext, data);
+        this.messageId = messageContext.getCorrelationId();
+        this.totalMessages = totalMessages;
+    }
+
+    ChunkedSpeakerDataResponse(MessageData data, MessageContext messageContext, int totalMessages, int messageIndex) {
+        this(data, messageContext, totalMessages);
+        this.messageId = String.join(" : ", String.valueOf(messageIndex),
+                messageContext.getCorrelationId());
+    }
+
+    /**
+     * Creates list of ChunkedInfoMessages from list of InfoData.
+     */
+    public static List<ChunkedSpeakerDataResponse> createChunkedList(
+            Collection<? extends MessageData> dataCollection, MessageContext messageContext) {
+
+        if (CollectionUtils.isEmpty(dataCollection)) {
+            return Collections.singletonList(new ChunkedSpeakerDataResponse(null, messageContext, 0));
+        }
+
+        List<ChunkedSpeakerDataResponse> result = new ArrayList<>();
+        int i = 0;
+        for (MessageData messageData : dataCollection) {
+            result.add(new ChunkedSpeakerDataResponse(
+                    messageData, messageContext, dataCollection.size(), i++));
+        }
+        return result;
+    }
+}
--- a/src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/service/kafka/KafkaProducerService.java
+++ b/src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/service/kafka/KafkaProducerService.java
@@ -19,10 +19,12 @@ import static org.openkilda.floodlight.service.zookeeper.ZooKeeperService.ZK_COM
 
 import org.openkilda.bluegreen.LifecycleEvent;
 import org.openkilda.bluegreen.Signal;
+import org.openkilda.floodlight.api.response.ChunkedSpeakerDataResponse;
 import org.openkilda.floodlight.service.zookeeper.ZooKeeperEventObserver;
 import org.openkilda.floodlight.service.zookeeper.ZooKeeperService;
 import org.openkilda.messaging.AbstractMessage;
 import org.openkilda.messaging.Message;
+import org.openkilda.messaging.MessageContext;
 import org.openkilda.messaging.info.ChunkedInfoMessage;
 import org.openkilda.messaging.info.InfoData;
 import org.openkilda.messaging.info.InfoMessage;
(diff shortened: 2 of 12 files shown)


src-java/floodlight-service/floodlight-api/src/main/java/org/openkilda/floodlight/api/response/ChunkedSpeakerDataResponse.java
```diff
@@ -0,0 +1,74 @@
...
+    @JsonProperty("message_id")
+    private String messageId;
+
+    @JsonProperty("total_messages")
+    private int totalMessages;
+
+    public ChunkedSpeakerDataResponse(@JsonProperty("data") MessageData data,
+                                      @JsonProperty("message_context") @NonNull MessageContext messageContext,
+                                      @JsonProperty("total_messages") int totalMessages) {
+        super(messageContext, data);
```
oleksir, line 45: minor: I'd suggest to keep the same order of arguments in parent and child classes (and methods)

src-java/floodlight-service/floodlight-api/src/main/java/org/openkilda/floodlight/api/response/ChunkedSpeakerDataResponse.java
```diff
@@ -0,0 +1,74 @@
...
+        this.messageId = messageContext.getCorrelationId();
+        this.totalMessages = totalMessages;
+    }
+
+    ChunkedSpeakerDataResponse(MessageData data, int messageIndex, int totalMessages, MessageContext messageContext) {
+        super(messageContext, data);
```
oleksir, line 51: please use this() instead of super

src-java/floodlight-service/floodlight-api/src/main/java/org/openkilda/floodlight/api/response/ChunkedSpeakerDataResponse.java
```diff
@@ -0,0 +1,74 @@
...
+        this.totalMessages = totalMessages;
+    }
+
+    /**
+     * Creates list of ChunkedInfoMessages from list of InfoData.
+     */
+    public static List<ChunkedSpeakerDataResponse> createChunkedList(
+            Collection<? extends MessageData> dataCollection, MessageContext messageContext) {
+        List<ChunkedSpeakerDataResponse> result = new ArrayList<>();
+        if (dataCollection == null || dataCollection.isEmpty()) {
```
oleksir, line 63: please use CollectionUtils.isEmpty(dataCollection)

src-java/floodlight-service/floodlight-api/src/main/java/org/openkilda/floodlight/api/response/ChunkedSpeakerDataResponse.java
```diff
@@ -0,0 +1,74 @@
...
+            result.add(new ChunkedSpeakerDataResponse(null, messageContext, 0));
```
oleksir, line 64: consider to use return List.of() in if block and it will allow to flat the code by removing else, just go forward with normal code flow

src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/service/kafka/KafkaProducerService.java
```diff
@@ -87,6 +89,14 @@ public void sendChunkedMessageAndTrack(String topic, String key, Collection<? ex
         }
     }
 
+    @Override
+    public void sendChunkedSpeakerDataAndTrack(String topic, MessageContext messageContext,
+                                               Collection<? extends InfoData> data) {
+        for (AbstractMessage message : ChunkedSpeakerDataResponse.createChunkedList(data, messageContext)) {
+            sendMessageAndTrack(topic, messageContext.getCorrelationId(), message);
```
oleksir, line 96: what if sending will fail in the middle of the loop? probably would be good to have some log message at least?

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/service/SpeakerWorkerForDumpsService.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    private void completeChunkedResponse(String key, List<ChunkedSpeakerDataResponse> messages)
+            throws PipelineException {
+
+        CommandMessage pending = keyToRequest.remove(key);
+        messagesChains.remove(key);
+        chunkedMessageIdsPerRequest.remove(key);
+
+        if (pending != null && pending.getData() != null) {
```
oleksir, line 121: please consider
if (pending == null || pending.getData() == null) {
   return;
)

normal code flow

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/service/SpeakerWorkerForDumpsService.java
```diff
@@ -0,0 +1,146 @@
...
+            if (pending.getData() instanceof DumpRulesForFlowHsRequest) {
+                List<FlowDumpResponse> responses = messages.stream()
```
oleksir, line 123: since code is very simular for each type, perhaps the common method can be used with lambda as an argument

oleksir (changes requested):

src-java/floodlight-service/floodlight-api/src/main/java/org/openkilda/floodlight/api/response/ChunkedSpeakerDataResponse.java
IvanChupin: fixed

src-java/floodlight-service/floodlight-api/src/main/java/org/openkilda/floodlight/api/response/ChunkedSpeakerDataResponse.java
IvanChupin: done

src-java/floodlight-service/floodlight-modules/src/main/java/org/openkilda/floodlight/service/kafka/KafkaProducerService.java
IvanChupin: We already have logging  in SendStatusCallback class.
So if anything fails in the middle of the loop, it will log from the KafkaProducerService side.

Also, it will fail the whole operation on the final consumer side as well. since we have mechanism that counts the  received messages

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/service/SpeakerWorkerForDumpsService.java
IvanChupin: well noted👍🏼
Fixed

src-java/flowhs-topology/flowhs-storm-topology/src/main/java/org/openkilda/wfm/topology/flowhs/service/SpeakerWorkerForDumpsService.java
IvanChupin: well noted,
Fixed

src-java/floodlight-service/floodlight-api/src/main/java/org/openkilda/floodlight/api/response/ChunkedSpeakerDataResponse.java
oleksir: this(data, messageContext, totalMessages);
this.messageId = String.join(" : ", String.valueOf(messageIndex),
                messageContext.getCorrelationId());

yuliiamir:
The added logic has been tested and verified that a message with a size of more than 1000KB is automatically split by the system in the scope of flow validation. During testing 400 flows were created on one switch and the switch rules response size was 1135KB. After that flow validation was executed successfully:
**30Jul160949_324_celery6500_flow** 
`cor id: gghh-kiiuy-888-yuliia-kkk-mir123`

```
[
  {
    "flow_id": "30Jul160949_324_celery6500_flow",
    "direction": "forward",
    "as_expected": true,
    "pkt_counts": [
      0,
      0,
      0,
      0,
      0,
      0
    ],
    "byte_counts": [
      0,
      0,
      0,
      0,
      0,
      0
    ],
    "discrepancies": [],
    "flow_rules_total": 6,
    "switch_rules_total": 3194,
    "flow_meters_total": 1,
    "switch_meters_total": 434,
    "ingress_mirror_flow_is_present": false,
    "egress_mirror_flow_is_present": false
  },
  {
    "flow_id": "30Jul160949_324_celery6500_flow",
    "direction": "reverse",
    "as_expected": true,
    "pkt_counts": [
      0,
      0,
      0,
      0,
      0,
      0
    ],
    "byte_counts": [
      0,
      0,
      0,
      0,
      0,
      0
    ],
    "discrepancies": [],
    "flow_rules_total": 6,
    "switch_rules_total": 3194,
    "flow_meters_total": 1,
    "switch_meters_total": 434,
    "ingress_mirror_flow_is_present": false,
    "egress_mirror_flow_is_present": false
  }
]
```

Kibana logs: 
FlowDumpResponse was split into 4 chunks.

4Chunks

RecievedOneFlowDumpResponseForm4Chunks

oleksir (approved):

pablomuri (approved):

PR #5738, opened 2024-09-17
[TEST]: Refactoring: Verification of the ping operation

yuliiamir:


src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/SubFlowPingDiscrepancies.groovy
```diff
@@ -0,0 +1,9 @@
+package org.openkilda.functionaltests.helpers.model
+
+import groovy.transform.TupleConstructor
+
+@TupleConstructor
+class SubFlowPingDiscrepancies {
+    String subFlowId
+    HashMap<FlowDirection, String> flowDiscrepancies
```
IvanChupin, line 8: Also, we can use Map interface here, instead of HashMap. if so, then in ComplexFlowPingResponse.groovy we can also use
`Map<FlowDirection, String> discrepancies = [:]` instead of 
`HashMap<FlowDirection, String> discrepancies = [:].`

IvanChupin (approved):

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/SubFlowPingDiscrepancies.groovy
yuliiamir: done

pablomuri (approved):

PR #5739, opened 2024-09-17
[TEST]: Improvement: DB calls optimization(ISLs bandwidth updating)

yuliiamir:


IvanChupin (approved):

src-java/testing/test-library/src/main/java/org/openkilda/testing/service/database/DatabaseSupportImpl.java
```diff
@@ -201,6 +201,27 @@ public boolean updateIslsAvailableBandwidth(List<Isl> islsToUpdate, long value)
...
+    /**
+     * Updates max_bandwidth property on specified ISLs.
+     *
+     * @param islsToUpdate ISLs to be changed
+     * @param availableValue available bandwidth to set
+     * @param maxValue max bandwidth to set
+     * @return true if at least 1 ISL was affected.
+     */
+    @Override
+    public boolean updateIslsAvailableAndMaxBandwidth(List<Isl> islsToUpdate, long availableValue, long maxValue) {
```
pablomuri, line 213: You aren't using the boolean. But I guess you added it just in case because the rest of the methods in the database interface also include it. Am I right?

pablomuri (approved):

src-java/testing/test-library/src/main/java/org/openkilda/testing/service/database/DatabaseSupportImpl.java
yuliiamir: yes, I added similar to other ISL params updating methods

PR #5741, opened 2024-10-02
[TEST]: Improvement: Verification of validation response

yuliiamir:


src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/ComplexFlowValidationResponse.groovy
```diff
@@ -0,0 +1,50 @@
...
+        this.subFlowsDiscrepancies = collectDiscrepanciesPerSubFlow(yFlowValidation.subFlowValidationResults)
+    }
+
+    ComplexFlowValidationResponse(HaFlowValidationResult haFlowValidation) {
+        this.asExpected = haFlowValidation.asExpected
+        this.subFlowsDiscrepancies = collectDiscrepanciesPerSubFlow(haFlowValidation.subFlowValidationResults)
+
+    }
+
+    static private List<SubFlowValidationDiscrepancies> collectDiscrepanciesPerSubFlow(List<FlowValidationDto> subFlowValidationDetails) {
```
IvanChupin, line 28: it might be better to use "private static" instead of "static private". 
for the sake of common approach

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/ComplexFlowValidationResponse.groovy
```diff
@@ -0,0 +1,50 @@
...
+        List<SubFlowValidationDiscrepancies> discrepanciesPerSubFlow = subFlowValidationDetails.groupBy { it.flowId }.collect{
+            HashMap<FlowDirection, List<PathDiscrepancyDto>> discrepancy = [:]
+            it.value.each {
+                verifyValidateLogic(it)
+                it.asExpected ?: discrepancy.put(FlowDirection.getByDirection(it.direction.toLowerCase()),  it.discrepancies)
```
IvanChupin, line 33: It might be not the case,
 
but is it possible that there are at least two discrepancies  for the same flowid and the same direction?
what if in the following loop :
```
it.value.each {
                verifyValidateLogic(it)
                it.asExpected ?: discrepancy.put(FlowDirection.getByDirection(it.direction.toLowerCase()),  it.discrepancies)
}
```
there will be at least two iteration with the same flowId and the same direction?

IvanChupin (approved):

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/ComplexFlowValidationResponse.groovy
yuliiamir: We interact with List<FlowValidationDto>  response and we use List<PathDiscrepancyDto> discrepancies that should can contain more than one discrepancy per direction.

yuliiamir: added check for direction per sub-flow (to clarify that we don't have 2 FORWARD validation responses with one discrepancy instead of one FORWARD and two discrepancies) + if we have some invalid direction, it fails during getting FlowDirection

pablomuri (approved):

PR #5751, opened 2024-11-14
nbworker add feature to reroute y-flow for switch/link evacuation process

IvanChupin:
Now it is possible to trigger y-flow reroute process by calling 
`/v1/switches/{switch-id}/under-maintenance` API with the following body:
```
{
  "under_maintenance": true,
  "evacuate": true
}
```

for transit switches of the y-flow. (not y-point, shared switches)

FOR TEST TEAM:
To test feature manually you have to go to the Switch details page, toggle switch into maintenance mode and then press evacuate button. This should evacuate flows and y-flow subflows from the switch. It will trigger reroute process for y-flows and flows that uses particular current switch.

if you click on the evacuate button without maintenance mode it should show you pop up info message
image
 
 
The same logic applied to the ISL details page.
  

For the info: 
If you have some isl in maintenance mode then you can not trigger evacuate flow process from switch without having it in maintenance mode.(and vice versa) What I was saying is that ISL maintenance mode and switch maintenance mode are not depend on each other at all and have no correlation.  

From the autotest point of view, please, ping me to discuss.

### DIFF
--- a/src-gui/ui/src/app/common/constants/constants.ts
+++ b/src-gui/ui/src/app/common/constants/constants.ts
@@ -36,6 +36,8 @@ export const MessageObj = {
     bfd_flag_updated: 'BFD flag updated successfully!',
     flows_evacuated: 'All flows are evacuated successfully!',
     error_flows_evacuated: 'Error in evacuating flows!',
+    info_cannot_evacuate_flows_from_switch: 'Can not evacuate flows while switch is not under maintenance.',
+    info_cannot_evacuate_flows_from_isl: 'Can not evacuate flows while ISL is not under maintenance.',
     reverse_graph_no_data: 'Backward graph API did not return data.',
     forward_graph_no_data: 'Forward graph API did not return data.',
     updating_isl_bandwidth: 'Updating ISL max bandwidth',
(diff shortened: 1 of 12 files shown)


yuliiamir:
Found issue: Bulk reroute has not been called for flows when triggering evacuation as an additional step after enabling Maintenance mode (for both ISl and switch).

yuliiamir:
Verified for Regular Flow and Y-Flow:

1. Evacuation from an intermediate switch that was under maintenance(flows use different intermediate switch Isls): 

- flows evacuation was called in the scope of maintenance mode enabling;

- flows evacuation was called as an additional step after enabling maintenance mode. 

Sw3WithRegularFlowAndYFlowBeforeReroute

After bulk evacuation, ONLY flows that have src/dst which is the intermediate switch are still on this switch. _Note, that reroute requests for these flows were sent and the flow path can be changed to the best one if it exists._

Sw3AfterBulkReroute

2. Evacuation from ISL under maintenance:

- flows evacuation was called in the scope of maintenance mode enabling;

- flows evacuation was called as an additional step after enabling maintenance mode. 


IslEvacuation2(6)(19)7

All flows were evacuated from ISL under maintenance. Note, when one of the Y-Flow sub-flows has this ISL in its path, reroute  is called for both sub-flows (both sub-flows can be rerouted (one in the scope of avoiding ISL under maintenance, another if the best path is found )).
_Note, that if there is not enough bandwidth for flows to be evacuated, these flows remain on the same path_

The required tests have been added in the scope of https://github.com/telstra/open-kilda/pull/5754

pablomuri (approved):
Great job!

PR #5767, opened 2025-02-10
[TEST]: Improvement: Switches: New interaction approach: Complex Flow

yuliiamir:


src-java/testing/functional-tests/src/test/groovy/org/openkilda/functionaltests/spec/flows/yflows/YFlowPathSwapSpec.groovy
```diff
@@ -210,7 +212,7 @@ class YFlowPathSwapSpec extends HealthCheckSpecification {
         }
 
         and: "All involved switches passes switch validation"
-        switchHelper.synchronizeAndCollectFixedDiscrepancies(updatedPathAfterPortUp.getInvolvedSwitches()).isEmpty()
+        synchronizeAndCollectFixedDiscrepancies(involvedSwitches).isEmpty()
```
yuliiamir, line 215: As path didn't changed (line #207-#212), we can use involvedSwitches to clarify that switches pass verification.

src-java/testing/functional-tests/src/main/groovy/org/openkilda/functionaltests/helpers/model/Switches.groovy
```diff
@@ -150,6 +158,18 @@ class Switches {
...
     }
 
+    /**
+     * Find all switches that are involved in a complex flow path
+     * @param complexFlowPath is a path of Y-Flow/HA-Flow to collect switchesIds
+     * @return list of SwitchExtended objects for further manipulation
+     */
+    List<SwitchExtended> findSwitchesInPath(FlowWithSubFlowsEntityPath complexFlowPath) {
+        def switchesToFind = complexFlowPath.getInvolvedSwitches()
+        def desiredSwitches = switches.findAll { it.switchId in switchesToFind }
```
IvanChupin, line 168: it might be worth to add .toSet() method to the following line:
`def switchesToFind = complexFlowPath.getInvolvedSwitches()`
otherwise` it.switchId in switchesToFind` search has linear complexity. O(N)

anyway if switchesToFind is big then it is worth to use `set`, otherwise no difference.

Above apply to all of the cases where we search in collection.

yuliiamir: our testing topology is not so big, maybe these changes can be added as an improvement later

pablomuri (approved):

PR #5779, opened 2025-03-28
blind ssrf fix

IvanChupin:
url validation added for saml config create endpoint and saml config update endpoint.
added unit tests.

### DIFF
--- a/src-gui/src/test/java/org/openkilda/saml/validator/SamlValidatorTest.java
+++ b/src-gui/src/test/java/org/openkilda/saml/validator/SamlValidatorTest.java
@@ -1,0 +1,169 @@
+/* Copyright 2025 Telstra Open Source
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


src-gui/src/test/java/org/openkilda/saml/validator/SamlValidatorTest.java
```diff
@@ -0,0 +1,169 @@
...
+        String malformedUrl = "htp://example.com";
+
+        thrown = assertThrows(RequestValidationException.class, () ->
+                samlConfigService.validateCreateProvider(file, "name", "entityId",
+                        malformedUrl, false, new ArrayList<>()));
+
+        assertEquals("URL is not valid", thrown.getMessage());
+    }
+
+    // Test Case 1: Valid URL (http/https)
```
pablomuri, line 115: This comment is repeated. Is it ok? line 141 is also repeated

pablomuri (approved):

src-gui/src/test/java/org/openkilda/saml/validator/SamlValidatorTest.java
IvanChupin: Thanks, fixed

